PR #2, opened 2022-06-19
Add initial project structure

benjiro:
Create and setup the initial project structure

image


src/OpenFeature.Api/OpenFeature.Api.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
toddbaert, line 4: I think this is the right target choice :+1:

Issue #3, opened 2022-06-26
Basic Flag Evaluation


benjiro:
Implement the most basic version of the specification to allow for retrieving flags in the defined forms (string, number, boolean, and structure)

benjiro:
Implemented under https://github.com/open-feature/dotnet-sdk/pull/5

PR #5, opened 2022-06-26
Basic flag evaluation support

benjiro:
Issue: https://github.com/open-feature/dotnet-sdk/issues/3

This PR implements the most basic form of the specification to allow retrival of string, number, boolean and structured flags.

Note this is a barebones start, IE I haven't Async'd anything yet


toddbaert:
I've seen this and plan on taking a close look in the next 2 days while I'm travelling! Thank you @benjiro

OpenFeature.sln
```diff
@@ -1,6 +1,6 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenFeature.Api", "src\OpenFeature.Api\OpenFeature.Api.csproj", "{07A6E6BD-FB7E-4B3B-9CBE-65AE9D0EB223}"
```
toddbaert, line 3: I think this is the right choice.

In the future, we may actually pull apart the application-author facing parts of the SDK into an "API" package, and have the full SDK consume that to keep things very small for the libraries, but there's no need for that yet.

src/OpenFeature/OpenFeature.cs
```diff
@@ -0,0 +1,30 @@
...
+using OpenFeature.Model;
+
+namespace OpenFeature
+{
+    public sealed class OpenFeature
+    {
+        private EvaluationContext _evaluationContext;
+        private IFeatureProvider _featureProvider = new NoOpFeatureProvider();
+        private readonly List<IHook> _hooks = new List<IHook>();
+        //TODO Add ILogger interface so we use the standard logging interface
```
toddbaert, line 12: Perhaps relevant: https://github.com/open-feature/spec/issues/39

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -0,0 +1,150 @@
...
+                .ToList();
+            
+            var hookContext = new HookContext<bool>(
+                flagKey,
+                defaultValue,
+                FlagValueType.Boolean,
+                _metadata,
+                OpenFeature.GetProviderMetadata(),
+                context
+            );*/
```
toddbaert, line 78: Looks like this is commented on purpose, b/c you haven't implemented hooks, yet, but just commenting to make sure you didn't mean to remove it.

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -0,0 +1,150 @@
...
+                    case FlagValueType.String:
+                        response = provider.ResolveStringValue(flagKey, (string)Convert.ChangeType(defaultValue, typeof(string)),
+                            context, options).ToFlagEvaluationDetails() as FlagEvaluationDetails<T>;
+                        break;
+                    case FlagValueType.Object:
+                        response = provider.ResolveStructureValue(flagKey, defaultValue, context, options).ToFlagEvaluationDetails();
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException(nameof(flagValueType), flagValueType, "Unknown flag value type");
+                }
```
toddbaert, line 105: You may be able to get rid of this switch case by using [delegates](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/delegates/). In your calls to this `EvaluateFlag` you can pass a ref to the `provider.ResolveXxxx` method you want to call, since the caller knows which one needs to be called. 

I did something similar in TS: https://github.com/open-feature/node-sdk/blob/98f93067d7779bf391d314af6a3c2b153523ba13/src/client.ts#L146

Up to you if you want to do this!

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+            finally
+            {
+                //TriggerHook(HookLifeCycle.Finally, allHooksReversed, hookContext, config);
+            }
+
+            return response;
+        }
+
+        private static void TriggerHook<T>(HookLifeCycle lifeCycle, IEnumerable<IHook> hooks, HookContext<T> context,
```
toddbaert, line 125: This is unused but uncommented.

toddbaert (approved):
Looks good to me overall as a partial implementation!

Thanks so much for your contribution!

src/OpenFeature/OpenFeatureClient.cs
benjiro: Yea i have got through implementing hooks and decided I should just get the basics done first. Ill rebase and remove this out for now

src/OpenFeature/OpenFeatureClient.cs
benjiro: Yea completely blanked about using a delegate. I've updated the implementation based on your recommendation

src/OpenFeature/OpenFeatureClient.cs
benjiro: Removed

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -0,0 +1,150 @@
...
+            FlagEvaluationOptions options, Exception error = null, FlagEvaluationDetails<T> evaluationDetails = null)
+        {
+            foreach (var hook in hooks)
+            {
+                switch (lifeCycle)
+                {
+                    case HookLifeCycle.Before:
+                        hook.Before(context, options.HookHints);
```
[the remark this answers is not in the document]
benjiro: Yea I will need to async the provider interface as well. We will probably need to provide both sync and async versions of the API. What are your thoughts on this? Standard for c# is to prefix any async methods with `{METHOD_NAME}Async`

src/OpenFeature/OpenFeature.cs
benjiro: Yea 100%, my thought has we would provide a method to set the [ILogger](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line) on the OpenFeature singleton and then we would just log using that. This allow the end users to easily integrate with their choice of logger (MS, Serilog, NLog, etc)

benjiro:
I'm going to move this PR forwards, so I can open the PR for hooks + context

src/OpenFeature/OpenFeatureClient.cs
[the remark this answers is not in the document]
toddbaert: > Thinking about it I don't think there is much point in supporting synchronous, async model was introduced in 2012 in c# language version 5. This is the defacto standard now. I'll address the async under hooks implementation PR instead of this.

I think this is the right direction to go.

Issue #7, opened 2022-06-26
Evaluation Context Support


benjiro:
Add support for evaluation context described in the specification https://github.com/open-feature/spec/blob/main/specification/evaluation-context.md

Issue #6, opened 2022-06-26
Hooks support


benjiro:
Add the hook behaviour described in the specification https://github.com/open-feature/spec/blob/main/specification/hooks.md

Issue #4, opened 2022-06-26
Setup Github Actions for CI


benjiro:
Create a CI/CD pipeline for the repo

benjiro:
- Build
- Test
- Code Coverage
- Setup styling compliance via dotnet-format
- Publishing package to nuget

beeme1mr:
For the time being, we can use [openfeature-core@groups.io](mailto:openfeature-core@groups.io) for publishing packages.

beeme1mr:
@benjiro, I'll take care of it. Unfortunately, the group email isn't working how I expected so I can't confirm the org. It may take a bit to figure out. FYI, @toddbaert.

beeme1mr:
@benjiro, I've contacted the NuGet support team because I'm not receiving the verification email. Hopefully we can get this addressed ASAP.

beeme1mr:
The NuGet support team still hasn't responded. I've contacted the CNCF to see if they offer an email service. If I don't hear something soon, I can setup a gmail account with an alias.

beeme1mr:
Got it working! @benjiro, do you have a NuGet account? I can add you as a collaborator.

benjiro:
@beeme1mr yea my nuget id is [benjiro](https://www.nuget.org/profiles/benjiro)

toddbaert:
Thanks for your work with the nuget red-tape @beeme1mr

beeme1mr:
Still nothing from the NuGet team. I'll send them another email.

beeme1mr:
The prefix OpenFeature.* has been reserved.

PR #8, opened 2022-07-02
Add support for EvaluationContext & Hooks

benjiro:
This PR will resolve the following issues https://github.com/open-feature/dotnet-sdk/issues/7 https://github.com/open-feature/dotnet-sdk/issues/6 https://github.com/open-feature/dotnet-sdk/issues/4

- Support hooks as per https://github.com/open-feature/spec/blob/main/specification/hooks.md
- Support Evaluation Context as per https://github.com/open-feature/spec/blob/main/specification/evaluation-context.md
- Add support for ILogger, defaults /dev/null https://docs.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line
- Async the world
- Add build scripts


src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -9,69 +12,171 @@ namespace OpenFeature
...
+
+                response = (await resolveValueDelegate.Invoke(flagKey, defaultValue, hookContext.EvaluationContext, options))
                     .ToFlagEvaluationDetails();
+                
+                await TriggerAfterHooks(allHooksReversed, hookContext, response, options);
             }
             catch (Exception e)
             {
+                OpenFeature.Logger.LogError(e, "Error while evaluating flag {FlagKey}", flagKey);
+                // TODO needs to handle error codes being thrown from provider that maps to Errors enums
```
benjiro, line 102: Plan on introducing a OpenFeatureException that the providers can throw and set the error code. Here is a working set of how I'm thinking to solve it. https://github.com/benjiro/dotnet-sdk/commit/b1956976e6698c4333f271ff2bd08e749ac3c37f

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -9,69 +12,171 @@ namespace OpenFeature
...
+                evaluationContext
+            );
+
             FlagEvaluationDetails<T> response;
             try
             {
-                response = resolveValueDelegate.Invoke(flagKey, defaultValue, context, config)
+                await TriggerBeforeHooks(allHooks, hookContext, options);
+
+                response = (await resolveValueDelegate.Invoke(flagKey, defaultValue, hookContext.EvaluationContext, options))
```
toddbaert, line 94: Total nitpick, but I'm not a huge fan of `response`. `resolution`, `evaluation` or something might be better.

src/OpenFeature/IHook.cs
```diff
@@ -12,9 +13,9 @@ namespace OpenFeature.Model
...
     public interface IHook
     {
-        EvaluationContext Before<T>(HookContext<T> context, IReadOnlyDictionary<string, object> hints = null);
-        void After<T>(HookContext<T> context, FlagEvaluationDetails<T> details, IReadOnlyDictionary<string, object> hints = null);
-        void Error<T>(HookContext<T> context, Exception error, IReadOnlyDictionary<string, object> hints = null);
-        void Finally<T>(HookContext<T> context, IReadOnlyDictionary<string, object> hints = null);
+        Task<EvaluationContext> Before<T>(HookContext<T> context, IReadOnlyDictionary<string, object> hints = null);
+        Task After<T>(HookContext<T> context, FlagEvaluationDetails<T> details, IReadOnlyDictionary<string, object> hints = null);
+        Task Error<T>(HookContext<T> context, Exception error, IReadOnlyDictionary<string, object> hints = null);
+        Task Finally<T>(HookContext<T> context, IReadOnlyDictionary<string, object> hints = null);
```
toddbaert, line 19: I could be missing something here, but I think this makes it required to have a hook implement every stage. Hooks don't need to feature every stage. Many would only implement one: https://github.com/open-feature/spec/blob/main/specification/hooks.md#requirement-431

I think the easiest way for you to implement this is to provide a [default implementation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods) which is basically a no-op. Then implementations would only implement the stage(s) they need.

This is what @justinabrahms did with the Java SDK (with TS/JS, method impls can be optional, so it's easy from the TS/JS side): https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/javasdk/Hook.java

src/OpenFeature/OpenFeature.cs
```diff
@@ -1,30 +1,34 @@
...
 {
     public sealed class OpenFeature
     {
-        private EvaluationContext _evaluationContext;
+        private EvaluationContext _evaluationContext = new EvaluationContext();
         private IFeatureProvider _featureProvider = new NoOpFeatureProvider();
         private readonly List<IHook> _hooks = new List<IHook>();
-        //TODO Add ILogger interface so we use the standard logging interface
-        
+        public static ILogger Logger { get; private set; } = new Logger<OpenFeature>(new NullLoggerFactory());
```
toddbaert, line 14: Relevant for later: https://github.com/open-feature/spec/issues/39

toddbaert (commented):
This looks great! Thanks again for your work and I'm sorry I took a while to finish the last review. I think the delegation pattern looks great, and I think async everywhere makes perfect sense!

I left a few comments.

One other thing I noticed is you covered basic test cases well, but some more specific ones seem to be missing (order of hook evaluation, for example).

I wouldn't consider it a blocker for this PR, since we're not publishing anything yet, but I would suggest you basically go through the [spec JSON](https://github.com/open-feature/spec/blob/main/specification.json), and convert every normative statement to a test. You can also check out the TS/JS test suite [here](https://github.com/open-feature/node-sdk/tree/main/test), which basically does that.

Eventually, we will also have a cross platform test suite. You can help with that if you'd rather go that way then hand-writing so many tests: https://github.com/open-feature/research/issues/11

src/OpenFeature/IHook.cs
benjiro: Yea this was a total oversight, I think it would probably make sense to just split each hook lifecycle out into it's own interface

```csharp
public interface IBeforeHook
public interface IAfterHook
public interface IErrorHool
public interface IFinallyHook
```

Then the consumer can just implement what they want.

```csharp
public class HookDemo : IBeforeHook, IAfterHook
```

src/OpenFeature/OpenFeature.cs
benjiro: Yea I was just playing around here, I can move it down to the client level like suggested in open-feature/spec#39

src/OpenFeature/IHook.cs
benjiro: Yea so after waking up and thinking about this more, C# added a [default implementation](https://docs.microsoft.com/en-au/dotnet/csharp/whats-new/tutorials/default-interface-methods-versions) of an interface that would allow this to work. Since we are .net standard 2.0 we don't have access to this functionality.

The best approach is to make the IHook interface an internal interface and expose an abstract class `Hook` that has a default implementation unless overriden

src/OpenFeature/OpenFeatureClient.cs
benjiro: Done 👍🏻

benjiro:
@toddbaert Thanks for the review 🙇🏻 

I've resolved your concerns around the Hook to allow for partial implementations. Re Logging, I've created an [issue](https://github.com/open-feature/dotnet-sdk/issues/13) in the repo to address this. I would like to do it under another PR as this one is already getting quite big. I can remove the logging stuff from this PR if you would like. 

Re-testing, I think it's good to keep a fully compliant test suite within the solution. The developer experience is much better if they are able to code and test in one IDE. Also agree with the agonistic approach, I think these can both coexist together. I've created an issue to implement a more comprehensive test suite. I'll resolve this under a subsequent pr.

src/OpenFeature/OpenFeature.cs
toddbaert: Ya, I dont think you should worry about the logger too much now, I just wanted to give you the heads-up.

src/OpenFeature/IHook.cs
toddbaert: Makes sense to me. Too bad default methods aren't available! This seems like a good alternative though.

toddbaert (approved):

PR #17, opened 2022-07-13
Make required changes to get ready to start releasing packages

benjiro:
Issue: #4

These are required changes to prepare for building NuGet packages

Opted to use [MinVer](https://github.com/adamralph/minver) to manage versions based on git tags. Will start the counter at 0.1.0 as discussed in https://github.com/open-feature/spec/issues/109


build/RELEASING.md
```diff
@@ -0,0 +1,19 @@
+﻿# Release process
```
benjiro, line 1: This is just notes for now once we automated the process we can update this file

src/OpenFeature/OpenFeature.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
```
toddbaert, line 4: I thought `net462` was included in `netstandard2.0` https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0 ?

Issue #20, opened 2022-07-14
Documentation


benjiro:
- [x] Code needs to be documented - https://github.com/open-feature/dotnet-sdk/pull/21
- [x] Readme needs to be updated with more examples - https://github.com/open-feature/dotnet-sdk/pull/41
- [x] Add readme to nuget package
- [x] Add examples/samples

PR #21, opened 2022-07-16
Add xmldoc comments to code

benjiro:
Issue: https://github.com/open-feature/dotnet-sdk/issues/20

- Document public code using xmldoc, and generate doc file
- Avoid parallel test running to fix race condition when setting multiple provides on the Singleton instance which causes problems in the OpenFeature singleton tests


oleg-nenashev (approved):
LGTM after a quick check

PR #23, opened 2022-07-17
Update namespace from OpenFeature -> OpenFeature.SDK

benjiro:
It is recommended not to share a namespace and class name to avoid import issues and code like OpenFeature.OpenFeature.Instance.*

Add test case for FeatureClient constructor that tests null provider


beeme1mr (approved):
Makes sense, thanks making the change.

Issue #27, opened 2022-07-26
Remove Hook and use only IHook?


toddbaert:
I think it's possible to define virtual methods on interfaces. Since the Hook abstract class has no state, I think there's no reason not to simply use `IHook` everywhere for more flexibility (multiple inheritance, etc)

benjiro:
@toddbaert I believe that was a feature added in C#8 and since we are targeting netstandard 2.0 to support .net framework 4.6.2 onwards we won't have access to C# 7.3 and below. I might be wrong here but here is the spec that was added int C#8 to enable what your suggesting https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods

benjiro:
Alternatively, we can just IHook and abstract since we don't need the capabilities of abstract since everything is optional. You can still achieve multiple inheritance with base classes

```csharp
class MyCoolHook : Hook, IEventCallback
{
....
}
```

Tell me what you think here is the PR with the proposed change https://github.com/open-feature/dotnet-sdk/pull/34

benjiro:
Yea sorry brain fart, I thought you were referring to implementing multiple interfaces 😅 . C# doesn't support multiple inheritances because they could never figure out a clean way to implement it across the CLR languages (C#, VB etc). The closest thing is the default implementation on interfaces in C#8 but that means we would need to drop support for net framework 4.6 >. 

There are a few other options

1. Abstract class which allows for optional implementations of methods. 
2. Split the interface into per lifecycle event (IBeforeHook, IAfterHook, ... Etc)
```csharp
public DemoHook : IBeforeHook, IAfterHook
```
3. Use interface and the implementation either throw a method `NotImplementedException` which we would need to make sure is handled by the call site if thrown. Or alternatively return null.

I'm leaning towards 1, I don't really see the need to allow class inheritances unless you can think of real important use case. This is a common approach in C#.

toddbaert:
> I'm leaning towards 1, I don't really see the need to allow class inheritances unless you can think of real important use case. This is a common approach in C#.

Yep. I agree. The real motivation for this only makes sense with C#8, so I think we can just move forward with an abstract or class.

PR #34, opened 2022-08-05
Chore: Remove IHook inteface and just use Hook class

benjiro:
#27 

There is no reason to have the internal IHook interface anymore. C#8
included default implementations on interfaces but since we are on netstandard2.0
we only have access to C#7.3

### DIFF
--- a/src/OpenFeature.SDK/Hook.cs
+++ b/src/OpenFeature.SDK/Hook.cs
@@ -5,14 +5,6 @@ using OpenFeature.SDK.Model;
 
 namespace OpenFeature.SDK
 {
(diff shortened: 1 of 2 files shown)


toddbaert:
See my comment [here](https://github.com/open-feature/dotnet-sdk/issues/27#issuecomment-1208281839)

benjiro:
@toddbaert No problems, I've replied here https://github.com/open-feature/dotnet-sdk/issues/27#issuecomment-1208706469

toddbaert (approved):
So Hook is basically now a "noop" hook... seems fine to me!

PR #35, opened 2022-08-05
Chore: Name the solution and project OpenFeature.SDK

benjiro:
Should of made this change with the namespace update under https://github.com/open-feature/dotnet-sdk/pull/23


beeme1mr:
Will this have any implication on the docs?

benjiro:
No it won't break the docs but it might break binary compat which would just mean recompiling the app which they would have to do if they change the target version

beeme1mr (approved):

PR #39, opened 2022-08-07
Chore: Add github action to perform release

benjiro:
This action will trigger when a new tag is push match {major}.{minor}.{patch}
semver format.

This that need to be setup
-  [x] NUGET_API_KEY needs to be set in github secrets, A key can be generated on nuget.org
- [x] We need to lockdown tag pushing if we haven't already done so https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules

### DIFF
--- a/.github/workflows/release-package.yml
+++ b/.github/workflows/release-package.yml
@@ -3,7 +3,7 @@ name: Release Package
 on:
   push:
     tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
   release-package:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-package.yml
```diff
@@ -0,0 +1,36 @@
+name: Release Package
+
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
```
beeme1mr, line 6:
```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+'
```
I configured the tag rule to expect a `v` prefix. Please let me know if this is an issue and I can update the rule.

beeme1mr (approved):
I had one question about tag prefixes but it's minor. I'll approve and let you decide what approach makes the most sense.

It's worth noting that the Nuget API token is only valid for one year. I'll sent a calendar reminder for myself to rotate the token before it expires.

PR #43, opened 2022-08-10
feat: update ctx merge order, add client ctx

toddbaert:
- Adds client context
- Updates context merge behavior
- Adds vs code settings / assocaited gitignores

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -341,3 +341,10 @@ ASALocalRun/
 /.sonarqube
 
 /src/LastMajorVersionBinaries
+
+# vscode
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
--- a/src/OpenFeature.SDK/OpenFeatureClient.cs
+++ b/src/OpenFeature.SDK/OpenFeatureClient.cs
@@ -20,6 +20,13 @@ namespace OpenFeature.SDK
         private readonly FeatureProvider _featureProvider;
         private readonly List<Hook> _hooks = new List<Hook>();
         private readonly ILogger _logger;
+        private readonly EvaluationContext _evaluationContext;
+
+        /// <summary>
+        /// Gets the client <see cref="EvaluationContext"/>
+        /// </summary>
+        /// <returns></returns>
+        public EvaluationContext GetContext() => this._evaluationContext;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="FeatureClient"/> class.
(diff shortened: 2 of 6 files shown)


toddbaert:
@benjiro This was a fairly painless experience. I basically just installed [this](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csharp) and then everything was good to go.

The only other thing I had a bit of trouble with was `dotnet-format`. I don't have this executable, but apparently the commend is included with dotnet now, so I just ran: `dotnet format OpenFeature.SDK.sln` which auto-fixed all the issues.

.gitignore
```diff
@@ -341,3 +341,10 @@ ASALocalRun/
 /.sonarqube
 
 /src/LastMajorVersionBinaries
+
+# vscode
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
```
toddbaert, line 350: These are work committing, the rest are not.

src/OpenFeature.SDK/OpenFeatureClient.cs
```diff
@@ -20,6 +20,13 @@ public sealed class FeatureClient : IFeatureClient
         private readonly FeatureProvider _featureProvider;
         private readonly List<Hook> _hooks = new List<Hook>();
         private readonly ILogger _logger;
+        private readonly EvaluationContext _evaluationContext;
+
+        /// <summary>
+        /// Gets the client <see cref="EvaluationContext"/>
+        /// </summary>
+        /// <returns></returns>
+        public EvaluationContext GetContext() => this._evaluationContext;
```
toddbaert, line 29: The client now has it's own context, as per spec. It can be supplied in the `.getClient()` method.

benjiro:
> @benjiro This was a fairly painless experience in vscode. I basically just installed [this](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csharp) and then everything was good to go.
> 
> The only other thing I had a bit of trouble with was `dotnet-format` (what the workflow runs). I don't have this executable, but apparently the command is included with `dotnet` now, so I just ran: `dotnet format OpenFeature.SDK.sln` which auto-fixed all the issues.

It seems I missed adding the tool to dotnet-tools.json config file. I'll update it under another PR and add instructions on how to install the tool and run the tool.

Basically, to add it to `.config/dotnet-tools.json` you run `dotnet tool install dotnet-format` then when first using the repo you run `dotnet tool restore` and it will download all the tools defined in `dotnet-tools.json`

To use the tool you run `dotnet format --folder` on the top level directory.

benjiro (approved):
LGTM. thanks for adding improving spec 4.3.4 tests, and picking up on the missing context on the client. 🏆

PR #45, opened 2022-08-11
bug: Correct release action

benjiro:
Notified these issues when trying to publish the new version

- Must tell minver that tag prefix is v
- Update proj name after renaming of the project
- Remove the version matrix since we don't need to run on multiple architecture


toddbaert (approved):

benjiro:
@toddbaert i took the changes from your other PR and applied them over the top as well 👍🏻

Issue #25, opened 2022-07-21
Evaluation context structured value.


kinyoklion:
Looking at the EvaluationContext I see that you can use any `object` as a value in the context. I am not clear how a provider could make use of non-trivial types, or be able to easily serialize the overall context without some level of interface requirement. (Unless they provided provider specific types, or specified specific types, but that doesn't seem very interoperable.)

In the basic use case, where there are top level basic types (string, number, boolean), I can see how a provider could request something as a type based on the meta-data that they likely have.

In the case of a `structure` use case I am not sure how that could be accomplished. I do see how it could be accomplished if the types were effectively restricted to each of the base types, as well as `ImmutableList` and `ImmutableDictionary` of those those types and more `ImmutableLists` and `ImmutableDictionary`.

Or is there going to be a builder which can mange those constraints?

Imagining a context of this general shape:
```
{
  "name": "example",
  "complex": {
    "field1": "value"
  }
}
```

How could a provider access the value of field1? Aside from using reflection, or using some pre-specified type?

Edit:
The node implementation contains this type, which I think expresses how this data can be made navigable and understood.
```
  export type JSONValue = null | string | number | boolean | Date | { [x: string]: JSONValue } | Array<JSONValue>;
```

Thank you,
Ryan

benjiro:
Thanks for the issue @kinyoklion, sorry I'm pretty tied up with work commitments at the moment so I haven't had a chance to look at this in detail.

The current implementation is crude and needs some work as you've discovered, the main reason I chose not to include some form of serialisation was I didn't want to tie the SDK to a specific library like Newtonsoft, System.Text.Json, etc. I believe this logic should be up to the provider to decide how they want to use it. The interactions in the context do need some work, currently, you can get a specific key using the Get method with the known type but if you are using array accessors you'd need to case the object which isn't ideal.

In your example, I would expect the provider to define a structured object say 

```csharp
public class ComplexObject 
{
    public Name {get;set;}
    public Dictionary<string, string> Complex {get;set;}
}

context.Add("complex-object", new ComplexObject());

var complexObject = context.Get<ComplexObject>("complex-object")
complexObject.complex["field1"]
```

The provider would transform that into the Provider specific SDK, lets take cloudbee/rollout for example which uses a ContextBuilder object. it would look something like

```csharp
// https://docs.cloudbees.com/docs/cloudbees-feature-management-api/latest/api-reference/csharp-api#_interface_icontext
[...]

PR #53, opened 2022-08-27
Rework EvaluationContext to use Structure type

benjiro:
Born from the discussions in #25 and https://github.com/open-feature/java-sdk/issues/50

- Adds `Structure` type which represents JSON data
- Adds `Value` wrapper class that houses the supported types (Structure, List, bool, int, double, Value - note ints are stored and doubles, and rounded when returned)
- `EvaluationContext` exposes methods to interact with its underlying keyvalue pairs of string, `Value`
- Update interfaces to use `Structure` instead of generics

### DIFF
--- a/src/OpenFeature.SDK/Model/EvaluationContext.cs
+++ b/src/OpenFeature.SDK/Model/EvaluationContext.cs
@@ -1,7 +1,5 @@
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 
 namespace OpenFeature.SDK.Model
 {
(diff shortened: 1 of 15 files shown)


toddbaert:
PSA: I'm going to bring this one home at the request of @benjiro (thanks for doing most of the work already!)

src/OpenFeature.SDK/Model/EvaluationContext.cs
```diff
@@ -10,96 +8,175 @@ namespace OpenFeature.SDK.Model
...
+            return this;
+        }
+
+        /// <summary>
+        /// Add a new double Value to the evaluation context
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public EvaluationContext Add(string key, double value)
```
toddbaert, line 94: In [java](https://github.com/open-feature/java-sdk/blob/d5bd0fbf02b00da71149bf99cf108e0020f0afa2/src/main/java/dev/openfeature/javasdk/EvaluationContext.java#L16), we use [lombok's delegate](https://projectlombok.org/features/experimental/Delegate) to automatically proxy these methods to the encapsulated `structure`. I would like to do something like that here, but I couldn't find any such solutions.

@kinyoklion maybe you have an idea?

toddbaert:
@kinyoklion I would like your take on [this relevant issue](https://github.com/open-feature/java-sdk/issues/58) if you have a moment, I've come to value your input :sweat_smile:

src/OpenFeature.SDK/Model/EvaluationContext.cs
kinyoklion: In regards to directly replicating what can be done with lombok, I do not think that there is a wide-spread or idiomatic equivalent. 

There are ways to accomplish it using something like fody/cecil, but I don't think that would be good for a library like this.

toddbaert: Cool, I don't think it's a hard requirement, just wondering if you had a simple solution up your sleeve.

beeme1mr (approved):

toddbaert:
@benjiro  @kinyoklion I've added changes analogous to https://github.com/open-feature/java-sdk/pull/61 here as well.

kinyoklion (approved):
Code looks good to me. Not sure if the empty XML comments are a problem.

toddbaert:
> Code looks good to me. Not sure if the empty XML comments are a problem.

I thought I got them all. Can you point out any I missed?

benjiro (commented):
LGTM 🏆 Thanks for finishing this up. Few params comments missing but looks fine otherwise.

kinyoklion:
Not a problem exactly, but this does allow for circular references. I assume that would be a "Just don't do that." Just thinking through if a provider would need to handle that situation.

toddbaert:
> Not a problem exactly, but this does allow for circular references. I assume that would be a "Just don't do that." Just thinking through if a provider would need to handle that situation.

Ya, good point. I've seen similar issues with some vendor's SDKs as well, especially converting circular objects to JSON in javascript. It might be worth noting somewhere, maybe even in our general documentation.

I will update the param comments and merge. I think we should hesitate on a release for a while though, since I think we may have one more breaking change related to: https://github.com/open-feature/spec/issues/138

PR #55, opened 2022-09-08
Add setter for context

toddbaert:
Adding this for consistency with other SDKs.

### DIFF
--- a/src/OpenFeature.SDK/OpenFeatureClient.cs
+++ b/src/OpenFeature.SDK/OpenFeatureClient.cs
@@ -20,13 +20,18 @@ namespace OpenFeature.SDK
         private readonly FeatureProvider _featureProvider;
         private readonly List<Hook> _hooks = new List<Hook>();
         private readonly ILogger _logger;
-        private readonly EvaluationContext _evaluationContext;
+        private EvaluationContext _evaluationContext;
 
         /// <summary>
-        /// Gets the client <see cref="EvaluationContext"/>
+        /// Gets the EvaluationContext of this client<see cref="EvaluationContext"/>
         /// </summary>
-        /// <returns></returns>
+        /// <returns><see cref="EvaluationContext"/>of this client</returns>
         public EvaluationContext GetContext() => this._evaluationContext;
+
+        /// <summary>
+        /// Sets the EvaluationContext of the client<see cref="EvaluationContext"/>
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/OpenFeature.SDK/OpenFeatureClient.cs
```diff
@@ -20,14 +20,19 @@ public sealed class FeatureClient : IFeatureClient
...
+        /// Gets the EvaluationContext of this client<see cref="EvaluationContext"/>
         /// </summary>
-        /// <returns></returns>
+        /// <returns><see cref="EvaluationContext"/>of this client</returns>
         public EvaluationContext GetContext() => this._evaluationContext;
 
+        /// <summary>
+        /// Sets the EvaluationContext of the client<see cref="EvaluationContext"/>
+        /// </summary>
+        public void SetContext(EvaluationContext evaluationContext) => this._evaluationContext = evaluationContext;
```
kinyoklion, line 34: It isn't specifically related to this change, but is this code generally expected to be thread safe?

toddbaert: > It isn't specifically related to this change, but is this code generally expected to be thread safe?

Hmm, there's nothing in the spec about it, but I think that we'd likely want the context in the client and the api singleton to be thread safe, so that we could have some predictability around it.

Are you thinking we should synchronize this method?

kinyoklion: I've not looked through the code much yet with an eye for thread safety. Synchronizing set/get access to this method would affect which instance is being used during an evaluation when it was concurrently modified, which would make ordering a bit more predictable. 

 That said the context itself is mutable, so thread behavior around mutating the context itself is likely more problematic.

I suspect there are currently opportunities for concurrent modification exceptions to be thrown if the context, or contained structures, was being mutated while other things may be trying to enumerate 

Probably doesn't need to mingle with this PR.

toddbaert: I will add an issue to investigate thread safety overall in the this and the Java SDK.

I suspect that whatever thread safety issues exist here we also have in Java.

kinyoklion (approved):

src/OpenFeature.SDK/OpenFeatureClient.cs
toddbaert: Created https://github.com/open-feature/dotnet-sdk/issues/56

PR #57, opened 2022-09-09
BREAKING: add structure->value, object value constructor

toddbaert:
A c# flavored version of: https://github.com/open-feature/java-sdk/pull/65. I hope these are the last breaking changes before a 1.0rc.

- add object constructor and accessor to `Value`
- use `Value` instead of `Structure` in object resolver (https://github.com/open-feature/spec/issues/138#issuecomment-1241202093)

### DIFF
--- a/src/OpenFeature.SDK/Model/Value.cs
+++ b/src/OpenFeature.SDK/Model/Value.cs
@@ -25,13 +25,13 @@ namespace OpenFeature.SDK.Model
         {
             // integer is a special case, convert those.
             this._innerValue = value is int ? Convert.ToDouble(value) : value;
-            if (!(this.IsNull()
-                || this.IsBoolean()
-                || this.IsString()
-                || this.IsNumber()
-                || this.IsStructure()
-                || this.IsList()
-                || this.IsDateTime()))
+            if (!(this.IsNull
+                || this.IsBoolean
+                || this.IsString
+                || this.IsNumber
+                || this.IsStructure
+                || this.IsList
+                || this.IsDateTime))
             {
                 throw new ArgumentException("Invalid value type: " + value.GetType());
             }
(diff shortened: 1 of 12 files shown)


src/OpenFeature.SDK/Model/Value.cs
```diff
@@ -17,6 +17,27 @@ public class Value
...
 
+        /// <summary>
+        /// Creates a Value with the inner set to the object
+        /// </summary>
+        /// <param name="value"><see cref="Object">The object to set as the inner value</see></param>
+        public Value(Object value)
+        {
+            // integer is a special case, convert those.
+            this._innerValue = value is int ? Convert.ToDouble(value) : value;
+            if (!(this.IsNull()
```
kinyoklion, line 28: Just an off-hand comment, but I wonder if all these is/as methods should be properties?

kinyoklion (approved):

src/OpenFeature.SDK/Model/Value.cs
toddbaert: I'd be totally ok with that. I'm comfortable with C#, but use it infrequently enough that I don't have a good feel for what's idiomatic in terms of such stylistic choices.

If you think properties are a better choice I'll happily switch. I'd much rather do so now when we're already doing breaking changes if you think it's a better choice.

kinyoklion: I would say that properties would be more idiomatic. Generally if it could be thought of as data, and you are getting it, and it doesn't cause a bunch of side-effects, then properties are a good choice.

I didn't notice until I was using it and it felt a bit off.

MS has some more specific guidelines around it.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

toddbaert: Will make this change. Thanks again!

kinyoklion (approved):

PR #58, opened 2022-09-13
Document EvaluationContext members.

kinyoklion:
When working on a provider I noticed that there were a number of empty documentation comments in the EvaluationContext.

In addition to filling those out I thought it would be nice if it was clear what the behavior was regarding missing keys, duplicate keys, etc.


toddbaert (approved):
LGTM, thanks!

toddbaert:
Looks like you need to sign off your commits.

kinyoklion:
> Looks like you need to sign off your commits.

I think I have rectified it. Thank you.

Issue #59, opened 2022-09-19
Proposal: Use "release please" for easy and democratic releases


toddbaert:
I'm wondering if we could get a consensus on using [release please](https://github.com/googleapis/release-please).

Basically, it's an github action that runs after commits/merges to main. It checks for semantic commits, and if it detects a `feat/fix/breaking` change, it creates a running PR (see example in Java SDK [here](https://github.com/open-feature/java-sdk/pull/76)). The action automatically and intelligently increments the versions numbers in this PR with new commits to main by searching for [annotations](https://github.com/open-feature/java-sdk/blob/2eec1a5519b9efab7d7f9dc8b1cbd84d9218368b/pom.xml#L7) in XML/JSON/etc.

When these PRs are merged, the existing release process runs. Currently, we use this in the Java SDK, Go SDK, Node SDK, and all the contrib repos. To ease things for deveopers, we use github's "sqush and merge" setting to use the PR title as the commit message. We also have a github action that lints the commit message. This means that the only thing that needs to follow semantic commits is the PR title.

Benefits:

- breaking changes are identified, audited and recorded at PR time, not at release time, reducing burden on release manager.
- releases are democratic (they are PRs opened by a bot that must be merged)
- semantic versioning is done automatically
- CHANGELOG is built automatically

cc @benjiro

benjiro:
100% agree to standardise the release experience across the repos.

Ill have a go a setting it all up.

PR #61, opened 2022-09-22
chore: use release please

benjiro:
#59

Standardise the release/development experience across the openfeature SDKs by using release-please.

- Remove usage of minver
- Remove dotnet releaser
- Setup release please and use simple type
- Add action to lint commit messages


toddbaert:
Thanks for doing this @benjiro!

This makes sense to me. My initial ideas was to use the "simple" strategy and annotate versions in the .proj files like [so](https://github.com/open-feature/java-sdk/blob/2eec1a5519b9efab7d7f9dc8b1cbd84d9218368b/pom.xml#L7), but if the `dotnet-yoshi` strategy works as it seems to, I think it's a good way to go.

Thanks again! Maybe we can do a fix release that updates something trivial (like a comment typo or something) and test all works as expected.

toddbaert (approved):

benjiro:
@toddbaert How does the `simple` strategy work? Is it up to the end user to increment it or does `release-please` know how to grep pom.xml and update it?

toddbaert:
It looks for the `<!--x-release-please-version -->` (or `// x-release-please-version`, `# release-please-version`) in other file types, depending on the comment style for that language, and increments it. There's nothing manual. The release plz action already "knows" the version, the annotation just tells it what field to update. You can even update readmes that way: https://github.com/open-feature/java-sdk/blob/f8d62eea2f31ccf42bbafa3b591a83383bd6e10f/README.md?plain=1#L63-L72. The only other think you need to do is mention these files in the [config](https://github.com/open-feature/java-sdk/blob/f8d62eea2f31ccf42bbafa3b591a83383bd6e10f/release-please-config.json#L10-L13).

I'm pretty sure you can use these annotations in any strategy though. The only difference with the more intelligent strategies is they "know" where to update in many cases.

I don't think one solution is better than the other, unless you can think of some reason way.

PR #62, opened 2022-09-22
fix!: Rename namespace from OpenFeature.SDK to OpenFeatureSDK

benjiro:
Discovered when implementing a provider, I noticed due conflict with OpenFeature.SDK namespace and the singleton class being called OpenFeature.
It requires the full namespace to use the OpenFeature singleton.

Avoid this by making the root namespace `OpenFeatureSDK` instead of `OpenFeature.SDK`

An example of the issue
image

After the proposed change
```csharp
using OpenFeatureSDK;

OpenFeature.Instance.SetProvider(new CloudBeesProvider());
services.AddSingleton(provider => OpenFeature.Instance.GetClient());
```


toddbaert:
I would say this is a `fix`, which will include it in the changelog. I recommend you change the PR title.

toddbaert (approved):

Issue #71, opened 2022-09-27
Update SDK to be compliant with spec v0.5.0


beeme1mr:
Update the SDK to be compliant with the v0.5.0 spec.

### ⚠ BREAKING CHANGES

- [x] https://github.com/open-feature/spec/pull/142 Error codes must be a constrained set of values (e.g. enum). #72

### Additional changes

- [x] https://github.com/open-feature/spec/pull/140 Clarify that `reason` must be a string in the `flag resolution details` and `evaluation details`.
- [x] https://github.com/open-feature/spec/pull/142 Add optional `error message` to `flag resolution details`. #72
 
## General

 - [x] Add/update a spec version badge on the readme. #74 

[![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.5.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.5.0)

**Full Changelog**: https://github.com/open-feature/spec/compare/v0.4.0...v0.5.0

kinyoklion:
Anyone mind if I pick this up?

benjiro:
Closing as all specification 0.5.0 are now resolved

PR #72, opened 2022-09-27
feat!: ErrorType as enum, add ErrorMessage string

kinyoklion:
Addresses: https://github.com/open-feature/dotnet-sdk/issues/71

Extend the `ErrorType` to include `INVALID_CONTEXT` and `TARGETING_KEY_MISSING`.
Change the `ErrorType` in `FlagEvaluationDetails` to an enum.
Update the `FeatureProviderException` to have an enumerated code. Use `message` for the `error message`.
Update tests, and update specification references in tests.

### DIFF
--- a/src/OpenFeatureSDK/Error/FeatureProviderException.cs
+++ b/src/OpenFeatureSDK/Error/FeatureProviderException.cs
@@ -1,6 +1,5 @@
 using System;
 using OpenFeatureSDK.Constant;
-using OpenFeatureSDK.Extension;
 
 namespace OpenFeatureSDK.Error
 {
@@ -11,9 +10,9 @@ namespace OpenFeatureSDK.Error
     public class FeatureProviderException : Exception
     {
         /// <summary>
-        /// Description of error that occured when evaluating a flag
+        /// Error that occurred during evaluation
         /// </summary>
-        public string ErrorDescription { get; }
+        public ErrorType ErrorType { get; }
 
         /// <summary>
         /// Initialize a new instance of the <see cref="FeatureProviderException"/> class
@@ -24,19 +23,7 @@ namespace OpenFeatureSDK.Error
         public FeatureProviderException(ErrorType errorType, string message = null, Exception innerException = null)
             : base(message, innerException)
         {
-            this.ErrorDescription = errorType.GetDescription();
-        }
-
-        /// <summary>
-        /// Initialize a new instance of the <see cref="FeatureProviderException"/> class
-        /// </summary>
-        /// <param name="errorCode">A string representation describing the error that occured</param>
-        /// <param name="message">Exception message</param>
-        /// <param name="innerException">Optional inner exception</param>
-        public FeatureProviderException(string errorCode, string message = null, Exception innerException = null)
-            : base(message, innerException)
-        {
-            this.ErrorDescription = errorCode;
+            this.ErrorType = errorType;
         }
     }
 }
--- a/src/OpenFeatureSDK/Model/FlagEvaluationDetails.cs
+++ b/src/OpenFeatureSDK/Model/FlagEvaluationDetails.cs
@@ -1,5 +1,4 @@
 using OpenFeatureSDK.Constant;
-using OpenFeatureSDK.Extension;
 
 namespace OpenFeatureSDK.Model
 {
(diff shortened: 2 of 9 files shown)


src/OpenFeatureSDK/Model/FlagEvaluationDetails.cs
```diff
@@ -1,5 +1,4 @@
 using OpenFeatureSDK.Constant;
```
kinyoklion, line 1: This file was renamed `Evalusation` -> `Evaluation`

test/OpenFeatureSDK.Tests/FeatureProviderTests.cs
```diff
@@ -21,13 +22,22 @@ public void Provider_Must_Have_Metadata()
...
-        [Specification("2.3.1", "The `feature provider` interface MUST define methods for typed flag resolution, including boolean, numeric, string, and structure.")]
-        [Specification("2.4", "In cases of normal execution, the `provider` MUST populate the `flag resolution` structure's `value` field with the resolved flag value.")]
-        [Specification("2.5", "In cases of normal execution, the `provider` SHOULD populate the `flag resolution` structure's `variant` field with a string identifier corresponding to the returned flag value.")]
-        [Specification("2.6", "The `provider` SHOULD populate the `flag resolution` structure's `reason` field with a string indicating the semantic reason for the returned flag value.")]
-        [Specification("2.7", "In cases of normal execution, the `provider` MUST NOT populate the `flag resolution` structure's `error code` field, or otherwise must populate it with a null or falsy value.")]
-        [Specification("2.9", "In cases of normal execution, the `provider` MUST NOT populate the `flag resolution` structure's `error code` field, or otherwise must populate it with a null or falsy value.")]
```
kinyoklion, line 30: `2.9` was a duplicate of `2.7`. I've updated it to `2.9.1` and transcribed the text. This test already demonstrated the associated specification supporting generics.

test/OpenFeatureSDK.Tests/OpenFeatureClientTests.cs
```diff
@@ -321,6 +321,36 @@ public async Task Should_Resolve_StructureValue()
             featureProviderMock.Verify(x => x.ResolveStructureValue(flagName, defaultValue, It.IsAny<EvaluationContext>()), Times.Once);
         }
 
+        [Fact]
```
kinyoklion, line 324: I added an extra test here for the case where exceptions are not being used.

benjiro (approved):
LGTM, thanks for your contribution

src/OpenFeatureSDK/Model/FlagEvaluationDetails.cs
benjiro, line 1: Good catch on the typo 🏆

src/OpenFeatureSDK/Error/FeatureProviderException.cs
```diff
@@ -24,19 +23,7 @@ public class FeatureProviderException : Exception
         public FeatureProviderException(ErrorType errorType, string message = null, Exception innerException = null)
             : base(message, innerException)
         {
```
toddbaert, line 25: In Java we use dedicated exception types for each error code. I'm not saying that's better, just an option to consider.

This might actually be easier for some providers :shrug:

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -257,9 +257,9 @@ public FeatureClient(FeatureProvider featureProvider, string name, string versio
             catch (FeatureProviderException ex)
             {
                 this._logger.LogError(ex, "Error while evaluating flag {FlagKey}. Error {ErrorType}", flagKey,
-                    ex.ErrorDescription);
-                evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, ex.ErrorDescription, Reason.Error,
-                    string.Empty);
+                    ex.ErrorType.GetDescription());
+                evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, ex.ErrorType, Reason.Error,
+                    string.Empty, ex.Message);
```
toddbaert, line 262: Not directly related to your change... but is `null` a better choice than `string.Empty` for the variant?

toddbaert (approved):

src/OpenFeatureSDK/OpenFeatureClient.cs
kinyoklion: I think that `null` is reasonable, and also what I would have done. Reference and nullable types still seem to be the most idiomatic way to handle optionality in C#.

I am not sure how fast the community is moving this way, but C# has supported additional nullable contexts for a bit. Which allows for something akin to null safety with static analysis.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types
(Now enabled by default for dotnet 6 and newer templates.)

At which points these `null` values will probably feel a little better.

PR #77, opened 2022-10-04
chore: poc, Implement builders and immutable contexts.

kinyoklion:
Proof of concept implementing builders for `EvaluationContext` and `Structure` types to address context threading concerns.

```
var sample = Structure.Builder().Set("key", "val").Build();
var context = EvalContext.Builder().Set("my-struct", sample).Build();

```

This allows for the context provided to the client to be immutable, and for immutability for the duration of a hook.

### DIFF
--- a/src/OpenFeatureSDK/Model/EvaluationContext.cs
+++ b/src/OpenFeatureSDK/Model/EvaluationContext.cs
@@ -12,11 +12,18 @@ namespace OpenFeatureSDK.Model
     {
         private readonly Structure _structure;
 
+        /// <summary>
+        /// Internal constructor used by the builder.
+        /// </summary>
+        /// <param name="content">The content of the context.</param>
         internal EvaluationContext(Structure content)
         {
             this._structure = content;
         }
 
+        /// <summary>
+        /// Private constructor for making an empty <see cref="EvaluationContext"/>.
+        /// </summary>
         private EvaluationContext()
         {
             this._structure = Structure.Empty;
@@ -82,6 +89,15 @@ namespace OpenFeatureSDK.Model
         public IEnumerator<KeyValuePair<string, Value>> GetEnumerator()
         {
             return this._structure.GetEnumerator();
+        }
+
+        /// <summary>
+        /// Get a builder which can build an <see cref="EvaluationContext"/>.
+        /// </summary>
+        /// <returns>The builder</returns>
+        public static EvaluationContextBuilder Builder()
+        {
+            return new EvaluationContextBuilder();
         }
     }
 }
--- a/src/OpenFeatureSDK/Model/EvaluationContext.cs
+++ b/src/OpenFeatureSDK/Model/EvaluationContext.cs
@@ -22,9 +22,10 @@ namespace OpenFeatureSDK.Model
             this._structure = Structure.Empty;
         }
 
-        private static EvaluationContext _empty = new EvaluationContext();
-
-        public static EvaluationContext Empty => _empty;
+        /// <summary>
+        /// An empty evaluation context.
+        /// </summary>
+        public static readonly EvaluationContext Empty = new EvaluationContext();
 
         /// <summary>
         /// Gets the Value at the specified key
--- a/src/OpenFeatureSDK/Model/EvaluationContext.cs
+++ b/src/OpenFeatureSDK/Model/EvaluationContext.cs
@@ -8,9 +8,23 @@ namespace OpenFeatureSDK.Model
     /// to the feature flag evaluation context.
     /// </summary>
     /// <seealso href="https://github.com/open-feature/spec/blob/main/specification/evaluation-context.md">Evaluation context</seealso>
-    public class EvaluationContext
+    public sealed class EvaluationContext
     {
-        private readonly Structure _structure = new Structure();
+        private readonly Structure _structure;
+
+        internal EvaluationContext(Structure content)
+        {
+            this._structure = content;
+        }
+
+        private EvaluationContext()
+        {
+            this._structure = Structure.Empty;
+        }
+
+        private static EvaluationContext _empty = new EvaluationContext();
+
+        public static EvaluationContext Empty => _empty;
 
         /// <summary>
         /// Gets the Value at the specified key
@@ -36,15 +50,6 @@ namespace OpenFeatureSDK.Model
         public bool ContainsKey(string key) => this._structure.ContainsKey(key);
 
         /// <summary>
-        /// Removes the Value at the specified key
-        /// </summary>
-        /// <param name="key">The key of the value to be removed</param>
-        /// <exception cref="ArgumentNullException">
-        /// Thrown when the key is <see langword="null" />
-        /// </exception>
-        public void Remove(string key) => this._structure.Remove(key);
-
-        /// <summary>
         /// Gets the value associated with the specified key
         /// </summary>
         /// <param name="value">The <see cref="Value"/> or <see langword="null" /> if the key was not present</param>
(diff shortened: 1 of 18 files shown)


kinyoklion:
I also have a branch where I implemented interfaces for the builders, structure, and eval context: https://github.com/open-feature/dotnet-sdk/pull/new/rlamb/immutable-with-builders-interfaces

Working through this I wasn't as sure about implementing interfaces. Because if they are interfaces, then they are effectively  unconstrained with their threading guarantees. Though possibly that is fine. It is even easier to document that if you implement the interface directly, then all bets are off.

The nice thing about concrete types though is you can seal them be and be sure within reason they are your types.

kinyoklion:
I've also been thinking about this, and the interfaces that are in the Java PR.

src/OpenFeatureSDK/Model/EvaluationContext.cs
```diff
@@ -8,9 +8,31 @@ namespace OpenFeatureSDK.Model
     /// to the feature flag evaluation context.
     /// </summary>
     /// <seealso href="https://github.com/open-feature/spec/blob/main/specification/evaluation-context.md">Evaluation context</seealso>
-    public class EvaluationContext
+    public sealed class EvaluationContext
```
benjiro, line 11: Nice addition, completely missed this when writing the initial implementation

src/OpenFeatureSDK/Model/EvaluationContext.cs
```diff
@@ -64,182 +77,27 @@ public class EvaluationContext
             return new Dictionary<string, Value>(this._structure.AsDictionary());
```
benjiro, line 77: Should we be returning am immutable dictionary?

src/OpenFeatureSDK/Model/HookContext.cs
```diff
@@ -65,5 +65,17 @@ public class HookContext<T>
             this.ProviderMetadata = providerMetadata ?? throw new ArgumentNullException(nameof(providerMetadata));
             this.EvaluationContext = evaluationContext ?? throw new ArgumentNullException(nameof(evaluationContext));
         }
+
+        internal HookContext<T> WithNewEvalContext(EvaluationContext context)
```
benjiro, line 69: Does it make sense to shorten this? Every where else we have referred to it in full
```suggestion
        internal HookContext<T> WithNewEvaluationContext(EvaluationContext context)
```

src/OpenFeatureSDK/Model/Structure.cs
```diff
@@ -1,32 +1,45 @@
...
-            this._attributes = new Dictionary<string, Value>();
+            this._attributes = attributes;
         }
 
+        /// <summary>
+        /// Private constructor for creating an empty <see cref="Structure"/>.
+        /// </summary>
+        private Structure()
+        {
+            this._attributes = ImmutableDictionary<string, Value>.Empty; ;
```
benjiro, line 28: Remove an extra semi-colon
```suggestion
            this._attributes = ImmutableDictionary<string, Value>.Empty;

```

src/OpenFeatureSDK/Model/Value.cs
```diff
@@ -23,6 +23,10 @@ public class Value
         /// <param name="value"><see cref="Object">The object to set as the inner value</see></param>
         public Value(Object value)
         {
+            if (value is IList<Value> list)
```
benjiro, line 26: Nice catch

src/OpenFeatureSDK/OpenFeatureSDK.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggerVer)" />
+    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
```
benjiro, line 10: Can you move the dep version into [Common.props](https://github.com/open-feature/dotnet-sdk/blob/main/build/Common.props#L22). We probably also want to allow a range of Immutable versions instead of hard locking on 6. This SDK currently supports back to net462

benjiro (commented):
@kinyoklion awesome work, I left a few comments, but everything looks solid. 🏆

src/OpenFeatureSDK/Model/EvaluationContext.cs
kinyoklion: It would probably be for the best.

src/OpenFeatureSDK/Model/HookContext.cs
kinyoklion: I think it makes sense to use the full name. Will commit your suggestion.

src/OpenFeatureSDK/Model/Structure.cs
kinyoklion: Apparently this line of code was sad about something. Committing your suggestion.

src/OpenFeatureSDK/OpenFeatureSDK.csproj
kinyoklion: Yeah, I can do that.

I can add a range as well. 6 works with 462, but so does 1.x and 7.
https://www.nuget.org/packages/System.Collections.Immutable/6.0.0#supportedframeworks-body-tab

src/OpenFeatureSDK/Model/EvaluationContext.cs
kinyoklion: Updated. It actually made things better. The structure is an immutable dictionary now, so it just flows through.

src/OpenFeatureSDK/OpenFeatureSDK.csproj
kinyoklion: I made a range of 1.7.1 inclusive through 7.0.0 exclusive. As 7.0.0 isn't released yet. Not much changes about these collections.

kinyoklion:
GC thrashing was brought up as a possibility. This could potentially be susceptible to that.

One way to reduce that would be to change the types of `Value`, and possibly `Struct` to be `structs`. Then they would be value types and mostly inhabit the stack, or be inline in other objects. Because there are collections, those are still inherently references, but you would only encounter the cost if using them. 

That would be more code than this currently is though. It would also use a bit more memory per value, but the dozens of bytes range. Because each of the types would exist in the struct. and only be used if the type was that type. So you have an enum and a number of extra fields.

A lot of this could change internally, and code would still compile. A breaking change at the ABI level though, for people just dropping in assemblies.

src/OpenFeatureSDK/OpenFeatureSDK.csproj
benjiro: Thanks, yea it's more to avoid unintentionally upgrading an older project consuming the openfeature dependency on immutable.

toddbaert:
Looks great!

src/OpenFeatureSDK/Model/StructureBuilder.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Remove the specified key.
+        /// </summary>
+        /// <param name="key">The key to remove</param>
+        /// <returns>This builder</returns>
+        public StructureBuilder Remove(string key)
+        {
+            this._attributes.Remove(key);
+            return this;
+        }
```
toddbaert, line 145: Is this method needed now? Is is realistic that an builder would add and then remove something before having `build()` called on it?

src/OpenFeatureSDK/Model/EvaluationContextBuilder.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        /// <summary>
+        /// Remove the given key from the context.
+        /// </summary>
+        /// <param name="key">The key to remove</param>
+        /// <returns>This builder</returns>
+        public EvaluationContextBuilder Remove(string key)
+        {
+            this._attributes.Remove(key);
+            return this;
```
toddbaert, line 126: see: https://github.com/open-feature/dotnet-sdk/pull/77/files#r988214395

src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -11,7 +11,7 @@ namespace OpenFeatureSDK
     /// <seealso href="https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#flag-evaluation-api"/>
     public sealed class OpenFeature
     {
-        private EvaluationContext _evaluationContext = new EvaluationContext();
+        private EvaluationContext _evaluationContext = EvaluationContext.Empty;
         private FeatureProvider _featureProvider = new NoOpFeatureProvider();
         private readonly List<Hook> _hooks = new List<Hook>();
```
toddbaert, line 16: While we're addressing thread safety, do you think we need to worry about the hook list here being mutated in multiple threads?

kinyoklion:
I could look at value types some if we are interested in tackling that now. It may take me a little bit though. I am not as certain how much of a problem that will be.

toddbaert:
> I could look at value types some if we are interested in tackling that now. It may take me a little bit though. I am not as certain how much of a problem that will be.

If it's not a breaking change, my gut feeling is to defer it.

src/OpenFeatureSDK/OpenFeature.cs
kinyoklion: Yeah, that is true. I can take a look at that as well. Likely, to finish up thread safety, we may need some changes more than just the immutable builder.

kinyoklion: Yeah, the hooks are a problem.

Currently I think there is a problem with the global context as well. A threading issue, but also I think right now your client will have a global context based on when you get the client. So if you update the global context it doesn't affect clients you have already gotten from the SDK.

I suspect that is supposed to be like hooks, where adding a hook affect subsequent evaluations from existing clients.

toddbaert: > I think right now your client will have a global context based on when you get the client.

I don't think this is explicitly spelled out in the spec, but it's implied. I think I should add something about it. It doesn't necessarily need to be addressed in this PR.

The same should be true with regards to the configured provider - the currently registered one should be used for all clients, regardless of when they were created.

cc @beeme1mr

kinyoklion: Another question, are hooks supposed to be executed in some specific order based on the order they were registered? There is an order of the sources of hooks, but multiple hooks registered at the API level for instance.

It affects the type of concurrency I would provide here. The C# `ConcurrentStack` has all the correct functionality, but the contents are reversed. (I can enumerable reverse this, so the cost shouldn't be that much.)
 
`ConcurrentQueue` puts things in the right order, but doesn't have a clear method. So you cannot atomically remove everything.

Or just making a synchronized collection, I prefer the concurrent solutions as they are generally lockless.

toddbaert: Personally I didn't want to provide guarantees of hook ordering _within_ a stage. The spec never dictates any behavior there.

> ConcurrentQueue puts things in the right order, but doesn't have a clear method. So you cannot atomically remove everything.

Removing all hooks is also not specified. Most implementations do it b/c it helps with testing, but you could make this a package private method if you want. I don't think there's a real compelling use case for allowing authors to remove hooks, though I could be wrong about that.

toddbaert:
Another point: in Go, Java, and JS, the provider is not a member field in the client. Instead, the currently registered provider is retrieved from the API. From a developer experience perspective, this means that existing clients are "updated" when a new provider is set, which I think is a better experience...?

Again, this isn't in the spec, but perhaps it should be (interested in feedback here). The challenge is that this does introduce more need for locking. In Java I would probably use a read/write lock for this, so n threads could read the provider without contention, and the lock is only unavailable when the provider is updated.

We can also continue to leave this particular case undefined in the spec, with each SDK handling this edge case differently. At a minimum we should document this behavior in each SDK though.

src/OpenFeatureSDK/Model/StructureBuilder.cs
kinyoklion: Removed the Removes.

toddbaert (approved):

kinyoklion:
It seems that GitHub is unhappy with me somehow. Code coverage report not uploading.

toddbaert:
@benjiro any objections to merging this?

benjiro (approved):
LGTM

Issue #56, opened 2022-09-08
Review thread safety of SDK


toddbaert:
As brought up [here](https://github.com/open-feature/dotnet-sdk/pull/55#discussion_r966269026), there's some potential issues with thread safety in the SDK, particularly around state maintained in the API singleton and clients. Let's use this issue to investigate and discuss some of these.

cc @kinyoklion

kinyoklion:
I don't have a strong proposal yet, but I can add a concrete example of where this is a problem.

The evaluation context that is passed to a client is not immutable, so it is possible to pass a reference to the client, and then continue to modify that reference.

This specific issue could be documented. Once an evaluation context has been passed to the client, then you cannot modify it any more. It could also be eliminated through having discrete builders that create immutable contexts, though at the cost of more code to maintain and some stricter rules to maintain. (Structured type data would need to be a value type after being built as well.)

I do think it would be best if there were at least some level of restriction or enforcement of immutability to prevent this. Documentation is great, but code is better, and it surfaces the issue much faster.

If you have a global context, and you are adding and removing things from it based on requests or other state, then it would be pretty easy to get into this situation.

If you modify the context, and concurrently do evaluations, then sometimes it will work, and sometimes it will not work. When it doesn't work you get something like this.
```
System.AggregateException: One or more errors occurred. (Collection was modified; enumeration operation may not execute.)
 ---> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.Dictionary`2.KeyCollection.Enumerator.MoveNext()
   at OpenFeature.SDK.Model.EvaluationContext.Merge(EvaluationContext other) in /home/rlamb/code/open-feature/dotnet-sdk/src/OpenFeature.SDK/Model/EvaluationContext.cs:line 223
   at OpenFeature.SDK.FeatureClient.EvaluateFlag[T](Func`4 resolveValueDelegate, FlagValueType flagValueType, String flagKey, T defaultValue, EvaluationContext context, FlagEvaluationOptions options) in /home/rlamb/code/open-feature/dotnet-sdk/src/OpenFeature.SDK/OpenFeatureClient.cs:line 222
   at OpenFeature.SDK.FeatureClient.GetBooleanDetails(String flagKey, Boolean defaultValue, EvaluationContext context, FlagEvaluationOptions config) in /home/rlamb/code/open-feature/dotnet-sdk/src/OpenFeature.SDK/OpenFeatureClient.cs:line 104
```

[...]

PR #79, opened 2022-10-07
feat!: Thread safe hooks, provider, and context

kinyoklion:
- Change storage type of global hooks to a concurrent stack.
    - Should be able to add and remove from this without locking,
- Protect references used from multiple threads with a Read/Write lock.
- Change read only collection types to immutable collection types.
    - Readonly collections are just views into existing mutable collections. So they are still subject to the same modification problems. Immutable collections are supposed to be fixed and unchanging. Considering they are an interface you could still violate this constraint, but it is less likely.

Related to: https://github.com/open-feature/dotnet-sdk/issues/56

### DIFF
--- a/src/OpenFeatureSDK/FeatureProvider.cs
+++ b/src/OpenFeatureSDK/FeatureProvider.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Threading.Tasks;
 using OpenFeatureSDK.Model;
 
@@ -22,8 +23,8 @@ namespace OpenFeatureSDK
         /// error (if applicable): Provider, Invocation, Client, API
         /// finally: Provider, Invocation, Client, API
         /// </summary>
-        /// <returns></returns>
-        public virtual IReadOnlyList<Hook> GetProviderHooks() => Array.Empty<Hook>();
+        /// <returns>Immutable list of hooks</returns>
+        public virtual IImmutableList<Hook> GetProviderHooks() => ImmutableList<Hook>.Empty;
 
         /// <summary>
         /// Metadata describing the provider.
--- a/src/OpenFeatureSDK/FeatureProvider.cs
+++ b/src/OpenFeatureSDK/FeatureProvider.cs
@@ -1,5 +1,3 @@
-using System;
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
 using OpenFeatureSDK.Model;
--- a/src/OpenFeatureSDK/OpenFeature.cs
+++ b/src/OpenFeatureSDK/OpenFeature.cs
@@ -51,6 +51,13 @@ namespace OpenFeatureSDK
 
         /// <summary>
         /// Gets the feature provider
+        /// <para>
+        /// The feature provider may be set from multiple threads, when accessing the global feature provider
+        /// it should be accessed once for an operation, and then that reference should be used for all dependent
+        /// operations. For instance, during an evaluation the flag resolution method, and the provider hooks
+        /// should be accessed from the same reference, not two independent calls to
+        /// <see cref="OpenFeature.GetProvider"/>.
+        /// </para>
         /// </summary>
         /// <returns><see cref="FeatureProvider"/></returns>
         public FeatureProvider GetProvider()
@@ -68,6 +75,11 @@ namespace OpenFeatureSDK
 
         /// <summary>
         /// Gets providers metadata
+        /// <para>
+        /// This method is not guaranteed to return the same provider instance that may be used during an evaluation
+        /// in the case where the provider may be changed from another thread.
+        /// For multiple dependent provider operations see <see cref="OpenFeature.GetProvider"/>.
+        /// </para>
         /// </summary>
         /// <returns><see cref="ClientMetadata"/></returns>
         public Metadata GetProviderMetadata() => this.GetProvider().GetMetadata();
@@ -131,6 +143,12 @@ namespace OpenFeatureSDK
 
         /// <summary>
         /// Gets the global <see cref="EvaluationContext"/>
+        /// <para>
+        /// The evaluation context may be set from multiple threads, when accessing the global evaluation context
+        /// it should be accessed once for an operation, and then that reference should be used for all dependent
+        /// operations.
+        /// <see cref="OpenFeature.GetProvider"/>.
+        /// </para>
         /// </summary>
         /// <returns>An <see cref="EvaluationContext"/></returns>
         public EvaluationContext GetContext()
--- a/src/OpenFeatureSDK/OpenFeature.cs
+++ b/src/OpenFeatureSDK/OpenFeature.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
--- a/src/OpenFeatureSDK/OpenFeature.cs
+++ b/src/OpenFeatureSDK/OpenFeature.cs
@@ -1,6 +1,7 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Extensions.Logging;
 using OpenFeatureSDK.Model;
 
@@ -13,15 +14,12 @@ namespace OpenFeatureSDK
     /// <seealso href="https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#flag-evaluation-api"/>
     public sealed class OpenFeature
     {
-        /// References are atomic in C#, but it is possible to get an old value due to register caching.
-        /// Making it volatile will result in reads bypassing that cache. It doesn't provide any ordering
-        /// guarantee. Considering this is just a method to replace the value, regardless of the current value,
-        /// this should be sufficient.
-        /// https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/variables#96-atomicity-of-variable-references
-        /// https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile
-        private volatile EvaluationContext _evaluationContext = EvaluationContext.Empty;
-        private volatile FeatureProvider _featureProvider = new NoOpFeatureProvider();
+        private EvaluationContext _evaluationContext = EvaluationContext.Empty;
+        private FeatureProvider _featureProvider = new NoOpFeatureProvider();
         private readonly ConcurrentStack<Hook> _hooks = new ConcurrentStack<Hook>();
+
+        private readonly ReaderWriterLockSlim _evaluationContextLock = new ReaderWriterLockSlim();
+        private readonly ReaderWriterLockSlim _featureProviderLock = new ReaderWriterLockSlim();
 
         /// <summary>
         /// Singleton instance of OpenFeature
@@ -38,19 +36,35 @@ namespace OpenFeatureSDK
         /// Sets the feature provider
         /// </summary>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public void SetProvider(FeatureProvider featureProvider) => this._featureProvider = featureProvider;
+        public void SetProvider(FeatureProvider featureProvider)
+        {
+            this._featureProviderLock.EnterWriteLock();
+            this._featureProvider = featureProvider;
+            this._featureProviderLock.ExitWriteLock();
+        }
 
         /// <summary>
         /// Gets the feature provider
         /// </summary>
         /// <returns><see cref="FeatureProvider"/></returns>
-        public FeatureProvider GetProvider() => this._featureProvider;
+        public FeatureProvider GetProvider()
+        {
+            this._featureProviderLock.EnterReadLock();
+            try
+            {
+                return this._featureProvider;
+            }
+            finally
+            {
+                this._featureProviderLock.ExitReadLock();
+            }
+        }
 
         /// <summary>
         /// Gets providers metadata
         /// </summary>
         /// <returns><see cref="ClientMetadata"/></returns>
-        public Metadata GetProviderMetadata() => this._featureProvider.GetMetadata();
+        public Metadata GetProviderMetadata() => this.GetProvider().GetMetadata();
 
         /// <summary>
         /// Create a new instance of <see cref="FeatureClient"/> using the current provider
@@ -60,7 +74,8 @@ namespace OpenFeatureSDK
         /// <param name="logger">Logger instance used by client</param>
         /// <param name="context">Context given to this client</param>
         /// <returns><see cref="FeatureClient"/></returns>
-        public FeatureClient GetClient(string name = null, string version = null, ILogger logger = null, EvaluationContext context = null) =>
+        public FeatureClient GetClient(string name = null, string version = null, ILogger logger = null,
+            EvaluationContext context = null) =>
             new FeatureClient(name, version, logger, context);
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -53,25 +61,30 @@ public sealed class OpenFeature
...
+        /// <para>
+        /// The items enumerated will reflect the registered hooks
+        /// at the start of enumeration. Hooks added during enumeration
+        /// will not be included.
+        /// </para>
         /// </summary>
-        /// <returns>A immutable list of <see cref="Hook"/></returns>
-        public IReadOnlyList<Hook> GetHooks() => this._hooks.AsReadOnly();
+        /// <returns>Enumeration of <see cref="Hook"/></returns>
+        public IEnumerable<Hook> GetHooks() => this._hooks.Reverse();
```
kinyoklion, line 87: In the open feature API and Client, the SDK controls the implementation type and can promise the use of a concurrent collection.

In the provider this is not the case, so that is constrained to an immutable collection.

kinyoklion: Also, I am reversing these for now. I wasn't completely comfortable with all the test changes required for them to be in the opposite order.

src/OpenFeatureSDK/FeatureProvider.cs
```diff
@@ -22,8 +21,8 @@ public abstract class FeatureProvider
         /// error (if applicable): Provider, Invocation, Client, API
         /// finally: Provider, Invocation, Client, API
         /// </summary>
-        /// <returns></returns>
-        public virtual IReadOnlyList<Hook> GetProviderHooks() => Array.Empty<Hook>();
+        /// <returns>Immutable list of hooks</returns>
+        public virtual IImmutableList<Hook> GetProviderHooks() => ImmutableList<Hook>.Empty;
```
kinyoklion, line 25: Alternatively this could be enumerable like other places, with documentation that the provider must use a thread-safe implementation. Technically a readonlylist, with the same constraint, would also works. I think immutable directs people in a way that the easiest solution is also the correct solution.

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -225,9 +228,9 @@ public FeatureClient(FeatureProvider featureProvider, string name, string versio
 
             var allHooks = new List<Hook>()
                 .Concat(OpenFeature.Instance.GetHooks())
-                .Concat(this._hooks)
+                .Concat(this.GetHooks())
                 .Concat(options?.Hooks ?? Enumerable.Empty<Hook>())
-                .Concat(this._featureProvider.GetProviderHooks())
+                .Concat(OpenFeature.Instance.GetProvider().GetProviderHooks())
```
kinyoklion, line 233: Ok. This isn't correct. You would't want the hooks from a different provider than the evaluation method. Those two things need to be resolved at the same time.

kinyoklion: I think that we need to get the reference, then use that reference to get the method and the hooks. We would not just want to get them in the call, because they could still be separated in time.

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -17,11 +18,32 @@ namespace OpenFeatureSDK
     public sealed class FeatureClient : IFeatureClient
     {
         private readonly ClientMetadata _metadata;
-        private readonly FeatureProvider _featureProvider;
-        private readonly List<Hook> _hooks = new List<Hook>();
+        private readonly ConcurrentStack<Hook> _hooks = new ConcurrentStack<Hook>();
         private readonly ILogger _logger;
         private EvaluationContext _evaluationContext;
 
+        /// <summary>
```
kinyoklion, line 25: This is a little cumbersome. I am not sure if it is more or less cumbersome than the alternative.
The alternative would be to convert all the methods from expression bodies to normal function bodies, then alias the provider and method, then pass those to evaluate.

src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -30,19 +36,41 @@ public sealed class OpenFeature
         /// Sets the feature provider
         /// </summary>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public void SetProvider(FeatureProvider featureProvider) => this._featureProvider = featureProvider;
+        public void SetProvider(FeatureProvider featureProvider)
+        {
+            this._featureProviderLock.EnterWriteLock();
+            try
```
kinyoklion, line 42: None of these code blocks can throw, but when using this type of lock I think using try/finally is a customary way to release the lock.

src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -30,19 +36,53 @@ public sealed class OpenFeature
...
 
         /// <summary>
         /// Gets the feature provider
+        /// <para>
+        /// The feature provider may be set from multiple threads, when accessing the global feature provider
+        /// it should be accessed once for an operation, and then that reference should be used for all dependent
+        /// operations. For instance, during an evaluation the flag resolution method, and the provider hooks
+        /// should be accessed from the same reference, not two independent calls to
+        /// <see cref="OpenFeature.GetProvider"/>.
+        /// </para>
```
kinyoklion, line 60: One thing I am not calling out here is that setting hooks, provider, and context are not guaranteed to be ordered among themselves. If you have one thread setting all 3, and another thread doing evaluations, then a single evaluation may use some combination of the values being set.

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -241,7 +273,7 @@ public FeatureClient(FeatureProvider featureProvider, string name, string versio
                 flagKey,
                 defaultValue,
                 flagValueType, this._metadata,
-                OpenFeature.Instance.GetProviderMetadata(),
+                provider.GetMetadata(),
```
kinyoklion, line 276: We don't want to call `GetProviderMetadata` here on the `OpenFeature` singleton, because it could have change between when we got the resolution method and here. We want to be sure to only use the cached reference.

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -36,15 +58,13 @@ public sealed class FeatureClient : IFeatureClient
         /// <summary>
         /// Initializes a new instance of the <see cref="FeatureClient"/> class.
         /// </summary>
-        /// <param name="featureProvider">Feature provider used by client <see cref="FeatureProvider"/></param>
         /// <param name="name">Name of client <see cref="ClientMetadata"/></param>
         /// <param name="version">Version of client <see cref="ClientMetadata"/></param>
         /// <param name="logger">Logger used by client</param>
         /// <param name="context">Context given to this client</param>
         /// <exception cref="ArgumentNullException">Throws if any of the required parameters are null</exception>
-        public FeatureClient(FeatureProvider featureProvider, string name, string version, ILogger logger = null, EvaluationContext context = null)
```
kinyoklion, line 45: Just realized I need to re-introduce the equivalent of this check elsewhere.

src/OpenFeatureSDK/FeatureProvider.cs
toddbaert: > I think immutable directs people in a way that the easiest solution is also the correct solution.

Agreed.

src/OpenFeatureSDK/OpenFeature.cs
toddbaert: Yep, agreed. In fact, I made a little lock-wrapper [class](https://github.com/open-feature/java-sdk/pull/131/files#diff-f7f996e84ae9d77890aa5aa0cde5a7e3b08401b43173268bc134083536ea26dc) that allows us to use a try-with-resources block in Java (I think c# has an equivalent with the `using` keyword and `IDisposable` implementations).

src/OpenFeatureSDK/OpenFeature.cs
toddbaert: > One thing I am not calling out here is that setting hooks, provider, and context are not guaranteed to be ordered among themselves. If you have one thread setting all 3, and another thread doing evaluations, then a single evaluation may use some combination of the values being set.

That makes sense, and I think mentioning this might be adding cognitive overload with not-too-much value returned.

As for the block comment, I had added something similar in the read-lock block in Java, so it makes sense to me and I think it's worth pointing out.

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -17,11 +18,38 @@ namespace OpenFeatureSDK
...
+                Func<FeatureProvider, Func<string, T, EvaluationContext, Task<ResolutionDetails<T>>>> method)
+        {
+            // Alias the provider reference so getting the method and returning the provider are
+            // guaranteed to be the same object.
+            var provider = OpenFeature.Instance.GetProvider();
+
+            if (provider == null)
+            {
+                provider = new NoOpFeatureProvider();
+                this._logger.LogDebug("No provider configured, using no-op provider");
```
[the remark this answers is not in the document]
toddbaert: I think No-op is a valid configuration, and may be used intentionally, especially if OpenFeature is implemented in 3rd party libraries.

For this reason, I think this should remain a `debug` level log.

src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -30,19 +36,53 @@ public sealed class OpenFeature
...
+            this._featureProviderLock.EnterReadLock();
+            try
+            {
+                return this._featureProvider;
+            }
+            finally
+            {
+                this._featureProviderLock.ExitReadLock();
+            }
+        }
```
toddbaert, line 74: Is it necessary to lock this? I think as long as we lock during it's retrieval in the evaluation, that's sufficient, right?

src/OpenFeatureSDK/OpenFeature.cs
```diff
@@ -81,13 +127,41 @@ public sealed class OpenFeature
...
+            this._evaluationContextLock.EnterReadLock();
+            try
+            {
+                return this._evaluationContext;
+            }
+            finally
+            {
+                this._evaluationContextLock.ExitReadLock();
+            }
+        }
```
toddbaert, line 165: Same as: https://github.com/open-feature/dotnet-sdk/pull/79/files#r992477160

src/OpenFeatureSDK/OpenFeatureClient.cs
```diff
@@ -17,11 +18,38 @@ namespace OpenFeatureSDK
...
+        /// </summary>
+        /// <param name="method">
+        ///     This method should return the desired flag resolution method from the given provider reference.
+        /// </param>
+        /// <typeparam name="T">The type of the resolution method</typeparam>
+        /// <returns>A tuple containing a resolution method and the provider it came from.</returns>
+        private (Func<string, T, EvaluationContext, Task<ResolutionDetails<T>>>, FeatureProvider)
+            ExtractProvider<T>(
+                Func<FeatureProvider, Func<string, T, EvaluationContext, Task<ResolutionDetails<T>>>> method)
+        {
```
toddbaert, line 40: It may be worth read-locking during this operation, and include the hooks and context here. The way this is implemented right now, evaluation context and hooks can change during the process of evaluation, though you've handled the provider.

src/OpenFeatureSDK/OpenFeature.cs
kinyoklion: So, here each of the 3 things are individually thread safe. 

The provider will use a stable reference, which covers the provider hooks and other data.

The context is immutable, and we get a reference at a specific point in time. The reference globally may change, but we are decoupled from that because we got a reference. Locking for the duration of getting the context and the provider wouldn't prevent interleaved setting of them. You could set one, and then the evaluation gets the lock, you have to wait, and then set the other. 

The last bit are the other two sets of hooks. These are interesting. There isn't anything which inherently couples them to anything else. They are copied, so the list will be stable in the places they are iterate, but up until they are iterated, they could change.

If you have cross dependent hooks then they should be added with AddHooks, which will add them atomically. I should document that.

kinyoklion: I added a couple comments to the hook methods.

src/OpenFeatureSDK/OpenFeature.cs
kinyoklion: Especially for double locks, where they could nest the using statement.

src/OpenFeatureSDK/OpenFeature.cs
toddbaert: > If you have cross dependent hooks

Worth saying I generally think this should be avoided, but I think people might do it.

toddbaert: > Locking for the duration of getting the context and the provider wouldn't prevent interleaved setting of them.

Wouldn't read-locking in the evaluation on a shared lock which is write-locked by both the provider and context mutators address exactly this? Or am I misunderstanding?

kinyoklion: Not unless you also locked for the duration of setting both of them.

If they can be independently set, as they can in the implementation at the moment, then it doesn't.

```
Thread 1:
SetProvider(provider)
  writeLock(_provider)
// Evaluation could start here
SetContext(context)
  writeLock(_context)

Thread 2:
Evaluate()
  readLock(_provider)
    readLock(_context)
```

So, given this code someone could intend to set the provider and context, but it would be possible for evaluation to grab the lock before thread 1 set the context, but after it set the provider.

In order to not interleave it would need to be more like

```
Thread 1:
writeLock(_provider)
  writeLock(_context)
    setProvider
    setContext
    
Thread 2:
Evaluate()
  readLock(_provider)
    readLock(_context)
```

Or use a single lock

src/OpenFeatureSDK/OpenFeature.cs
kinyoklion: I think there is another consideration here, in that the getters are public, so they maybe should still have ordering guarantees. 

I think there is a way to allow re-entrant read locks though, so if we did need to lock for multiple operations, then we could change to the re-entrant form. Allowing acquiring read locks inside read locks.

kinyoklion: https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?view=net-6.0#entering-the-lock-recursively

src/OpenFeatureSDK/OpenFeature.cs
toddbaert: Ahh, yes, I agree. I misunderstood your initial statement.

kinyoklion:
I realized that I missed the client level context locking. So I have added that.

kinyoklion:
I've noticed another thing that needs to be resolved. The ReaderWriterLock type is disposable, so the containing type should implement disposable. Not really a problem for the global OpenFeature instance, because it isn't going anywhere. It is inconvenient for the client instances to be disposable though.

So I may want to consider a normal lock in that case.

benjiro (approved):
LGTM 👍🏻

kinyoklion:
@toddbaert @benjiro I switched the OpenFeatureClient locks to normal locks. This does mean if you are using one client from many threads, that it will have more contention. Maybe... make more clients.

The alternative would be to make it disposable, but that then becomes like a, "what color are your functions" kind of situation.

In OpenFeature.cs I am not disposing. I added a comment. The singleton isn't meant to go away, so disposing it doesn't make sense.

toddbaert:
I know there's been a few changes on this, but I think it's this is good work and certainly more threadsafe than what we had before. I'm in favor of merging it unless there's objections.

PR #85, opened 2022-10-18
chore: release 1.0.0

toddbaert:
Release-As: 1.0.0


toddbaert:
Merging this comment (!without any change to the body!) will cause Release Please to open a 1.0 release.

beeme1mr (approved):
Merging this just opens a proper release PR.

weyert (approved):
I think its looking great.

Only suggestion I would help whether it would be worthwhile to add a logging hook as an example?

toddbaert:
> I think its looking great.
> 
> Only suggestion I would help whether it would be worthwhile to add a logging hook as an example?

Good call out.

I actually have a separate PR that adds a link to the doc where we have extensive code examples and explanations of hooks:
https://github.com/open-feature/dotnet-sdk/pull/86

https://docs.openfeature.dev/docs/reference/concepts/hooks

kinyoklion (approved):
Things look good to me. I am not as familiar with the documentation or other procedural elements, but I don't have any major code concerns at this point,

benjiro (approved):

toddbaert:
Merging this. We can have any required discussion about what follow ups we want in the subsequent release PR that will get created.

PR #96, opened 2022-11-10
chore: Update README basic usage.

kinyoklion:
Signed-off-by: Ryan Lamb <4955475+kinyoklion@users.noreply.github.com>

## This PR
Updates the README with a working basic usage.

### Notes

The NoOpProvider was used in the usage example, but now it is internal, so we either need to not use it, as done here, or we should consider making it non-internal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,14 @@ The packages will aim to support all current .NET versions. Refer to the current
 ### Basic Usage
 
 ```csharp
-using OpenFeature;
+using OpenFeature.Model;
 
 // Sets the provider used by the client
-Api.Instance.SetProvider(new NoOpProvider());
+// If no provider is set, then a default NoOpProvider will be used.
+//OpenFeature.Api.Instance.SetProvider(new MyProvider());
+
 // Gets a instance of the feature flag client
-var client = OpenFeature.Instance.GetClient();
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,12 +17,14 @@ The packages will aim to support all current .NET versions. Refer to the current
 ### Basic Usage
 
 ```csharp
-using OpenFeature;
+using OpenFeature.Model;
```
kinyoklion, line 20: Needed for EvaluationContext.

Removed OpenFeature, as we use them explicitly. `OpenFeature.Api...`

README.md
```diff
@@ -17,12 +17,14 @@ The packages will aim to support all current .NET versions. Refer to the current
 ### Basic Usage
 
 ```csharp
-using OpenFeature;
+using OpenFeature.Model;
 
 // Sets the provider used by the client
-Api.Instance.SetProvider(new NoOpProvider());
+// If no provider is set, then a default NoOpProvider will be used.
+//OpenFeature.Api.Instance.SetProvider(new MyProvider());
```
kinyoklion, line 24: Maybe useful to see what it would look like?

benjiro (approved):

PR #97, opened 2022-11-10
feat: include net7 in the test suit

benjiro:
Since .net 7 was released this week, we should also run our test suite across it


toddbaert (approved):

toddbaert:
We're on the bleeding edge.

PR #102, opened 2023-01-13
feat: Make IFeatureClient interface public.

cdonnellytx:
Signed-off-by: Chris Donnelly <cdonnellytx@users.noreply.github.com>

## This PR

- Exposes the `IFeatureClient` interface to consumers by making it public.
- Pulls up the XMLDoc from the concrete class to the interface, filling in gaps (e.g., the interface-level docs) from the Java SDK.

### Background and Motivation

This allows consumers using dependency injection to properly create a mock of the feature client.  Currently, the exposed `FeatureClient` class is sealed, therefore making it extremely difficult ([though not impossible](https://stackoverflow.com/a/12187583)) for consumers to mock in parallel without side-effects.

Consider

```csharp
public class MyService
{
    private readonly FeatureClient _featureClient;

    public MyService(FeatureClient featureClient, ...)
    {
        _featureClient = featureClient;
    }

    public async Task<int> DoSomethingAsync() => await _featureClient.GetBooleanValue("flag") ? 42 : 47;
}
```

A unit test for `MyService` would be dependent on the concrete `FeatureClient` class, and thus on the `Api.Instance` singleton, limiting tests to be more brittle and dependent on state.

Allowing `MyService` to use `IFeatureClient` would get around this problem, and allow for unit tests to not depend on shared state, allowing for test parallelization where possible.

### Notes

Modifying a public interface in any way (even adding new methods/properties) is considered a breaking change, so each client surface enhancement/change would be breaking by that definition.  We may wish to pull up methods from the concrete class to the interface (e.g., `GetContext` / `SetContext`) now before committing to the existing interface. 

### Follow-up Tasks

Avoiding unrelated changes is also why I opted not to make `GetMetadata()` into a `Metadata` read-only property on the interface -- while this makes sense from a .NET idiomatic view, it would conflict with the current way the concrete class exposes `GetMetadata()`.  Fixing it later would be a breaking change, though.

Similarly, I opted to keep `Api.Instance.GetClient()` returning the concrete class, as it would be a breaking change to make it return `IFeatureClient`.  I am open to that idea, though.

### How to test

No additional tests should be necessary at this time.

### DIFF
--- a/src/OpenFeature/IFeatureClient.cs
+++ b/src/OpenFeature/IFeatureClient.cs
@@ -4,24 +4,124 @@ using OpenFeature.Model;
 
 namespace OpenFeature
 {
-    internal interface IFeatureClient
+    /// <summary>
+    /// Interface used to resolve flags of varying types.
+    /// </summary>
+    public interface IFeatureClient
     {
+        /// <summary>
+        /// Appends hooks to client
+        /// <para>
+        /// The appending operation will be atomic.
+        /// </para>
+        /// </summary>
+        /// <param name="hooks">A list of Hooks that implement the <see cref="Hook"/> interface</param>
         void AddHooks(IEnumerable<Hook> hooks);
+
+        /// <summary>
+        /// Gets client metadata
+        /// </summary>
+        /// <returns>Client metadata <see cref="ClientMetadata"/></returns>
         ClientMetadata GetMetadata();
 
+        /// <summary>
+        /// Resolves a boolean feature flag
+        /// </summary>
+        /// <param name="flagKey">Feature flag key</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <param name="context"><see cref="EvaluationContext">Evaluation Context</see></param>
+        /// <param name="config"><see cref="EvaluationContext">Flag Evaluation Options</see></param>
+        /// <returns>Resolved flag details <see cref="FlagEvaluationDetails{T}"/></returns>
         Task<bool> GetBooleanValue(string flagKey, bool defaultValue, EvaluationContext context = null, FlagEvaluationOptions config = null);
+
+        /// <summary>
+        /// Resolves a boolean feature flag
+        /// </summary>
+        /// <param name="flagKey">Feature flag key</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <param name="context"><see cref="EvaluationContext">Evaluation Context</see></param>
+        /// <param name="config"><see cref="EvaluationContext">Flag Evaluation Options</see></param>
+        /// <returns>Resolved flag details <see cref="FlagEvaluationDetails{T}"/></returns>
         Task<FlagEvaluationDetails<bool>> GetBooleanDetails(string flagKey, bool defaultValue, EvaluationContext context = null, FlagEvaluationOptions config = null);
 
+        /// <summary>
+        /// Resolves a string feature flag
+        /// </summary>
+        /// <param name="flagKey">Feature flag key</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <param name="context"><see cref="EvaluationContext">Evaluation Context</see></param>
+        /// <param name="config"><see cref="EvaluationContext">Flag Evaluation Options</see></param>
+        /// <returns>Resolved flag details <see cref="FlagEvaluationDetails{T}"/></returns>
         Task<string> GetStringValue(string flagKey, string defaultValue, EvaluationContext context = null, FlagEvaluationOptions config = null);
+
+        /// <summary>
+        /// Resolves a string feature flag
+        /// </summary>
+        /// <param name="flagKey">Feature flag key</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <param name="context"><see cref="EvaluationContext">Evaluation Context</see></param>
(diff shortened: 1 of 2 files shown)


src/OpenFeature/IFeatureClient.cs
```diff
@@ -4,24 +4,124 @@
...
         ClientMetadata GetMetadata();
 
+        /// <summary>
+        /// Resolves a boolean feature flag
+        /// </summary>
+        /// <param name="flagKey">Feature flag key</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <param name="context"><see cref="EvaluationContext">Evaluation Context</see></param>
+        /// <param name="config"><see cref="EvaluationContext">Flag Evaluation Options</see></param>
+        /// <returns>Resolved flag details <see cref="FlagEvaluationDetails{T}"/></returns>
```
toddbaert, line 34: This wasn't something you introduced, @cdonnellytx, so I don't mind if you address it or not, but these docs are _slightly_ innacurate. This just resolves the flag _value_ not wrapped in the `FlagEvaluationDetails` wrapper.

toddbaert (approved):
Thanks for your contribution and well-reasoned PR, @cdonnellytx ! I agree with the overall goal and solution.

I would like input from @kinyoklion and @benjiro on this point:

> Modifying a public interface in any way (even adding new methods/properties) is considered a breaking change, so each client surface enhancement/change would be breaking by that definition. We may wish to pull up methods from the concrete class to the interface (e.g., GetContext / SetContext) now before committing to the existing interface.

I'm leaning towards adding more methods to the interface, so they could be easily mocked as well.

src/OpenFeature/IFeatureClient.cs
cdonnellytx: I can fix it; do you want me to do it in a separate PR or in this one?

cdonnellytx:
> I'm leaning towards adding more methods to the interface, so they could be easily mocked as well.

I would recommend doing so.  At a minimum I would recommend what's exposed on [the Java interface](https://github.com/open-feature/java-sdk/blob/21929328630eba00be741392457f68bacf59f376/src/main/java/dev/openfeature/sdk/Client.java):

- `GetContext` / `SetContext` -- either as-is or as a property
- `GetHooks`

(An aside: if you do want to convert the getter/setter methods to properties, backward compatibility on the class itself can be preserved.)

src/OpenFeature/IFeatureClient.cs
toddbaert: I'll leave that up to you, I'll review/approve either way.

kinyoklion:
> > I'm leaning towards adding more methods to the interface, so they could be easily mocked as well.
> 
> I would recommend doing so. At a minimum I would recommend what's exposed on [the Java interface](https://github.com/open-feature/java-sdk/blob/21929328630eba00be741392457f68bacf59f376/src/main/java/dev/openfeature/sdk/Client.java):
> 
> * `GetContext` / `SetContext` -- either as-is or as a property
> * `GetHooks`
> 
> (An aside: if you do want to convert the getter/setter methods to properties, backward compatibility on the class itself can be preserved.

I think the adding the interface itself is a great change.

These methods I have some concerns with. Primarily I don't think they should have been public to begin with on the concrete type (probably should have been internal). They aren't required for any functionality an OF client is intended to provide, and don't seem to be part of the spec. 

I guess adding them to the interface doesn't worsen that problem, and being as they had been exposed they may as well also be added to the interface.

src/OpenFeature/IFeatureClient.cs
cdonnellytx: OK, I have updated the docs.

cdonnellytx:
> I guess adding them to the interface doesn't worsen that problem, and being as they had been exposed they may as well also be added to the interface.

I agree.  

I'll go ahead and add them to the PR.

toddbaert:
> Primarily I don't think they should have been public to begin with on the concrete type (probably should have been internal). They aren't required for any functionality an OF client is intended to provide, and don't seem to be part of the spec.

I think this is true for `GetHooks`/`GetContext`, but not for `SetContext`. The spec does mention in [3.2.1](https://github.com/open-feature/spec/blob/main/specification/sections/03-evaluation-context.md#requirement-321) that is means of providing context per client is required. It doesn't need to be implemented with a mutator though.

kinyoklion:
> > Primarily I don't think they should have been public to begin with on the concrete type (probably should have been internal). They aren't required for any functionality an OF client is intended to provide, and don't seem to be part of the spec.
> 
> I think this is true for `GetHooks`/`GetContext`, but not for `SetContext`. The spec does mention in [3.2.1](https://github.com/open-feature/spec/blob/main/specification/sections/03-evaluation-context.md?rgh-link-date=2023-01-17T19%3A27%3A37Z#requirement-321) that is means of providing context per client is required. It doesn't need to be implemented with a mutator though.
> 
> One justification is that we wanted to support a "client-per-request" paradigm, so that you could basically associate a particular thread or continuation with a state, for instance, session data.

Yes, sorry, The set should be exposed for sure.

kinyoklion (approved):

toddbaert:
I plan on merging this and releasing it with 1.1 tomorrow, unless I hear objections.

benjiro (approved):
Thanks for your contribution @cdonnellytx

PR #108, opened 2023-01-18
docs: inheritdoc at class level

say25:
Signed-off-by: Steven Yeh <say25@cornell.edu>

## This PR

- Moves inheritdoc to the class level so it doesn't need to be done on every method.

### Related Issues

#102

### How to test

- Docs only, I guess validate that the docs are still on all the methods?

### DIFF
--- a/src/OpenFeature/OpenFeatureClient.cs
+++ b/src/OpenFeature/OpenFeatureClient.cs
@@ -12,6 +12,7 @@ using OpenFeature.Model;
 
 namespace OpenFeature
 {
+    /// <inheritdoc />
     /// <summary>
     ///
     /// </summary>
@@ -53,7 +54,6 @@ namespace OpenFeature
             return (method(provider), provider);
         }
 
-        /// <inheritdoc />
         public EvaluationContext GetContext()
         {
             lock (this._evaluationContextLock)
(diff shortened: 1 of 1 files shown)


toddbaert:
I really like this in theory, but https://github.com/dotnet/roslyn/issues/42392 seems to make this impractical at the moment. I can't seem to find a way to make the compiler or my IDE not complain about this.

@say25 if you can find a solution, please let me know, but until then I think having `inheritdoc` on every member is better than disabling the check.

say25:
> I really like this in theory, but [dotnet/roslyn#42392](https://github.com/dotnet/roslyn/issues/42392) seems to make this impractical at the moment. I can't seem to find a way to make the compiler or my IDE not complain about this.
> 
> @say25 if you can find a solution, please let me know, but until then I think having `inheritdoc` on every member is better than disabling the check.

Yea I'm going to just close this, @cdonnellytx and I work together and he pointed out this issue to me as well. Seems that doc generation is impacted in this way

toddbaert:
Thanks, regardless!

toddbaert:
@say25 If you're looking for another way to contribute, I can think of 2 options (no pressure at all, of course):

- create a summary of things that need doing to bring this repo more in-line with the org's [repo-reqs](https://github.com/open-feature/.github/blob/main/CONTRIBUTING.md#repository-requirements). See an example [for another repo ](https://github.com/open-feature/.github/issues/24).

- if the above seems uninteresting, you might be interested in simply taking on one of these issues without creating a summary... for example, implementing SBOM generation via [CycloneDX](https://github.com/CycloneDX/cyclonedx-dotnet).

say25:
I'll keep that in mind, I think at the moment this project is satisfying our needs

PR #109, opened 2023-01-20
docs: fix broken links

cdonnellytx:
Signed-off-by: Chris Donnelly [cdonnellytx@users.noreply.github.com](mailto:cdonnellytx@users.noreply.github.com)

## This PR

- Fixes broken spec links in the XML documentation
- Fixes spec references in the tests (both code and description)
- Adds one additional test to cover 4.4.5 "after" phase

### Related Issues

Partial fix for #84 (skips v0.5.1, goes to v0.5.2).

### Notes

I started this because I found the `see` links were broken.  Then I realized I should probably fix the specification attribute numbers.  Then I added one test.  Stopping there :)

### Follow-up Tasks

These are the remaining spec items that I saw that do not have tests labeled:

- **1.1.6** The client creation function MUST NOT throw, or otherwise abnormally terminate.
- **1.4.12** In cases of abnormal execution, the `evaluation details` structure's `error message` field MAY contain a string containing additional details about the nature of the error.
- **3.2.1** The API, Client and invocation MUST have a method for supplying `evaluation context`.
- **4.3.2** The `before` stage MUST run before flag resolution occurs. It accepts a `hook context` (required) and `hook hints` (optional) as parameters and returns either an `evaluation context` or nothing.

Additionally there is no "after" equivalent for 4.4.7, which may or may not be a concern.  (It currently returns the default value, instead of the evaluated value.)

### DIFF
--- a/test/OpenFeature.Tests/OpenFeatureHookTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureHookTests.cs
@@ -551,6 +551,7 @@ namespace OpenFeature.Tests
         }
 
         [Fact]
+        [Specification("4.4.5", "If an error occurs in the `before` or `after` hooks, the `error` hooks MUST be invoked.")]
         [Specification("4.4.7", "If an error occurs in the `before` hooks, the default value MUST be returned.")]
         public async Task When_Error_Occurs_In_Before_Hook_Should_Return_Default_Value()
         {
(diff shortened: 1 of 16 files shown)


benjiro (approved):

toddbaert:
> Additionally there is no "after" equivalent for 4.4.7, which may or may not be a concern. (It currently returns the default value, instead of the evaluated value.)

I think this is a good point @cdonnellytx ... I don't know if the behavior here is divergent between SDKs or not. I think we should specify if the evaluated or default value is returned. I will investigate this...

Personally, I think since "after" are really about side-effects, like telemetry, etc, we probably should return the evaluated value... but I'm not sure. I'm interested in your opinion.

test/OpenFeature.Tests/OpenFeatureHookTests.cs
```diff
@@ -582,5 +584,57 @@ public async Task When_Error_Occurs_In_Before_Hook_Should_Return_Default_Value()
             hook.Verify(x => x.Error(It.IsAny<HookContext<It.IsAnyType>>(), exceptionToThrow, null), Times.Once);
             hook.Verify(x => x.Finally(It.IsAny<HookContext<It.IsAnyType>>(), null), Times.Once);
         }
+
+        [Fact]
+        [Specification("4.4.5", "If an error occurs in the `before` or `after` hooks, the `error` hooks MUST be invoked.")]
+        public async Task When_Error_Occurs_In_After_Hook_Should_Invoke_Error_Hook()
```
toddbaert, line 590: Nice catch!

toddbaert (approved):
I crossed checked with the spec, and the numbers/links seem good.

I also opened [this](https://github.com/open-feature/spec/issues/174).

Thanks again @cdonnellytx !

PS: I've sent you an org invite if you're interested!

PR #114, opened 2023-02-07
feat: split errors

### LINKED ISSUE #110: [FEATURE] implement standard error classes
### Requirements

We already have [error codes](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Constant/ErrorType.cs) defined, but it would be nice to define error classes to make things a bit more ergonomic for contributors. Basically something like is already implemented in [this contrib](https://github.com/open-feature/dotnet-sdk-contrib/tree/main/src/OpenFeature.Contrib.Providers.GOFeatureFlag/exception) ; simple classes wrapping these error types that can easily be thrown.

valentk777:
Signed-off-by: Valentinas Kaminskas <valentk777@gmail.com>

## This PR
In this PR, I created a separate class for each error type. 

Fixes #110

### DIFF
--- a/src/OpenFeature/Error/FlagNotFoundException.cs
+++ b/src/OpenFeature/Error/FlagNotFoundException.cs
@@ -1,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Provider was unable to find the flag error when evaluating a flag.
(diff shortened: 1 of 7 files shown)


valentk777:
I created PR, but I do not know what is wrong with this one :D 
image
image
I re-created PR multiple times but this validation still not pass

PR #115, opened 2023-02-07
feat: split errors to classes by types

### LINKED ISSUE #110: [FEATURE] implement standard error classes

valentk777:
Signed-off-by: Valentinas <31697821+valentk777@users.noreply.github.com>

In this PR, I created a separate class for each error type.

Fixes https://github.com/open-feature/dotnet-sdk/issues/110

### DIFF
--- a/src/OpenFeature/Error/FlagNotFoundException.cs
+++ b/src/OpenFeature/Error/FlagNotFoundException.cs
@@ -5,7 +5,6 @@ namespace OpenFeature.Error
 {
     /// <summary>
     /// Provider was unable to find the flag error when evaluating a flag.
-    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
     /// </summary>
     public class FlagNotFoundException : FeatureProviderException
     {
--- a/src/OpenFeature/Error/FlagNotFoundException.cs
+++ b/src/OpenFeature/Error/FlagNotFoundException.cs
@@ -1,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Provider was unable to find the flag error when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
+    /// </summary>
+    public class FlagNotFoundException : FeatureProviderException
+    {
+        /// <summary>
+        /// Initialize a new instance of the <see cref="FlagNotFoundException"/> class
+        /// </summary>
+        /// <param name="message">Exception message</param>
+        /// <param name="innerException">Optional inner exception</param>
+        public FlagNotFoundException(string message = null, Exception innerException = null)
+            : base(ErrorType.FlagNotFound, message, innerException)
+        {
+        }
+    }
+}
--- a/src/OpenFeature/Error/GeneralException.cs
+++ b/src/OpenFeature/Error/GeneralException.cs
@@ -5,7 +5,6 @@ namespace OpenFeature.Error
 {
     /// <summary>
     /// Abnormal execution of the provider when evaluating a flag.
-    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
     /// </summary>
     public class GeneralException : FeatureProviderException
     {
--- a/src/OpenFeature/Error/GeneralException.cs
+++ b/src/OpenFeature/Error/GeneralException.cs
@@ -1,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Abnormal execution of the provider when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
+    /// </summary>
+    public class GeneralException : FeatureProviderException
+    {
+        /// <summary>
+        /// Initialize a new instance of the <see cref="GeneralException"/> class
+        /// </summary>
+        /// <param name="message">Exception message</param>
+        /// <param name="innerException">Optional inner exception</param>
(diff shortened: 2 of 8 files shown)


src/OpenFeature/Error/FlagNotFoundException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Provider was unable to find the flag error when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/GeneralException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Abnormal execution of the provider when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/InvalidContextException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Context does not satisfy provider requirements when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/ParseErrorException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Provider failed to parse the flag response when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/ProviderNotReadyException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Provider has yet been initialized when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/TargetingKeyMissingException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Context does not contain a targeting key and the provider requires one when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

src/OpenFeature/Error/TypeMismatchException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using OpenFeature.Constant;
+
+namespace OpenFeature.Error
+{
+    /// <summary>
+    /// Request type does not match the expected type when evaluating a flag.
+    /// This exception should be thrown when evaluating a flag inside a IFeatureFlag provider
```
beeme1mr, line 8:
```suggestion
```

beeme1mr (approved):
Looks good to me. I would just recommend that you remove the section line in each summary.

toddbaert (approved):
codecov has failed, but I don't think coverage of this sort of change is valuable.

Thanks @valentk777 . I agree with @beeme1mr suggestions, but otherwise looks good!

src/OpenFeature/Error/FlagNotFoundException.cs
kinyoklion: How would people feel about something aside from "should"? Should makes it sound like you should throw these exceptions, when really you can/could, but I wouldn't personally encourage it. The SDK also supports returning error codes, and the performance (and application debug-ability)  of that will be better than an exception. (Which probably matters more to FlagNotFound, evaluating a flag that doesn't exist yet, or has been deleted, isn't an exceptional situation. Versus an invalid context, which is more of a programming error, and therefore less performance critical.)

valentk777: removed. Thanks for the comments. I leave this sentence because it was mentioned in FeatureProviderException. And all these exceptions derive from that base (having this commented line here).

valentk777:
Thank you for your reviews. Code updated.

src/OpenFeature/Error/FlagNotFoundException.cs
toddbaert: @valentk777 has removed the line entirely, which I think is the right move. I agree the "should" seemed like more of a directive than we'd want.

kinyoklion (approved):

PR #116, opened 2023-02-08
test: add initial api benchmarks

### LINKED ISSUE #22: Benchmark Suite
Use https://benchmarkdotnet.org/ to make a suite of performance benchmarks to ensure OpenFeature.SDK has little overhead as possible.

valentk777:
Signed-off-by: Valentinas <31697821+valentk777@users.noreply.github.com>

Added initial API benchmark (only for a few functions). Later we can increase the functions count by covering hooks as well. 

## This PR
Adds initial project to run benchmarks for API part.

Fixes #22

### Notes
I added the result table here. 

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,9 @@ only and does not affect the actual build of the project.
 ## Benchmarking
 
 We use [BenchmarkDotNet](https://benchmarkdotnet.org/articles/overview.html) NuGet package to benchmark a code.
-To run pipelines locally, you can procede with a following command in a root project directory.
+
+To run pipelines locally, you can follow these commands from a root project directory.
+
 ```
 dotnet restore
 dotnet build --configuration Release --output "./release" --no-restore
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,3 +119,13 @@ As with other OpenFeature SDKs, dotnet-sdk follows the
 This project includes a [`.editorconfig`](./.editorconfig) file which is
 supported by all the IDEs/editor mentioned above. It works with the IDE/editor
 only and does not affect the actual build of the project.
+
+## Benchmarking
+
+We use [BenchmarkDotNet](https://benchmarkdotnet.org/articles/overview.html) NuGet package to benchmark a code.
+To run pipelines locally, you can procede with a following command in a root project directory.
+```
+dotnet restore
+dotnet build --configuration Release --output "./release" --no-restore
+dotnet release\OpenFeature.Benchmarks.dll
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,5 +129,5 @@ To run pipelines locally, you can follow these commands from a root project dire
 ```
 dotnet restore
 dotnet build --configuration Release --output "./release" --no-restore
-dotnet release\OpenFeature.Benchmarks.dll
+dotnet release/OpenFeature.Benchmarks.dll
 ```
--- a/test/OpenFeature.Benchmarks/OpenFeature.Benchmarks.csproj
+++ b/test/OpenFeature.Benchmarks/OpenFeature.Benchmarks.csproj
@@ -1,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <RootNamespace>OpenFeature.Benchmark</RootNamespace>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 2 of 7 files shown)


beeme1mr:
Hey @valentk777, thanks for the PR! Could you please adjust the prefix of the PR title? A `feat` would bump the SDK version by a minor version. The closest conventional commit prefix would either be `chore`, `perf` or `test`. I'll let you choose your favorite 😄.

valentk777:
Updated. @beeme1mr thanks for comment. Do not know that!

toddbaert:
I'll look closer at this later today, and if it looks good, either create a new issue or add to this one to run it in the CI, if reasonable.

test/OpenFeature.Benchmarks/OpenFeatureClientBenchmarks.cs
```diff
@@ -0,0 +1,175 @@
...
+            _defaultIntegerValue = fixture.Create<int>();
+            _defaultDoubleValue = fixture.Create<double>();
+            _defaultStructureValue = fixture.Create<Value>();
+            _emptyFlagOptions = new FlagEvaluationOptions(ImmutableList<Hook>.Empty, ImmutableDictionary<string, object>.Empty);
+        }
+
+        [Benchmark]
+        public async Task OpenFeatureClient_GetBooleanValue_WithoutEvaluationContext_WithoutFlagEvaluationOptions()
+        {
+            Api.Instance.SetProvider(new NoOpFeatureProvider());
```
kinyoklion, line 43: Should setting the provider and getting the client be setup outside the bench?

CONTRIBUTING.md
```diff
@@ -119,3 +119,15 @@ As with other OpenFeature SDKs, dotnet-sdk follows the
...
+## Benchmarking
+
+We use [BenchmarkDotNet](https://benchmarkdotnet.org/articles/overview.html) NuGet package to benchmark a code.
+
+To run pipelines locally, you can follow these commands from a root project directory.
+
+```
+dotnet restore
+dotnet build --configuration Release --output "./release" --no-restore
+dotnet release\OpenFeature.Benchmarks.dll
```
toddbaert, line 132:
```suggestion
dotnet release/OpenFeature.Benchmarks.dll
```

I think this should work on powershell, right?

toddbaert:
@valentk777 Thanks!

This is valuable as is, but is there a way we could automate it in the CI in a reliable way? The issue I always have with running something like this in github actions is I'm really not sure that github makes any performance guarantees, so I'm not sure that running this in the CI would produce consistent results.

Anyway, it's useful for testing different patches/branches on the same system!

valentk777:
> 

Hello @toddbaert.
Yes, it looks like it is possible. I never did that but with separate PR I could try. 
This is an example workflow. It will require storing results in order to compare them (not sure if storing part works here as expected), but at least it will run pipelines and produce results. that's one step further to automation. 
https://github.com/benchmark-action/github-action-benchmark/blob/master/.github/workflows/benchmarkdotnet.yml

If you create a new ticket, I could take that with another PR. Is this okay? :)

test/OpenFeature.Benchmarks/OpenFeatureClientBenchmarks.cs
valentk777: Updated. Thanks for comment

CONTRIBUTING.md
valentk777: thanks for the comment. updated

toddbaert:
> If you create a new ticket, I could take that with another PR. Is this okay? :)

Let's think about it... Like I said, even if we get a workflow working, it might not produce consistent enough results to be valuable as a CI job.

beeme1mr (approved):
Thanks for adding benchmarks! We can look into automatically running them in a follow up issue.

test/OpenFeature.Benchmarks/OpenFeature.Benchmarks.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
benjiro, line 4: Should we run it against net7 as well?

benjiro (approved):

test/OpenFeature.Benchmarks/OpenFeature.Benchmarks.csproj
toddbaert: I get `BenchmarkDotNet has failed to build the auto-generated boilerplate code` when I try net7.0. Either there's some config I'm missing or BenchmarkDotNet isn't compatible with 7.

I'll merge this and we can always come back to the 7 issue later, if we want.

Issue #136, opened 2023-07-12
System.Collections.Immutable needs to be updated to allow for version 7


danielp37:
Is there a reason why the PackageReference to System.Collection.Immutable doesn't allow version 7. Version 7 has been out for 8 months now and this restriction is preventing us from being able to use this library.

toddbaert:
Hey @danielp37 ! Thanks for opening this issue.

I've removed the upper bound on this dependency [here](https://github.com/open-feature/dotnet-sdk/pull/137).

cc @benjiro @kinyoklion do you think there's a reason to put one on this package?

toddbaert:
After some discussion we've decided just to increment this upper bound. A bit of a bummer, since we might be back here in a year unless we automate this somehow. Removing the entire upper bound seems risky as @kinyoklion notes in the PR.

PR #137, opened 2023-07-14
fix: max System.Collections.Immutable version ++

toddbaert:
remove max System.Collections.Immutable version

fixes: https://github.com/open-feature/dotnet-sdk/issues/136

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -23,6 +23,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Collections.Immutable" Version="[1.7.1, 7.0.0)" />
+    <PackageReference Include="System.Collections.Immutable" Version="[1.7.1, 8.0.0)" />
   </ItemGroup>
 </Project>


benjiro (approved):

kinyoklion:
Generally speaking this is a little concerning, because MS could change something and cause unpredictable behavior. On the other hand they have given up on the version having much meaning, so maybe this is the only alternative.

toddbaert:
> Generally speaking this is a little concerning, because MS could change something and cause unpredictable behavior. On the other hand they have given up on the version having much meaning, so maybe this is the only alternative.

I was concerned about this too... but I think if there was a breaking change the impacted us, we would see it relatively quickly in the CI, as long is it [was in .NET 6 or 7.](https://github.com/open-feature/dotnet-sdk/blob/main/.github/workflows/ci.yml#L24-L26). Maybe I'm wrong about that.

kinyoklion:
> > Generally speaking this is a little concerning, because MS could change something and cause unpredictable behavior. On the other hand they have given up on the version having much meaning, so maybe this is the only alternative.
> 
> I was concerned about this too... but I think if there was a breaking change the impacted us, we would see it relatively quickly in the CI, as long is it [was in .NET 6 or 7.](https://github.com/open-feature/dotnet-sdk/blob/main/.github/workflows/ci.yml?rgh-link-date=2023-07-14T18%3A16%3A28Z#L24-L26). Maybe I'm wrong about that.

I am sure we would detect it fast, but would consumer applications basically be breaking at the same time? Because they nuget install and different packages are fetched?

kinyoklion:
Generally speaking dotnet tries to install the lowest compatible version (backward from most package systems), at least it used to, so it may not be a problem.

toddbaert:
> Generally speaking dotnet tries to install the lowest compatible version (backward from most package systems), at least it used to, so it may not be a problem.

I just found this out, yes. Locally this is causing `1.7.1` to be resolved, which doesn't really help in testing.

I just manually forced version 7 and run the unit and e2e test suite and that worked... so I guess the more conservative option would be to allow up to v8.

Any thoughts?

toddbaert:
> Generally speaking a dotnet application will actually automatically generate binding redirects allowing you to use any version. ASP.Net applications do not, which is generally where this problem comes up.
> 
> For instance, for LD, this comes up with Microsoft.Logging.Extensions all the time:
> 
> ```
> <dependentAssembly>
>   <assemblyIdentity name="Microsoft.Extensions.Logging"
>     publicKeyToken="adb9793829ddae60"
>     culture="en-us" />
>   <bindingRedirect oldVersion="6.0.0" newVersion="7.0.0.0" />
> </dependentAssembly>
> ```
> 
> Updating it to 8.0.0 is safest. But it does mean it needs basically annually updated. Because MS will release the next one.

OK I'll do the conservative thing for now and then think on it more. Maybe we can automate it somehow.

kinyoklion (approved):

kinyoklion:
May want to update the PR title, for the release note.

PR #141, opened 2023-08-09
Remove moq library

### LINKED ISSUE #140: Replace `Moq` dependency with a different Mocking library
Due to the controversies around the `Moq` library (see https://github.com/moq/moq/issues/1372), we may want to consider switching to another mocking library, like e.g. [NSubstitute](https://github.com/nsubstitute/NSubstitute) to avoid any privacy issues. 
[1] Contains an example of how this was done by another previous `Moq` user.

Note: Currently we are using v4.18 of Moq, and the controversial change has been introduced with v4.20 and above, so currently this repo is not directly affected by the change. However, we should not merge any renovate PRs bumping the Moq dependency

**References**:
[1] https://github.com/Afterlife-Guide/AppSettings.Merge/pull/6

askpt:
## This PR

- adds this new feature

### Related Issues

Fixes #140


benjiro:
@askpt I'm fine with this going a head, thanks for doing it 👍🏻 

Are you able to fix the failing dotnet format error

askpt:
@benjiro The library is not fully removed in this PR. There are still a few tests missing that have more advanced scenarios. I don't mind fixing the build issues and creating a new issue for these tests.

benjiro:
If you rebase off `main` it will fix the warnings that are making E2E Test, and Test linux fail. 

` error EnableGenerateDocumentationFile: Set MSBuild property 'GenerateDocumentationFile' to 'true' in project file to enable IDE0005 (Remove unnecessary usings/imports) on build (https://github.com/dotnet/roslyn/issues/41640) [/home/runner/work/dotnet-sdk/dotnet-sdk/test/OpenFeature.E2ETests/OpenFeature.E2ETests.csproj::TargetFramework=net6.0]`

bacherfl (commented):
thanks for doing this and great work with the adapted tests so far - I've just had a look and all the assertions we previously had in place are still present.

bacherfl (approved):
This looks good to me! For me it would be ok to tackle the remaining test cases in a follow up - @benjiro what do you think?

askpt:
> This looks good to me! For me it would be ok to tackle the remaining test cases in a follow up - @benjiro what do you think?

Pushed more tests now. Only a handful missing. We can merge as it is.

bacherfl:
> > This looks good to me! For me it would be ok to tackle the remaining test cases in a follow up - @benjiro what do you think?
> 
> Pushed more tests now. Only a handful missing. We can merge as it is.

Thanks! Regarding the failing codecov check: This seems to be caused by a different behavior of NSubstitute vs Moq. The difference is in the `Register_Hooks_Should_Be_Available_At_All_Levels` in the `OpenFeatureHookTest` - This was previously covering the [exception handling](https://app.codecov.io/gh/open-feature/dotnet-sdk/pull/141/blob/src/OpenFeature/OpenFeatureClient.cs#L287) in the `Error` hook, as Moq mocks throw an exception for methods where no behavior has been defined, while NSubstitute does not seem to do that, therefore we do not reach this catch block in the aforementioned test.
Given that the purpose of this test is just to check if the `Register` method is available on all levels to register hooks, I think covering the exception handling of the Error hook can be part of a dedicated PR for that

askpt:
@bacherfl 
> > > This looks good to me! For me it would be ok to tackle the remaining test cases in a follow up - @benjiro what do you think?
> > 
> > 
> > Pushed more tests now. Only a handful missing. We can merge as it is.
> 
> Thanks! Regarding the failing codecov check: This seems to be caused by a different behavior of NSubstitute vs Moq. The difference is in the `Register_Hooks_Should_Be_Available_At_All_Levels` in the `OpenFeatureHookTest` - This was previously covering the [exception handling](https://app.codecov.io/gh/open-feature/dotnet-sdk/pull/141/blob/src/OpenFeature/OpenFeatureClient.cs#L287) in the `Error` hook, as Moq mocks throw an exception for methods where no behavior has been defined, while NSubstitute does not seem to do that, therefore we do not reach this catch block in the aforementioned test. Given that the purpose of this test is just to check if the `Register` method is available on all levels to register hooks, I think covering the exception handling of the Error hook can be part of a dedicated PR for that

These tests are the ones that miss the transition to NSubstitute:
- Hook_Hints_May_Be_Optional
- When_Error_Occurs_In_Before_Hook_Should_Return_Default_Value
- When_Error_Occurs_In_After_Hook_Should_Invoke_Error_Hook

I was trying to play around but they are failing because the sequence is not respected. 
I would suggest creating a new issue to migrate these three tests and I can try to fix it later.

See output:
`Expected to receive these calls in order:

    Hook.Before<Boolean>(any HookContext<Boolean>, Dictionary<String, Object>)
    FeatureProvider.ResolveBooleanValue(any String, any Boolean, any EvaluationContext)
    Hook.After<Boolean>(any HookContext<Boolean>, any FlagEvaluationDetails<Boolean>, Dictionary<String, Object>)
    Hook.Error<Boolean>(any HookContext<Boolean>, any Exception, Dictionary<String, Object>)
    Hook.Finally<Boolean>(any HookContext<Boolean>, Dictionary<String, Object>)

Actually received matching calls in this order:

    FeatureProvider.ResolveBooleanValue("test", True, EvaluationContext)`

Issue #126, opened 2023-05-16
Changes to support spec `v0.6.0`, `v0.7.0` compliance


toddbaert:
## Implement specification v0.7.0 changes

See [changes since last release](https://github.com/open-feature/spec/compare/v0.5.2...v0.7.0).

Please note there's no need to complete all these tasks in a single PR (in fact, that's not recommended).

### Functional requirements:

- [ ] implement flag metadata: https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a
- [x] open-feature/dotnet-sdk#131
  - [x] open-feature/spec#192 
- [x] implement init/shutdown (done with https://github.com/open-feature/dotnet-sdk/pull/158)
- [x] implement events: https://openfeature.dev/specification/sections/events
- [ ] #165

### Non-functional requirements:

- [ ] add relevant tests
- [ ] add examples and documentation, see template README
- [ ] update specification badge to `v0.7.0`: [![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.7.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.7.0)

### Keep in mind:

- backwards compatibility is _highly encouraged_
  - take every precaution to ensure existing code written by _application authors_ and existing providers implementations, do not break
- it's recommended these features are released together

askpt:
Closing this issue since it's a duplicate of https://github.com/open-feature/dotnet-sdk/milestone/1

PR #158, opened 2023-10-31
ffeat: Add support for provider shutdown and status.

kinyoklion:
## This PR

Adds support for provider initialization, shutdown, and status.

This PR does not add support for events, but does put several items in place to facilitate their implementation.

### Related Issues

Partially Addresses  #126


src/OpenFeature/Api.cs
```diff
@@ -36,31 +35,26 @@ public sealed class Api
...
 
         /// <summary>
-        /// Sets the feature provider
+        /// Sets the feature provider. In order to wait for the provider to be set, and initialization to complete,
+        /// await the returned task.
         /// </summary>
+        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public void SetProvider(FeatureProvider featureProvider)
+        public async Task SetProvider(FeatureProvider featureProvider)
```
kinyoklion, line 43: So, returning a task here is a breaking API change. This seems like the most appropriate action. Await the method if you care that it finishes initialization, don't await if you do not.

Other options would be adding another method, but this method would just be async without a way to know it finished, versus being sync.

This all applies to named providers as well.

src/OpenFeature/Api.cs
```diff
@@ -36,31 +35,26 @@ public sealed class Api
...
-            this._featureProviderLock.EnterWriteLock();
-            try
-            {
-                this._defaultProvider = featureProvider ?? this._defaultProvider;
-            }
-            finally
-            {
-                this._featureProviderLock.ExitWriteLock();
-            }
+            await this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
```
kinyoklion, line 45: I refactored the collection management out of the Api and moved it into the ProviderRepository. This is similar to the Java SDK.

src/OpenFeature/ProviderRepository.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// </param>
+        /// <param name="afterError">
+        /// called if an error happens during the initialization of the provider, only called if the provider needed
+        /// initialization
+        /// </param>
+        /// <param name="afterShutdown">called after a provider is shutdown, can be used to remove event handlers</param>
+        public async Task SetProvider(
+            FeatureProvider featureProvider,
+            EvaluationContext context,
+            Action<FeatureProvider> afterSet = null,
```
kinyoklion, line 60: All these "actions" can be used to facilitate implementing events support. I didn't want to put too much stuff together, so this should be a nice middle ground.

src/OpenFeature/Api.cs
benjiro: Yea I've been thinking about this myself its quite common to offer bother sync and async libraries especially when working on a old legacy project in full framework. Lets just leave it for now and see if anyone raise concern before going ahead an introduce a sync based methods.

toddbaert: > So, returning a task here is a breaking API change. This seems like the most appropriate action. Await the method if you care that it finishes initialization, don't await if you do not.
>
> Other options would be adding another method, but this method would just be async without a way to know it finished, versus being sync.

"events" are the mechanism by which you can call this synchronously and "know" the outcome (your READY/ERROR event handlers will run). I know they don't exist on this SDK yet, but they will eventually.

> Other options would be adding another method

This is what's been done in Java and JS, and what the spec recommends: https://openfeature.dev/specification/sections/flag-evaluation/#requirement-118

kinyoklion: (To elaborate `AndWait` or `Wait` in general have been used to provide sync APIs where the base API is async, so it would be a confusing name)

toddbaert: > The spec recommends setProviderAndWait

The spec intentionally doesn't prescribe a name for the function - everything outside of the block quotes is non-normative, and mostly there for example/clarity - we could certainly call them `setProvider` and `setProviderAsync`, which I agree would be optimal.

> (To elaborate AndWait or Wait in general have been used to provide sync APIs where the base API is async, so it would be a confusing name)

I agree, "wait" this implies the method will _actually_ block, in the Thread.Sleep sense, not await some callback/event.

Do you have objections to adding a `setProviderAsync` method instead? I agree that there's really no need for a synchronous `setProvider`, but given that there's equivalents in other SDKs and it would be a major version bump, I'm hesitant to changing it to async.

cc @benjiro

kinyoklion: Adding `setProviderAsync` is the lesser of two evils. The biggest problem with it is that it creates the implied meaning that `setProvider` is synchronous. We will document it, but unless people are thorough that will be the expectation.

To counter this we maybe should mark `setProvider` as obsolete, and then when there is a major version remove it. 

`[Obsolete("setProvider is deprecated, please use setProviderAsync instead. Note that setProvider executes asynchronously. To achieve the same functionality use setProviderAsync and do not wait for the result.")]`

toddbaert: I'd OK with either path, though my preference would be marking `setProvider` obsolete to avoid breaking changes.

Let's see what others think.

benjiro: Most dotnet devs are use to the Async suffix convention, so they will expect that the rest of the API follows which could lead to confusion. Usually when this approach is taken a library would provide both async, and sync methods. I think it would be unintuitive to mix the two (non async suffix/async suffix methods returning awaitable tasks). If we are going to go down the path of providing both sync/async methods i don't think there is any clean way without making a breaking change.

toddbaert: > Most dotnet devs are use to the Async suffix convention, so they will expect that the rest of the API follows which could lead to confusion. Usually when this approach is taken a library would provide both async, and sync methods. I think it would be unintuitive to mix the two (non async suffix/async suffix methods returning awaitable tasks). If we are going to go down the path of providing both sync/async methods i don't think there is any clean way without making a breaking change.

Well if that's the case, it's breaking either way.

So I guess the question is, is it better to release a 2.0 that supports both (with -Async suffixes) or should we just break this one method and make it async? Or is some 3rd way better?

kinyoklion: I think, if the API is fully async, then we don't need to worry about the Async suffix. It really is a case of libraries that implemented things synchronously, and then needed async methods as well. 

The interface for this API is almost all Async, and it happened that there was this sync method, so making it async would provide a consistent async interface. 

It may be worth checking the rest of the API to see if other items should move to async with this.

kinyoklion: (I guess the client is really where things are async, and the "Api" doesn't have that many methods that are async, or would conceivably be async.)

benjiro: I'm fine with keeping it all async and not providing a sync api. I think since we aren't 1.x.x yet its the best time to make a breaking change to SetProvider, i think we just need to call it out in the release notes.

benjiro (approved):

beeme1mr:
Sorry for the delayed response. Now that we have the ability to perform initialization from within the provider, it makes sense to make this an async call. Unfortunately, that means it will be a breaking change, but it should be easy for users to absorb this change. I did bring this up during the last community call, and no one had any concerns. I think I'll also do something similar in the JS SDK.

Before we release a 2.0 version, could the rest of the SDK be audited to make sure there are no other breaking changes we should consider?

FYI @toddbaert @kinyoklion @benjiro

src/OpenFeature/Api.cs
beeme1mr: This SDK is already version `1.3.1` but we can release a version `2.0` if it makes sense. I think it does make sense in this case since it simplifies waiting for a provider to initialize and the migration process should be straightforward.

toddbaert: I'm fine with the breaking change, especially if we agree we'll do the same breaking change in JS, which has the same async/await semantics.

test/OpenFeature.Tests/ProviderRepositoryTests.cs
```diff
@@ -0,0 +1,598 @@
...
+            provider1.GetStatus().Returns(ProviderStatus.NotReady);
+
+            var provider2 = Substitute.For<FeatureProvider>();
+            provider2.GetStatus().Returns(ProviderStatus.NotReady);
+
+            var context = new EvaluationContextBuilder().Build();
+
+            await repository.SetProvider("B", provider1, context);
+            await repository.SetProvider("A", provider1, context);
+            // Provider one is replaced for "A", but not "B.
```
toddbaert, line 385:
```suggestion
            // Provider one is replaced for "A", but not "B".
```

src/OpenFeature/ProviderRepository.cs
```diff
@@ -0,0 +1,291 @@
...
+#pragma warning disable CS4014
+                this.ShutdownIfUnused(oldProvider, afterShutdown, afterError);
+#pragma warning restore CS4014
+            }
+            finally
+            {
+                this._providersLock.ExitWriteLock();
+            }
+
+            await InitProvider(featureProvider, context, afterInitialization, afterError).ConfigureAwait(false);
```
toddbaert, line 190: At first I thought it was necessary to put this in the locked block, but I don't think it matters, since the important thing is to maintain thread safety of the provider dict, not make any guarantees about provider initialization order.

toddbaert (approved):
Testing looks good. I'm OK with the breaking change as discussed, but in that case I think we should probably wait for events and release both of these changes together.

kinyoklion:
> Testing looks good. I'm OK with the breaking change as discussed, but in that case I think we should probably wait for events and release both of these changes together.

Should I re-target this PR to a feature branch?

benjiro:
> > Testing looks good. I'm OK with the breaking change as discussed, but in that case I think we should probably wait for events and release both of these changes together.
> 
> Should I re-target this PR to a feature branch?

I would think it's fine to merge we just don't release/publish the next version until events are done.

toddbaert:
> > Testing looks good. I'm OK with the breaking change as discussed, but in that case I think we should probably wait for events and release both of these changes together.
> 
> Should I re-target this PR to a feature branch?

No no, we'll just refrain from releasing if possible until events. If we _need_ to release for some reason, it won't be a huge problem to release this on its own. The worse case would be we have another breaking change for events (not likely, it didn't happen in other SDKs) and we release a v3.

PR #166, opened 2024-01-02
feat: add support for eventing

bacherfl:
Work in Progress
This PR adds support for Eventing and thus solves a part of #126 

Note: I had to add some `Thread.Sleep`s in the unit tests to verify the event handlers, which are called by a separate thread, are called eventually. Coming from a Golang background, I would have liked to have something like [require.Eventually](https://pkg.go.dev/github.com/stretchr/testify/require#Eventually), but I did not find anything similar for dotnet. If anyone with more dotnet experience knows how to do this in a better way, I appreciate any suggestions :)


README.md
beeme1mr, line 1: Please update the features tables to reflect the updated status of "Eventing".

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using OpenFeature.Constant;
+using OpenFeature.Model;
+
+namespace OpenFeature
+{
+    internal class EventExecutor
+    {
+        private readonly Mutex _mutex = new Mutex();
```
kinyoklion, line 12: I think that a mutex is heavier than this needs, because it is a basic locking situation within 1 app domain. Could we use a `lock` statement with a lock object instead?

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,255 @@
...
+            var newEventPayload = new ProviderEventPayload { Type = ProviderEventTypes.ProviderStale };
+
+            eventExecutor.EventChannel.Writer.TryWrite(newEventPayload);
+
+            eventHandler.DidNotReceive().Invoke(newEventPayload);
+
+            eventHandler.DidNotReceive().Invoke(Arg.Is<ProviderEventPayload>(payload => payload.Type == ProviderEventTypes.ProviderStale));
+        }
+
+        [Fact]
```
kinyoklion, line 46: Are there any of these that can be tied back to their specification?

Example from other tests:
```
[Specification("1.1.2.2", "The provider mutator function MUST invoke the initialize function on the newly registered provider before using it to resolve flag values.")]
```

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+        public EventExecutor()
+        {
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+            this.ProcessEventAsync();
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+        }
+
+        internal void AddApiLevelHandler(ProviderEventTypes eventType, EventHandlerDelegate handler)
+        {
+            this._mutex.WaitOne();
```
Kavindu-Dodan, line 31: we should use the try-finally style for mutex usage so that we lock and release mutex regardless of any exception.

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+                eventHandlers = new List<EventHandlerDelegate>();
+                this._clientHandlers[client][eventType] = eventHandlers;
+            }
+
+            this._clientHandlers[client][eventType].Add(handler);
+
+            if (this._namedProviderReferences.TryGetValue(client, out var clientProviderReference))
+            {
+                this.EmitOnRegistration(clientProviderReference, eventType, handler);
+            }
```
Kavindu-Dodan, line 75: If the client name is not associated with a provider, then we have to associate the event with the default provider. This is due to how specifications define the behavior of named client [^1]

> If a client name is not bound to a particular provider, the client is associated with the default provider.

[^1]: https://openfeature.dev/specification/sections/flag-evaluation/#requirement-116

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+                    case ShutdownSignal _:
+                        this._shutdownSemaphore.Release();
+                        return;
+                }
+
+            }
+        }
+
+        // Method to signal shutdown
+        public async Task SignalShutdownAsync()
```
Kavindu-Dodan, line 264: I assume this needs to be wired to OpenFeature shutdown ? (refer - https://github.com/open-feature/dotnet-sdk/blob/2e1b27fdcd0b2dcc1c6e83a7f77eedbef911915a/src/OpenFeature/Api.cs#L206-L209) 🤔

Kavindu-Dodan (commented):
Great work 👏  I added few comments from spec perspective

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,294 @@
...
+                        break;
+                    case ShutdownSignal _:
+                        providerRef.ShutdownSemaphore.Release();
+                        return;
+                }
+            }
+        }
+
+        // Method to process events
+        private async Task ProcessEventAsync()
```
Kavindu-Dodan, line 218: personally I think this should be run in a dedicated thread instead of using async 🤔

src/OpenFeature/EventExecutor.cs
bacherfl: sure, will look into that

src/OpenFeature/EventExecutor.cs
bacherfl: yes, I will adapt the implementation to use a simple lock

src/OpenFeature/EventExecutor.cs
bacherfl: thx for the reference, seems like I missed that part - will adapt accordingly

test/OpenFeature.Tests/OpenFeatureEventTests.cs
bacherfl: Yes, after a quick look at the spec I found some potential mappings - I will go through the spec in detail and add them wherever it is applicable

bacherfl:
@Kavindu-Dodan @kinyoklion thank you for your feedback suggestions so far! I have adapted the PR accordingly. Looking at the spec I found some additional requirements that I would like to explicitly cover with some additional unit tests, so I will put this PR back into draft mode for now and will let you know when it's ready for another round of reviews

toddbaert:
I'm a bit late to this party (had some days off last week) but I will make this a priority tomorrow.

bacherfl:
This one should now be ready for review again

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,413 @@
...
+    [SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task")]
+    public class OpenFeatureEventTest : ClearOpenFeatureInstanceFixture
+    {
+        [Fact]
+        [Specification("5.1.2", "When a `provider` signals the occurrence of a particular `event`, the associated `client` and `API` event handlers MUST run.")]
+        public async Task Event_Executor_Should_Propagate_Events_ToGlobal_Handler()
+        {
+            var eventHandler = Substitute.For<EventHandlerDelegate>();
+
+            eventHandler.Invoke(Arg.Any<ProviderEventPayload>());
```
toddbaert, line 23: I'm not sure I understand why you invoke this here. Are you just testing that it can be?

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,368 @@
...
+                message = "Provider is in stale state";
+            }
+
+            if (message != "")
+            {
+                try
+                {
+                    handler.Invoke(new ProviderEventPayload
+                    {
+                        ProviderName = provider.Provider.GetMetadata().Name,
```
toddbaert, line 211:
```suggestion
                        ProviderName = provider.Provider?.GetMetadata()?.Name,
```

Technically this shouldn't be an issue, but I think we should be really defensive in the SDK.

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,368 @@
...
+                        Type = eventType,
+                        Message = message
+                    });
+                }
+                catch (Exception exc)
+                {
+                    Debug.WriteLine(exc);
```
toddbaert, line 218:
```suggestion
                    Debug.WriteLine("Error running handler: " + exc.toString());
```

I think this additional context is helpful.

src/OpenFeature/Model/ProviderEvents.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// The EventHandlerDelegate is an implementation of an Event Handler
+    /// </summary>
+    public delegate void EventHandlerDelegate(ProviderEventPayload eventDetails);
+
+    /// <summary>
+    /// Contains the payload of an OpenFeature Event.
+    /// </summary>
+    public class ProviderEventPayload
```
toddbaert, line 14: Can we add a test that populates and asserts the contents of some of the additional fields here including `Message` and `FlagChanges`?

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,413 @@
...
+        }
+
+        [Fact]
+        [Specification("5.1.2", "When a `provider` signals the occurrence of a particular `event`, the associated `client` and `API` event handlers MUST run.")]
+        [Specification("5.2.2", "The `API` MUST provide a function for associating `handler functions` with a particular `provider event type`.")]
+        [Specification("5.2.3", "The `event details` MUST contain the `provider name` associated with the event.")]
+        [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
+        [Specification("5.3.2", "If the provider's `initialize` function terminates abnormally, `PROVIDER_ERROR` handlers MUST run.")]
+        [Specification("5.3.3", "Handlers attached after the provider is already in the associated state, MUST run immediately.")]
+        public async Task API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Error()
```
toddbaert, line 134:
```suggestion
        public async Task API_Level_Event_Handlers_Should_Be_Informed_About_Error_State_After_Registering_Provider_Error()
```

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,413 @@
...
+
+            var testProvider = new TestProvider();
+            await Api.Instance.SetProvider(testProvider);
+
+            var newTestProvider = new TestProvider();
+            await Api.Instance.SetProvider(newTestProvider);
+
+            Thread.Sleep(1000);
+            eventHandler.Received(2).Invoke(Arg.Is<ProviderEventPayload>(payload => payload.ProviderName == testProvider.GetMetadata().Name));
+        }
```
toddbaert, line 200: Just because it's a different code-path, could you also assert this works as expected with `ProviderConfigurationChanged` events (manually emitted by the provider vs automatically by the SDK).

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,413 @@
...
+
+        [Fact]
+        [Specification("5.1.2", "When a `provider` signals the occurrence of a particular `event`, the associated `client` and `API` event handlers MUST run.")]
+        [Specification("5.2.1", "The `client` MUST provide a function for associating `handler functions` with a particular `provider event type`.")]
+        [Specification("5.2.3", "The `event details` MUST contain the `provider name` associated with the event.")]
+        [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
+        [Specification("5.2.5", "If a `handler function` terminates abnormally, other `handler` functions MUST run.")]
+        public async Task Client_Level_Event_Handlers_Should_Be_Executed_When_Other_Handler_Fails()
+        {
+            try
```
toddbaert, line 291: I'm confused by this try/catch. It seems unnecessary, and we wouldn't want handlers to throw (indeed it seems like you catch and swallow these errors as I'd expect)

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -0,0 +1,413 @@
...
+        [Fact]
+        [Specification("5.1.2", "When a `provider` signals the occurrence of a particular `event`, the associated `client` and `API` event handlers MUST run.")]
+        [Specification("5.2.1", "The `client` MUST provide a function for associating `handler functions` with a particular `provider event type`.")]
+        [Specification("5.2.3", "The `event details` MUST contain the `provider name` associated with the event.")]
+        [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
+        [Specification("5.2.5", "If a `handler function` terminates abnormally, other `handler` functions MUST run.")]
+        public async Task API_Level_Event_Handlers_Should_Be_Executed_When_Other_Handler_Fails()
+        {
+            try
+            {
```
toddbaert, line 234: Same as https://github.com/open-feature/dotnet-sdk/pull/166/files#r1446447754

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,368 @@
...
+                        break;
+                    case ShutdownSignal _:
+                        typedProviderRef.ShutdownSemaphore.Release();
+                        return;
+                }
+            }
+        }
+
+        // Method to process events
+        private async void ProcessEventAsync()
```
toddbaert, line 247: This tastes like Go (not that I have a problem with it, it's clear to me).

src/OpenFeature/Model/ProviderEvents.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public string Message { get; set; }
+        /// <summary>
+        /// A List of flags that have been changed.
+        /// </summary>
+        public List<string> FlagChanges { get; set; }
+        /// <summary>
+        /// Metadata information for the event.
+        /// </summary>
+        public Dictionary<string, object> EventMetadata { get; set; }
```
toddbaert, line 35: assert this in a test too, please

src/OpenFeature/EventExecutor.cs
toddbaert: I think also we should make this a `error` level log. Otherwise it will be tricky to trace errors in event handlers. We do this in [JS](https://github.com/open-feature/js-sdk/blob/072d2c7570274e003e32cd50807ed8a0839e7b4c/packages/shared/src/open-feature.ts#L94) and [Java](https://github.com/open-feature/java-sdk/blob/9851d4ba1c238b72673932badce04a4d1bc882fb/src/main/java/dev/openfeature/sdk/EventSupport.java#L142C36-L143C22)

But interested in what @kinyoklion @beeme1mr @Kavindu-Dodan think.

toddbaert (commented):
Really great job! Excellent testing and very readable overall. Left a few comments. I think my only blocker is the test I described [here](https://github.com/open-feature/dotnet-sdk/pull/166/files#r1446464228).

src/OpenFeature/EventExecutor.cs
kinyoklion: I actually think there is a bit more of a problem than that. We are theoretically using `Microsoft.Extensions.Logging` and allowing customization of that. But it looks like that is only implemented at the client level, and API level doesn't have any logger instance.

So, it should be using a logger instance that can be set by a developer. Then once it is doing that, then likely it should be error level.

(`Debug.WriteLine` is a system diagnostics feature)

toddbaert: Oh yes, good catch. I missed that entirely.

test/OpenFeature.Tests/OpenFeatureEventTests.cs
bacherfl: Seems like that was a copy-paste leftover. I will remove this again

test/OpenFeature.Tests/OpenFeatureEventTests.cs
bacherfl: good catch (... ;-) ) That was a left over from trying out different approaches on how to best handle exceptions thrown by the event handlers, but this can be removed now

src/OpenFeature/EventExecutor.cs
bacherfl: alright, I will adapt it to use a `Microsoft.Extensions.Logging` logger

bacherfl: I added a `SetLogger()` to ehe `Api` method that allows to set a `Microsoft.Extensions.Logging` implementation, which will propagated through to the `EventExecutor` (which is the only component of the `Api` that potentially logs something right now, but we may use this for other parts of the `Api` as well in the future).

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,371 @@
...
```
Kavindu-Dodan, line 250: Given we execute this in a thread, do we need this method to be async anymore?

src/OpenFeature/EventExecutor.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            while (true)
+            {
+                var item = await this.EventChannel.Reader.ReadAsync().ConfigureAwait(false);
+
+                switch (item)
+                {
+                    case Event e:
+                        lock (this._lockObj)
```
Kavindu-Dodan, line 259: Minor - I think we should use a different lock object as this method is executed concurrentrly

src/OpenFeature/EventExecutor.cs
Kavindu-Dodan: also check asyn requirement on `private async void ProcessFeatureProviderEventsAsync(object providerRef)` method

src/OpenFeature/EventExecutor.cs
Kavindu-Dodan: Yes, I think most of the implementation logic here is quite similar to Go-sdk. And I think that's great as we can correlate implementation and compare them :)

src/OpenFeature/EventExecutor.cs
bacherfl: I think this still needs to be `async` since there's an `await` being used within the method - removing the `async` will cause the compiler to complain

src/OpenFeature/EventExecutor.cs
bacherfl: I would rather keep using the same lock as this is the one protecting the `apiHandlers` / `clientHandlers`, `defaultProvider` etc. so using a different lock object might lead to unpredictable behavior. If performance becomes an issue we can look into more fine granular locking, but to avoid premature optimisation I opted to keep this as simple as possible for now

toddbaert: I don't think any of these are in hot paths (adding handlers or handling events). I wouldn't expect many handlers to be added or many events to be fire in a small or sustained time-frame.

toddbaert (approved):
Thanks @bacherfl ! Recent changes have me satisfied.

kinyoklion (approved):

src/OpenFeature/EventExecutor.cs
Kavindu-Dodan: Thanks , will resolve this conversation

src/OpenFeature/EventExecutor.cs
Kavindu-Dodan: Thanks for the explanation, I am satisfied with the explanation :)

Kavindu-Dodan (approved):

Kavindu-Dodan:
Approved the PR. Feel free to resolve my comments and merge. Nice work @bacherfl 🎉

src/OpenFeature/OpenFeature.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggerVer)" />
+    <PackageReference Include="System.Threading.Channels" Version="8.0.0" />
```
benjiro, line 10: Probably worth moving this to [Common.props](https://github.com/open-feature/dotnet-sdk/blob/main/build/Common.props#L27) and adding a range from [6.0.0,)

bacherfl: Thank you for the suggestion, I have updated the PR accordingly :)

toddbaert:
I'll merge this EOD Tuesday next week unless I hear objections!

benjiro (approved):

Issue #170, opened 2024-01-11
[BUG] System.Collections.Immutable version constraint causing grief


austindrenski:
### Observed behavior

Still trying to figure out all the triggering bits, but tldr; in a mixed-target solution (e.g. `net48`, `net8.0`, and `netstandard2.0`), `OpenFeature` is breaking source generators when loading the solution in IDEs (e.g. both Rider and Visual Studio), as well as when building directly from `MSBuild.exe` when certain MSBuild targets are used (seems to relate to the VS install and which task DLLs are loaded).

```
CSC : error CS8784:
  Generator 'SomeGenerator' failed to initialize. It will not contribute to the output and compilation errors may occur
  as a result. Exception was of type 'FileNotFoundException' with message 'Could not load file or assembly 'System.
  Collections.Immutable, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
  The system cannot find the file specified.'.

[C:\a\dotnet\SomeSolution\src\SomeProject\SomeProject.csproj::TargetFramework=netstandard2.0]
```

where `SomeProject.csproj`:

```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="../SomeProject.Generators/SomeProject.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
  </ItemGroup>

</Project>
```

and `SomeProject.Generators.csproj`:

```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.7.0" PrivateAssets="all" />
  </ItemGroup>

</Project>
```

and then `SomeProject.csproj` is eventually being used by `SomeLegacyWebProject.csproj`:

```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>library</OutputType>
    <TargetFramework>net48</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <!-- N.B. in reality this is a transitive reference coming from deep within the project graph -->
[...]

PR #171, opened 2024-01-11
fix: Remove upper-bound version constraint from SCI

austindrenski:
Fixes: #170

See: #136, #137, #170 

___

_Anticipate similar arguments/discussion as in #137, but opening as a companion to #170 because this is causing me real-world grief today 🙃_

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -24,7 +24,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Collections.Immutable" Version="[1.7.1, 8.0.0)" />
+    <PackageReference Include="System.Collections.Immutable" Version="[1.7.1,)" />
     <PackageReference Include="System.Threading.Channels" Version="[6.0.0, 8.0.0)" />
   </ItemGroup>
 </Project>
--- a/build/Common.props
+++ b/build/Common.props
@@ -25,6 +25,6 @@
 
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="[1.7.1,)" />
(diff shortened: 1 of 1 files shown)


beeme1mr:
No idea why the code coverage report is failing...

At any rate, I'll add the people from the referenced thread.

benjiro (approved):

toddbaert:
@kinyoklion Has you opinion on this changed at all since #170 ?

austindrenski:
Bumping because this is unfortunately causing an increasing amount of developer friction inside my org.

Including this package breaks our in-house source generators when invoked by IDE-triggered design-time builds in such a way that my normally strong msbuild-foo has thus far been unable to combat.

I would like to get some movement on this today, otherwise I'm looking at needing to publish a forked package to appease a growing horde of angered devs. _(This is, of course, a me problem, so I don't mention it to force a decision one way or the other, just to say that it would be nice to hear from folks on whether this is a quick action or something I need to self-serve for the foreseeable future.)_

/cc @toddbaert @kinyoklion

toddbaert:
> Can someone catch me up on why we're going out of our way to set upper-bound constraints in the first place?
> 
> Originally, I assumed it was something specific about `System.Collections.Immutable` that we were depending on (e.g. internal impl details, sketchy reflection, etc.), but now noticing that we do this in a bunch of places, so wondering if this is just an artifact of someone's repo template, or if there's some bigger concern that I'm missing 🤔

I think it was just out of fear of breaking changes in further releases of these dependencies, but I think I'm on board with just removing the upper bounds.

toddbaert (approved):

toddbaert:
@austindrenski I've removed the upper bound on `System.Threading.Channels` as well.

PR #172, opened 2024-01-11
refactor: Add TFMs for net{6,7,8}.0

austindrenski:
As a consumer of the OpenFeature SDK, I want `OpenFeature.dll` to be built and tested against all of the [supported TFMs](https://learn.microsoft.com/lifecycle/products/microsoft-net-and-net-core) that my projects might encounter in the wild.

This PR:
1. Updates the `global.json` to `8.0.100`
2. Updates workflows using _only_ `7.0.x` to instead resolve based on `global.json`
3. Updates the project(s) TFMs to include `net6.0`, `net7.0`, and `net8.0`
   - `net6.0` is EOL in `2024-11-12`
   - `net7.0` is EOL in `2024-05-14`

Fixes: #160, #169

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Setup .NET Core 6.0.x, 7.0.x
+    - name: Setup .NET SDK
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: |
@@ -27,7 +27,7 @@ jobs:
 
     - name: Run Tests
       run: dotnet test test/OpenFeature.Tests/ --configuration Release --logger GitHubActions
-  
+
   unit-tests-windows:
     runs-on: windows-latest
     steps:
@@ -35,7 +35,7 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Setup .NET Core 6.0.x, 7.0.x
+    - name: Setup .NET SDK
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: |
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -21,10 +21,8 @@ jobs:
       with:
         fetch-depth: 0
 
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Setup .NET Core 6.0.x, 7.0.x
+    - name: Setup .NET SDK
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: |
```
benjiro, line 24: Probably better to explicitly install 8.0.x because sometimes the actions base container isn't up to date. Thoughts?

benjiro (commented):
I'm fine with this probably a bigger question currently we aren't really using any features above .netstandard2.0 so targeting net6,7,8 is really pointless because there is no runtime benefit. I guess in the future we can look at optimisations and utilize some of the new runtime feature were available

.github/workflows/ci.yml
austindrenski: I usually don't specify the current version that will be resolved by `global.json` from a DRY perspective, but there's nothing _stopping_ us from doing that if you have a strong preference for explicitness.

The only caveat I would provide to _trying_ to be explicit by listing the version that'll be resolved by `global.json` is that `actions/setup-dotnet` will resolve the version from `global.json` (i.e. no-op'ing if everything is lined up correctly).

Here's a screencap from what `actions/setup-dotnet` did on a previous run for this PR:

[image](https://github.com/open-feature/dotnet-sdk/actions/runs/7494052087/job/20401116920?pr=172#step:3:1)

> [...] because sometimes the actions base container isn't up to date. Thoughts?

Want to respond to this directly because it's a question that I've had to answer before (so I know the sentiment!), but in my experience, there really isn't a case where you end up with the wrong version because the newest version hasn't yet made it into base container ___provided you're using a `global.json` with one of the roll-forward options enabled___ .

benjiro (approved):

toddbaert (approved):

Issue #54, opened 2022-09-01
[FEATURE] Publish dev builds to ghcr nuget


benjiro:
### Requirements

We want to be able to publish development builds to nuget so they are able to be distributed and tested by provider implementers before the major release.

PR #173, opened 2024-01-11
chore: Add support for GitHub Packages

austindrenski:
| `GITHUB_REF`     | version format                                     |
|------------------|----------------------------------------------------|
| `refs/heads/*`   | `#.#.#-ci.{%Y%m%d}T{%H%M%S}+sha.${GITHUB_SHA:0:9}` |
| `refs/pull/*`    | `#.#.#-pr.{%Y%m%d}T{%H%M%S}+sha.${GITHUB_SHA:0:9}` |

See: #54, open-feature/dotnet-sdk-contrib#134

Closes: #54

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,3 +44,46 @@ jobs:
 
     - name: Run Tests
       run: dotnet test test\OpenFeature.Tests\ --configuration Release --logger GitHubActions
+
+  packaging:
+    needs:
+    - unit-tests-linux
+    - unit-tests-windows
+
+    permissions:
+      contents: read
+      packages: write
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup .NET SDK
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: |
+          6.0.x
+          7.0.x
+
+    - name: Restore
+      run: dotnet restore
+
+    - name: Pack NuGet packages (CI versions)
+      if: startsWith(github.ref, 'refs/heads/')
+      run: dotnet pack --no-restore --version-suffix "ci.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}"
+
+    - name: Pack NuGet packages (PR versions)
+      if: startsWith(github.ref, 'refs/pull/')
+      run: dotnet pack --no-restore --version-suffix "pr.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}"
+
+    - name: Publish NuGet packages (base)
+      if: github.event.pull_request.head.repo.fork == false
+      run: dotnet nuget push "src/**/*.nupkg" --api-key "${{ secrets.GITHUB_TOKEN }}" --source https://nuget.pkg.github.com/open-feature/index.json
+
+    - name: Publish NuGet packages (fork)
+      if: github.event.pull_request.head.repo.fork == true
+      uses: actions/upload-artifact@v4.2.0
+      with:
+        name: nupkgs
+        path: src/**/*.nupkg
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -2,7 +2,10 @@
   <Import Project=".\Common.props" />
 
   <PropertyGroup>
+    <ContinuousIntegrationBuild Condition="'$(CI)' == 'true'">true</ContinuousIntegrationBuild>
+    <Deterministic Condition="'$(CI)' == 'true'">true</Deterministic>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackRelease>true</PackRelease>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -16,7 +19,7 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <Authors>OpenFeature Authors</Authors>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <Version>$(VersionNumber)</Version>
+    <VersionPrefix>$(VersionNumber)</VersionPrefix>
     <AssemblyVersion>$(VersionNumber)</AssemblyVersion>
     <FileVersion>$(VersionNumber)</FileVersion>
   </PropertyGroup>
@@ -32,11 +35,4 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(Deterministic)'=='true'">
-    <SourceRoot Include="$(MSBuildThisFileDirectory)/" />
-  </ItemGroup>
-
-  <PropertyGroup Condition="'$(Deterministic)'=='true'">
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
-  </PropertyGroup>
 </Project>
--- a/build/Common.props
+++ b/build/Common.props
@@ -28,4 +28,8 @@
     <PackageReference Include="System.Collections.Immutable" Version="[1.7.1,)" />
     <PackageReference Include="System.Threading.Channels" Version="[6.0.0,)" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(OS)' == 'Unix'">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="all" />
+  </ItemGroup>
 </Project>


.github/workflows/ci.yml
```diff
@@ -44,3 +44,39 @@ jobs:
...
+      if: startsWith(github.ref, 'refs/pull/')
+      run: dotnet pack --no-restore --version-suffix "pr.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}"
+
+    - name: Pack NuGet packages (Release versions)
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: dotnet pack --no-restore
+
+    - name: Publish NuGet packages
+      if: startsWith(github.ref, 'refs/heads/') || startsWith(github.ref, 'refs/pull/') || startsWith(github.ref, 'refs/tags/v')
+      run: dotnet nuget push "src/**/*.nupkg" --api-key "${GITHUB_TOKEN}" --source https://nuget.pkg.github.com/open-feature/index.json
```
beeme1mr, line 82: You'll need to use `secrets.NUGET_TOKEN`.

austindrenski: Hmmm, I use `GITHUB_TOKEN` directly in my private repos, is the custom token related to public repo/forks stuff?

austindrenski: In any event, updated and it seems happier so that's that!

austindrenski:
Well, still failed, but now I'm pretty sure it's just the fork issue. Assuming `${{ secrets.NUGET_TOKEN }}` is available to others, then this will work as is, otherwise, I can update the conditions to skip publishing from forks.

austindrenski:
> FYI the secrets.NUGET_TOKEN is for nuget.org registry but looking at this change its pushing to github registry. Also i believe forks won't have access to any secrets from memory. You might need to update the permissions on the workflow to get write access to github registeries on the secrets.GITHUB_TOKEN that the action runner generates https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes

Bah, I knew there was something I was forgetting. Thanks, will update.

> One problem with using githubs nuget registry is you need a github account setup to pull down the packages (this could of change since last time i checked). This is why i originally suggested using something like MyGet which offer free accounts for OpenSource projects.
> 
> I think gating the prerelease packages behind auth defeats the purpose, and makes for a cumbersome developer experience. Happy to go a head with it if people are fine with that.

Full agree that GitHub should just give us the option to allow un-auth'd package retrieval for OSS projects, but I also don't think the auth requirement is all that big of a blocker.

Now this could be some corporate dev bias, but of all the authenticated feeds I have to deal with at my day job, GitHub Packages is the least annoying of them lol.

But bigger picture, I think providing prerelease packages directly from GitHub is a really useful solution for discoverability, plus if you're already setup to consume one GitHub package feed, then that auth carries over to the N+1 feed too.

So wouldn't ever recommend we _only_ publish to GitHub Packages (e.g. as opposed to, say, nuget.org), but I think it's a really nice built-in option that comes pretty much for free (as in beer/time) and offers a nice UX in terms of code+packages+releases all in one place.

_(also, none of what I'm saying here precludes also publishing to myget in addition to github, I'm just saying there's no reason to __not__ offer github while we're at it.)_

benjiro (approved):

beeme1mr (approved):

toddbaert (changes requested):
I think this is wrong. I could be missing something, but pull_request_target runs in the context of the base branch - so this will make all our CI tests run the code in main, not the code in whatever PR.

austindrenski:
> I think this is wrong. I could be missing something, but pull_request_target runs in the context of the base branch - so this will make all our CI tests run the code in main, not the code in whatever PR.

Geeze, I think you're right. I'm not sure what I was doing when I changed that (testing token perms, maybe?).

Reverting now.

.github/workflows/ci.yml
```diff
@@ -44,3 +44,43 @@ jobs:
...
+      with:
+        dotnet-version: |
+          6.0.x
+          7.0.x
+
+    - name: Restore
+      run: dotnet restore
+
+    - name: Pack NuGet packages (CI versions)
+      if: startsWith(github.ref, 'refs/heads/')
```
toddbaert, line 73: I'm not sure how this will work, because we seem to have this job setup to run [only on `main`](https://github.com/open-feature/dotnet-sdk/pull/173/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR4-R11).

I wonder if a neater solution would be to do something like we do for Java, and just publish every merge to main to Github.

That wouldn't get us testing for branches, but I think it might be good enough?

austindrenski: I always have to refer to the docs when messing with the `on` syntax _(hence that oops you caught lol)_, but `on.pull_request: [ main ]` results in this workflow running for PRs _targeting_ `main` ___(emphasis added)___:

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow ___runs when___ a pull request is opened or reopened or ___when the head branch of the pull request is updated___.
>
> [...]
> 
> For example, this workflow will run when someone opens a pull request that targets a branch whose name starts with `releases/`:
> 
> ```yaml
> on:
>   pull_request:
>     types:
>       - opened
>     branches:
>       - 'releases/**'
> ```
>
> _via https://docs.github.com/actions/using-workflows/events-that-trigger-workflows#pull_request_

which seems to be true based on what I'm seeing on this PR (after fixing the `pull_request{_target => }` issue):

image

austindrenski: > [...]
>
> I wonder if a neater solution would be to do something like we do for Java, and just publish every merge to main to Github.
>
> That wouldn't get us testing for branches, but I think it might be good enough?

The workflow, as-is, _will_ publish a `x.y.z-ci.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}` package for every commit on `main`, so we're already good on that.

But the sticky part is seeing if we can get _fork_ PRs to publish the `x.y.z-pr.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}` like I'd hoped we'd be able to.

I use a workflow like this for internal packages all the time, so pretty sure its the `pr from fork` part that's causing issues.

Let me try one more tweak to try getting it to publish to the _fork's package feed_, and if that doesn't work, I'll set this step to skip for PRs until we can figure out another solution (e.g. uploading the nuget package as an artifact, but its always a PITA to use local packages, so was hoping to find a nicer solution).

toddbaert: Right OK, I confused myself on that one :sweat_smile: 

So, if I'm understanding correctly, your intent is to trigger this for any branch or tag, or any PR to main? I wonder if that's a bit excessive?

toddbaert: As you can see in this case, it will run (and fail) with every PR from a fork due to missing secrets.

toddbaert: I wonder if it's not better to instead only run this for a specific tag pattern or branch, or maybe on every push to main?

austindrenski: Okay last tweak to try and get this working:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index f9aa524..5649f08 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,13 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/v')
       run: dotnet pack --no-restore

-    - name: Publish NuGet packages
-      if: startsWith(github.ref, 'refs/heads/') || startsWith(github.ref, 'refs/pull/') || startsWith(github.ref, 'refs/tags/v')
+    - name: Publish NuGet packages (base)
+      if: github.event.pull_request.head.repo.fork == false
       run: dotnet nuget push "src/**/*.nupkg" --api-key "${{ secrets.GITHUB_TOKEN }}" --source https://nuget.pkg.github.com/open-feature/index.json
+
+    - name: Publish NuGet packages (fork)
+      if: github.event.pull_request.head.repo.fork == true
+      uses: actions/upload-artifact@v4.2.0
+      with:
+        name: nupkgs
+        path: src/**/*.nupkg
```

> So, if I'm understanding correctly, your intent is to trigger this for any branch or tag, or any PR to main?

Almost, my intent was to have this run for every push to `main` (excluding `**.md`-only changes), which is what the `ci.yml` workflow was already setup to do.

We could do it for tags too (PR includes support for that), but we would have to add `on.push.tags: [ v* ]` which I have not included here (yet?).


> I wonder if that's a bit excessive?

Personally, I don't think its excessive, but I could imagine it someday _maybe_ becoming excessive, but there are ways we could mitigate if/when that occurs (e.g. changing the `main` packages to run on a cron schedule conditional on changes having landed since the last run).

As for the _volume_ of packages this _could_ add to our package feed, we can setup a workflow to prune the CI packages, keeping only the most recent N versions, if it becomes an issue (or proactively, no reason we'd have to wait).

> I wonder if it's not better to instead only run this for a specific tag pattern or branch, or maybe on every push to main?

See above, but can add the tag pattern if you agree (e.g. `v*`), but I think the rest of the triggers make sense as is (i.e. we probably don't want this workflow triggering on _any_ branch push, since pushes to a branch that isn't opened as a PR yet could be a little much).

.github/workflows/ci.yml
```diff
@@ -44,3 +44,50 @@ jobs:
...
+      run: dotnet pack --no-restore --version-suffix "ci.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}"
+
+    - name: Pack NuGet packages (PR versions)
+      if: startsWith(github.ref, 'refs/pull/')
+      run: dotnet pack --no-restore --version-suffix "pr.$(date -u +%Y%m%dT%H%M%S)+sha.${GITHUB_SHA:0:9}"
+
+    - name: Pack NuGet packages (Release versions)
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: dotnet pack --no-restore
+
```
toddbaert, line 83: I guess there's no harm in publishing proper release versions to GH as well, but we could remove this since we'll have one of these in nuget.

austindrenski: Definitely don't want anyone mistaking this PR as proposing GH as an alternative to NuGet for official packages used in production code.

But I think the _packing_ step here isn't necessarily the problem (i.e. prefer as much consistency between release and non-release workflow runs as possible), but rather that the following step then _pushes_ the stable versions into the GH feed.

I'll update the conditions below to skip pushing in that case (and in the future we could have this push to NuGet, but not trying to shoehorn that in right now).

.github/workflows/ci.yml
toddbaert: > We could do it for tags too (PR includes support for that), but we would have to add on.push.tags: [ v* ] which I have not included here (yet?).

Well, that was one of the things I was confused by, and why I was trying to understand your goal. I saw `if: startsWith(github.ref, 'refs/tags/v')` but wasn't sure how that would ever apply.

.github/workflows/ci.yml
toddbaert: > I'll update the conditions below to skip pushing in that case (and in the future we could have this push to NuGet, but not trying to shoehorn that in right now).

We push official releases to nuget with our release please flow when [PRs like this](https://github.com/open-feature/dotnet-sdk/pull/154) are merged. I'm not sure we'd want to publish anything to nuget other than that.

austindrenski: Okay, yeah, seeing that now.

I think you're right that I'm creating more confusion by implying this _might_ do something for tag builds, so I'm just going to drop the notion of tag builds altogether in this PR.

.github/workflows/ci.yml
austindrenski: Addressed via https://github.com/open-feature/dotnet-sdk/pull/173#discussion_r1459859465

tldr; removing references to release versions and tags from this PR to avoid any confusion

toddbaert (approved):

austindrenski:
Screencap for future readers showing where to find the uploaded NuGet packages for PR builds from forks:

image

Clicking on the `nupkgs` artifact will trigger a download of `nupkgs.zip` which should contain any packages resulting from the workflow:

image

image

austindrenski:
Actually, hold on, don't merge this just yet @toddbaert. The uploaded package doesn't look like it included the version suffix. Reviewing now.

austindrenski:
> Actually, hold on, don't merge this just yet @toddbaert. The uploaded package doesn't look like it included the version suffix. Reviewing now.

Okay, I'd overlooked that `build/Common.prod.props` was using `<Version>` instead of `<VersionPrefix>`.

tldr; you can use either, but if you use `<VersionPrefix>` then you can supply `<VersionSuffix>` from the CLI via `--version-suffix` and MSBuild will do something along the lines of:

```csproj
<Version Condition="'$(Version)' == '' and '$(VersionSuffix)' == ''">$(VersionPrefix)</Version>
<Version Condition="'$(Version)' == '' and '$(VersionSuffix)' != ''">$(VersionPrefix)-$(VersionSuffix)</Version>
```

```diff
diff --git a/build/Common.prod.props b/build/Common.prod.props
index c9e5855..383da6f 100644
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -19,7 +19,7 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <Authors>OpenFeature Authors</Authors>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <Version>$(VersionNumber)</Version>
+    <VersionPrefix>$(VersionNumber)</VersionPrefix>
     <AssemblyVersion>$(VersionNumber)</AssemblyVersion>
     <FileVersion>$(VersionNumber)</FileVersion>
   </PropertyGroup>
```

austindrenski:
Resultant version of that last CI run looks better, but still concerned that the Deterministic + SourceLink checks are failing:

image

austindrenski:
Okay, not entirely sure why it caused issues, but removing these since they're now set at the top of the file:

```diff
diff --git a/build/Common.prod.props b/build/Common.prod.props
index 383da6f..4d073ec 100644
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -35,11 +35,4 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>

-  <ItemGroup Condition="'$(Deterministic)'=='true'">
-    <SourceRoot Include="$(MSBuildThisFileDirectory)/" />
-  </ItemGroup>
-
-  <PropertyGroup Condition="'$(Deterministic)'=='true'">
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
-  </PropertyGroup>
 </Project>
```


Local packing still results in the expected source root/paths:

image

build/Common.prod.props
```diff
@@ -16,7 +19,7 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <Authors>OpenFeature Authors</Authors>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <Version>$(VersionNumber)</Version>
```
toddbaert, line 19: Are you sure about this change? I'm worried about the impact that might have on the prod release. I'm not sure how this, `VersionNumber`, and `VersionPrefix` work in conjuction.

build/Common.prod.props
```diff
@@ -32,11 +35,4 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(Deterministic)'=='true'">
-    <SourceRoot Include="$(MSBuildThisFileDirectory)/" />
-  </ItemGroup>
-
-  <PropertyGroup Condition="'$(Deterministic)'=='true'">
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
-  </PropertyGroup>
```
toddbaert, line 41: Can you explain why you removed this?

build/Common.prod.props
austindrenski: Very fair question! I use this pattern all the time, albeit for internal packages, so just double-checked:

- [`src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultAssemblyInfo.targets`](https://github.com/dotnet/sdk/blob/47245cb134e77f41b5d664413e1fa98b781c8ec1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultAssemblyInfo.targets)
  ```csproj
  <!--
  ***********************************************************************************************
  Microsoft.NET.DefaultAssemblyInfo.targets
  
  WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
            created a backup copy.  Incorrect changes to this file will make it
            impossible to load or build your projects from the command-line or the IDE.
  
  Copyright (c) .NET Foundation. All rights reserved.
  ***********************************************************************************************
  -->
  <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  
    <PropertyGroup Condition=" '$(Version)' == '' ">
      <VersionPrefix Condition=" '$(VersionPrefix)' == '' ">1.0.0</VersionPrefix>
      <Version Condition=" '$(VersionSuffix)' != '' ">$(VersionPrefix)-$(VersionSuffix)</Version>
      <Version Condition=" '$(Version)' == '' ">$(VersionPrefix)</Version>
    </PropertyGroup>
  
    <PropertyGroup>
      <Authors Condition=" '$(Authors)'=='' ">$(AssemblyName)</Authors>
      <Company Condition=" '$(Company)'=='' ">$(Authors)</Company>
      <AssemblyTitle Condition=" '$(AssemblyTitle)' == '' ">$(AssemblyName)</AssemblyTitle>
      <Product Condition=" '$(Product)' == ''">$(AssemblyName)</Product>
    </PropertyGroup>
  
  </Project>
  ```
- https://learn.microsoft.com/dotnet/core/tools/dotnet-pack
  > - **`--version-suffix <VERSION_SUFFIX>`**
  > 
  >   Defines the value for the `VersionSuffix` MSBuild property. The effect of this property on the package version depends on the values of the `Version` and `VersionPrefix` properties, as shown in the following table:
  > 
  >   | Properties with values              | Package version                     |
  >   |-------------------------------------|-------------------------------------|
  >   | None                                | `1.0.0`                             |
  >   | `Version`                           | `$(Version)`                        |
  >   | `VersionPrefix` only                | `$(VersionPrefix)`                  |
  >   | `VersionSuffix` only                | `1.0.0-$(VersionSuffix)`            |
  >   | `VersionPrefix` and `VersionSuffix` | `$(VersionPrefix)-$(VersionSuffix)` |
  > 
  >   If you want to use `--version-suffix`, specify `VersionPrefix` and not `Version` in the project file. For example, if `VersionPrefix` is `0.1.2` and you pass `--version-suffix rc.1` to `dotnet pack`, the package version will be `0.1.2-rc.1`.
  > 
  >   If `Version` has a value and you pass `--version-suffix` to `dotnet pack`, the value specified for `--version-suffix` is ignored.
  >

toddbaert: Sorry, I didn't see your previous comment [here](https://github.com/open-feature/dotnet-sdk/pull/173#issuecomment-1901231142).

build/Common.prod.props
toddbaert: lol sorry, I see https://github.com/open-feature/dotnet-sdk/pull/173#issuecomment-1901244232 now

toddbaert (approved):
Approving again, and will merge in a moment, but consider [this](https://github.com/open-feature/dotnet-sdk/pull/173#discussion_r1459941779).

build/Common.prod.props
austindrenski: I removed `L39-L41` because I realized that my changes at the top of this file were already setting it. Those changes condition both `<ContinuousIntegrationBuild>` and `<Deterministic>` on the [`CI` env var which GitHub Actions always sets to `true`](https://docs.github.com/actions/learn-github-actions/variables#default-environment-variables) ensuring these properties are set by default for _all_ `dotnet` commands we run anywhere in our CI workflows.

> [!Note]
>
> _A happy side-effect of conditioning these on the `CI` env var is that it makes testing this stuff locally much easier, because you can just prefix your commands, e.g. `CI=true dotnet pack`, and everything triggers accordingly._
>
> _This is admittedly not dramatically different from running `dotnet pack -p:Deterministic=true`, but IMO its easier to grok these things as "when in CI, do ..." vs. "when deterministic do ..."._
>
> _This^ might not be a terribly compelling reason on its own, but in conjunction with knowing that every `dotnet` command we run in any of our workflows will always have these flags set correctly, without having to remember to pass `-p:Deterministic=true`, is, I think, nice to have._

```diff
diff --git a/build/Common.prod.props b/build/Common.prod.props
index 63a52d4..4d073ec 100644
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -2,7 +2,10 @@
   <Import Project=".\Common.props" />

   <PropertyGroup>
+    <ContinuousIntegrationBuild Condition="'$(CI)' == 'true'">true</ContinuousIntegrationBuild>
+    <Deterministic Condition="'$(CI)' == 'true'">true</Deterministic>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackRelease>true</PackRelease>
   </PropertyGroup>

[...]
```

___

As for why I had to remove `<SourceRoot>`, that I'm less sure about. When testing the packing step locally (i.e. `CI=true dotnet pack ...`), the resultant package was being reported as _not_ having Source Link enabled correctly, according to [NuGetPackageExplorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer).

Removing that block and then re-running `CI=true dotnet pack ...` produced a package that NuGetPackageExplorer reported as having Source Link enabled.

_(I wish I had a more rigorous justification for this one, but that's all I've got for now...)_

austindrenski: > lol sorry, I see [#173 (comment)](https://github.com/open-feature/dotnet-sdk/pull/173#issuecomment-1901244232) now

Ope, silly GitHub didn't show me this comment until after I replied to your question.

All the same, let me know if you want me to get you a better answer on the `SourceLink` one.

build/Common.props
```diff
@@ -28,4 +28,8 @@
     <PackageReference Include="System.Collections.Immutable" Version="[1.7.1,)" />
     <PackageReference Include="System.Threading.Channels" Version="[6.0.0,)" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(OS)' == 'Unix'">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="all" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
austindrenski: > Is this because we're not building on Windows in the job?

Yep! It gives `dotnet build` enough context to build the `net462` TFM on Linux, so we can run package builds on `ubuntu-latest` which, among other reasons, runs circles around `windows-latest` for `dotnet restore` and anything else that uses compression:
- `windows-latest`:
  ```
  Run dotnet test test\OpenFeature.Tests\ --configuration Release --logger GitHubActions
    Determining projects to restore...
    Restored D:\a\dotnet-sdk\dotnet-sdk\src\OpenFeature\OpenFeature.csproj (in 21.2 sec).
    Restored D:\a\dotnet-sdk\dotnet-sdk\test\OpenFeature.Tests\OpenFeature.Tests.csproj (in 1.69 min).
  ```
- `ubuntu-latest`:
  ```
  Run dotnet test test/OpenFeature.Tests/ --configuration Release --logger GitHubActions
    Determining projects to restore...
    Restored /home/runner/work/dotnet-sdk/dotnet-sdk/src/OpenFeature/OpenFeature.csproj (in 5.33 sec).
    Restored /home/runner/work/dotnet-sdk/dotnet-sdk/test/OpenFeature.Tests/OpenFeature.Tests.csproj (in 5.5 sec).
  ```


> Will this have any impact on the release build ([it does run on windows](https://github.com/open-feature/dotnet-sdk/blob/main/.github/workflows/release.yml))?

It should have no impact since it's conditioned on the `OS` var and the reference assemblies won't be included in the compilation, leaving `dotnet` to find the actual assemblies in the GAC (or on disk, can't remember which it actually checks, so don't quote me on this part).

But it does mean, that once everyone is comfortable with it (i.e. this workflow runs without issues a few times), we could then update `release.yml` to run on `ubuntu-latest` too.

That workflow obviously runs less frequently than `ci.yml`, so the `dotnet restore` perf won't be as big of a consideration in that case, but I personally feel better building my internal stable releases on the `ubuntu-latest` runners. Things just always seem to work better there lol (e.g. cli perf, line endings, etc).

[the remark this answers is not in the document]
toddbaert: Makes sense. I'm in favor of that change in the future. Thanks for the explanation 🙏

austindrenski:
Woohoo, it lives! 🥳🥳🥳 

- https://github.com/open-feature/dotnet-sdk/pkgs/nuget/OpenFeature

[image](https://github.com/open-feature/dotnet-sdk/pkgs/nuget/OpenFeature)

PR #178, opened 2024-01-16
chore: Enable Central Package Management (CPM)

austindrenski:
See: open-feature/dotnet-sdk-contrib#136

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,32 @@
+<Project>
+
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup Label="src">
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" />
+    <PackageVersion Include="System.Collections.Immutable" Version="1.7.1" />
+    <PackageVersion Include="System.Threading.Channels" Version="6.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Label="test">
+    <PackageVersion Include="AutoFixture" Version="4.17.0" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.13.1" />
+    <PackageVersion Include="coverlet.collector" Version="3.1.2" />
+    <PackageVersion Include="coverlet.msbuild" Version="3.1.2" />
+    <PackageVersion Include="FluentAssertions" Version="6.7.0" />
+    <PackageVersion Include="GitHubActionsTestLogger" Version="2.3.3" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageVersion Include="NSubstitute" Version="5.0.0" />
+    <PackageVersion Include="OpenFeature.Contrib.Providers.Flagd" Version="0.1.5" />
+    <PackageVersion Include="SpecFlow" Version="3.9.74" />
+    <PackageVersion Include="SpecFlow.Tools.MsBuild.Generation" Version="3.9.74" />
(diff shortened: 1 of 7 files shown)


Directory.Packages.props
```diff
@@ -0,0 +1,32 @@
...
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="[1.0.0, 2.0)" />
+    <PackageVersion Include="System.Collections.Immutable" Version="1.7.1" />
+    <PackageVersion Include="System.Threading.Channels" Version="6.0.0" />
+  </ItemGroup>
```
benjiro, line 12: Might be worth adding a common here to explain that these are the runtime depencies where the list below is the testing deps

Directory.Packages.props
```diff
@@ -0,0 +1,32 @@
...
+  <ItemGroup>
+    <PackageVersion Include="AutoFixture" Version="4.17.0" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.13.1" />
+    <PackageVersion Include="coverlet.collector" Version="3.1.2" />
+    <PackageVersion Include="coverlet.msbuild" Version="3.1.2" />
+    <PackageVersion Include="FluentAssertions" Version="6.7.0" />
+    <PackageVersion Include="GitHubActionsTestLogger" Version="2.3.3" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageVersion Include="NSubstitute" Version="5.0.0" />
+    <PackageVersion Include="NUnit" Version="3.13.2" />
```
benjiro, line 23: How did this sneak in i thought we were using xunit

Directory.Packages.props
```diff
@@ -0,0 +1,32 @@
...
+    <PackageVersion Include="OpenFeature.Contrib.Providers.Flagd" Version="0.1.5" />
+    <PackageVersion Include="SpecFlow" Version="3.9.74" />
+    <PackageVersion Include="SpecFlow.Tools.MsBuild.Generation" Version="3.9.74" />
+    <PackageVersion Include="SpecFlow.xUnit" Version="3.9.74" />
+    <PackageVersion Include="xunit" Version="[2.4.1, 3.0)" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="[2.4.3, 3.0)" />
```
benjiro, line 29: Is it even worth while have a range on xunit, we should just target the latest version as its a testing dependencies

benjiro (commented):
All for enabling CPM it makes it much easier to manage deps, nice work. Left a couple of comments about cleaning up some test deps that probably shouldn't be there i know you didn't add them but while your changing it may as well clean them up

Directory.Packages.props
austindrenski: How do you feel about just `Label="src"` and `Label="test"` to match the directories? Or were you thinking something more elaborate?

Directory.Packages.props
austindrenski: Actually, this looks like its referenced from `test/OpenFeatures.E2ETests/OpenFeatures.E2ETests.csproj`, but not seeing any actual code references to it, so going to try 🪓'ing it

Directory.Packages.props
austindrenski: IMO absolutely not, would love to just set these like normal. I was actually going to sneak that in here, but the latest version of xUnit has some new analyzers that lit up when I did that, so pulled back to minimize the excitement in this PR.

If you're onboard with it, though, I'll take this^ on as a follow up task once this lands.

Directory.Packages.props
benjiro: yea thats fine i was just thinking using xml comments  or something along those lines a label is fine as well

Directory.Packages.props
benjiro: Yea we can resolve those changes under another PR

benjiro (approved):

askpt (approved):
LGTM

kinyoklion (approved):

PR #184, opened 2024-01-17
feat!: add cancellation tokens

austindrenski:
See: https://github.com/open-feature/dotnet-sdk/pull/154#issuecomment-1894703244

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,13 +72,13 @@ dotnet add package OpenFeature
 public async Task Example()
 {
     // Register your feature flag provider
-    await Api.Instance.SetProvider(new InMemoryProvider());
+    await Api.Instance.SetProviderAsync(new InMemoryProvider());
 
     // Create a new client
     FeatureClient client = Api.Instance.GetClient();
 
     // Evaluate your feature flag
-    bool v2Enabled = await client.GetBooleanValue("v2_enabled", false);
+    bool v2Enabled = await client.GetBooleanValueAsync("v2_enabled", false);
 
     if ( v2Enabled )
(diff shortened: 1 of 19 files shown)


benjiro (approved):
I'm fine with this, probably my bad for missing putting the CTs in the first place. Thanks for fixing it up, makes sense to support this specially if the library is being consumed in a web app.

Async suffix is a well understood convention and mentioned many times in the originally docs. I'm good with adding it, breaking change is minimal.

Nice work

toddbaert (changes requested):
I definitely want to do this, but I want to collect a few breaking changes and create an issue for them all... then we can perhaps maintain a v2 branch for a while, and merge it all at once when we're ready to release.

I'll create that issue to discuss, but blocking for now. Let me know if you disagree.

kinyoklion:
We should maybe update the conventional commit tag to have an exclamation point. Makes it easy to see that it needs to be held off.

toddbaert:
I'm assigning this to myself. Next week I'll work on cleaning it up, which will hopefully also help me understand how we can best use these tokens in terms of how they integrate with the rest of the SDK's semantics.

> We should maybe update the conventional commit tag to have an exclamation point. Makes it easy to see that it needs to be held off.

Did this, thanks @kinyoklion

toddbaert:
I've started on this and https://github.com/open-feature/dotnet-sdk/pull/185

toddbaert:
Closing and continuing with https://github.com/open-feature/dotnet-sdk/pull/268

PR #185, opened 2024-01-17
refactor: Use ValueTask on plumbing methods 

austindrenski:
Switch to `ValueTask` on methods that are likely to complete synchronously and/or be populated with contrib-/user-provided delegates.

See: #184

___

This follows #184 with an idea (probably best punted for post-`2.0.0`) to switch plumbing-type methods from `Task` to `ValueTask` to take advantage of the zero-cost-sync-completion sugar that the `dotnet/runtime` folks have generously given us to play with.

I've got a bigger PR planned where I'll propose returning `ValueTask<T>` on the `IFeatureClient` methods in order to hide a TTL caching layer that users can use to amortize provider requests (e.g. users could configure `IFeatureClient` to cache a value for flag `K` for `N` seconds, and for those `N` seconds any subsequent requests to _any_ `IFeatureClient` for flag `K` would result in a sync read from a memory cache).

But that's for later, and for now this PR just looks to make those user-/contrib-provided plumbing methods super lightweight on the scalding hotpath.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,25 +291,25 @@ To satisfy the interface, all methods (`BeforeAsync`/`AfterAsync`/`FinallyAsync`
 ```csharp
 public class MyHook : Hook
 {
-  public Task<EvaluationContext> BeforeAsync<T>(HookContext<T> context,
+  public ValueTask<EvaluationContext> BeforeAsync<T>(HookContext<T> context,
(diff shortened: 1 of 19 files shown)


toddbaert:
I've started updating this and https://github.com/open-feature/dotnet-sdk/pull/184

toddbaert:
Closing and continuing with https://github.com/open-feature/dotnet-sdk/pull/268

Issue #186, opened 2024-01-17
[BUG] `Api.Shutdown()` hangs on reuse


austindrenski:
### Observed behavior

Stumbled onto this while investigating failing unit tests add as part of #181 that fail when run in concert, but pass when run independently.

Working theory is that we're not restarting the consumer thread. Patch forthcoming.

See: #181

austindrenski:
Ah, okay, its blocking on the bounded channel inside of `EventExecuter`. Looking at a couple of options.

PR #187, opened 2024-01-17
fix: Fix ArgumentOutOfRangeException for empty hooks

austindrenski:
Stumbled into this one while working on #186 (while working on #181).

See: dotnet/runtime#62121, dotnet/runtime#62126

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -127,7 +127,26 @@ namespace OpenFeature
         /// </para>
         /// </summary>
         /// <param name="hooks">A list of <see cref="Hook"/></param>
-        public void AddHooks(IEnumerable<Hook> hooks) => this._hooks.PushRange(hooks.ToArray());
+        public void AddHooks(IEnumerable<Hook> hooks)
+#if NET7_0_OR_GREATER
+            => this._hooks.PushRange(hooks as Hook[] ?? hooks.ToArray());
+#else
+        {
+            // See: https://github.com/dotnet/runtime/issues/62121
+            if (hooks is Hook[] array)
+            {
+                if (array.Length > 0)
+                    this._hooks.PushRange(array);
+
+                return;
+            }
+
+            array = hooks.ToArray();
+
+            if (array.Length > 0)
+                this._hooks.PushRange(array);
+        }
+#endif
 
         /// <summary>
         /// Adds a hook to global hooks list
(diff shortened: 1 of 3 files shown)


austindrenski:
Confirmed patch success in CI (as expected):

[image](https://github.com/open-feature/dotnet-sdk/actions/runs/7557211781?pr=187)

src/OpenFeature/Api.cs
```diff
@@ -127,7 +127,26 @@ public FeatureProvider GetProvider(string clientName)
         /// </para>
         /// </summary>
         /// <param name="hooks">A list of <see cref="Hook"/></param>
-        public void AddHooks(IEnumerable<Hook> hooks) => this._hooks.PushRange(hooks.ToArray());
+        public void AddHooks(IEnumerable<Hook> hooks)
+#if NET7_0_OR_GREATER
+            => this._hooks.PushRange(hooks as Hook[] ?? hooks.ToArray());
+#else
```
toddbaert, line 133: I guess the only reason for the use of this directive is to save use the additional checks? I suppose we could always just run the `#else` block?

I wouldn't say this is a hot code path, so I'm not sure if it's worth it, but I don't feel strongly about it.

src/OpenFeature/Api.cs
```diff
@@ -127,7 +127,26 @@ public FeatureProvider GetProvider(string clientName)
         /// </para>
         /// </summary>
         /// <param name="hooks">A list of <see cref="Hook"/></param>
-        public void AddHooks(IEnumerable<Hook> hooks) => this._hooks.PushRange(hooks.ToArray());
+        public void AddHooks(IEnumerable<Hook> hooks)
+#if NET7_0_OR_GREATER
+            => this._hooks.PushRange(hooks as Hook[] ?? hooks.ToArray());
```
toddbaert, line 132: I think we may have the same issue with the `client`: https://github.com/open-feature/dotnet-sdk/blob/a6062fe2b9f0d83490c7ce900e837863521f5f55/src/OpenFeature/OpenFeatureClient.cs#L109

src/OpenFeature/Api.cs
austindrenski: I agree, but it's not so much the hot path I was thinking about here, but rather flagging that this behavior can get 🪓'd in the future when our TFMs are all `NET7_0_OR_GREATER` (which theoretically will happen...someday).

Without the `#if`/`#else` I'd worry that some future reader might misunderstand what we're doing here (especially since this is super easy to miss if you're only running the default `net8.0` tests locally).

toddbaert: Good enough for me.

askpt (approved):
LGTM!

PR #188, opened 2024-01-17
fix: More robust shutdown/cleanup/reset

austindrenski:
Previously, we shutdown the consumer thread causing any reuse of the Api to block on the second event emitted to the event executor.

Fixes: #186

___

Stumbled into this one while working on #181 in which I've added unit tests that each setup their own independent `IHost` and then dispose of it at the conclusion of each test case.

But since the `Api.Instance` is inherently static/shared, subsequent test cases were observed blocking until CI timeout occurred.

This PR could reasonably be criticized for going a step too far by introducing a new dependency on `Microsoft.Bcl.AsyncInterfaces`, but I decided to propose it anyways on the basis that  `Microsoft.Bcl.AsyncInterfaces` is a transitive dependency of `Microsoft.Extensions.DependencyInjection.Abstractions` which itself is a transitive dependency of `Microsoft.Extensions.Logging.Abstractions >= 8.0.0`, and only applies to TFMs that do not ship `IAsyncDisposable` and friends in-the-box, i.e., `net462` and `netstandard2.0`.

An even cleaner solution would be to just bump our minimum required version of `Microsoft.Extensions.Logging.Abstractions` to `8.0.0` at which point `Microsoft.Bcl.AsyncInterfaces` comes transitively for free. I would prefer to do this, but figured that might be too controversial right before a release. _(That said, if you like this idea too, please chime in because I have a strong belief that exceedingly few consumers in the wild __actually__ need us to keep our MELA dependency pinned back in 2017. The newer versions of MELA are available for all of the relevant TFMs, which ought to mean there's no dependency hell argument at play on this one.)_

### DIFF
--- a/src/OpenFeature/OpenFeature.csproj
+++ b/src/OpenFeature/OpenFeature.csproj
@@ -7,6 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesVer)" Condition="'$(TargetFramework)' == 'net462'" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesVer)" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggerVer)" />
   </ItemGroup>
 
--- a/build/Common.props
+++ b/build/Common.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>latest</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
   </PropertyGroup>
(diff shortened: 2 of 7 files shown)


src/OpenFeature/OpenFeature.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesVer)" />
```
askpt, line 10: Is it possible to limit this package to the frameworks that need it as you mention in the PR description?
```suggestion
    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesVer) Condition="'$(TargetFramework)' == 'netstandard2.0'"" />
```

askpt (commented):
Adding a small suggestion concerning the package installation.

src/OpenFeature/OpenFeature.csproj
austindrenski: It should be, let me try that locally!

austindrenski: Yep, works as expected, thanks for the suggestion!

askpt (approved):
LGTM!

toddbaert (approved):
I tried to find something to nitpick, but I couldn't! :sweat_smile:

src/OpenFeature/OpenFeature.csproj
toddbaert: Nice call @askpt

toddbaert:
> An even cleaner solution would be to just bump our minimum required version of Microsoft.Extensions.Logging.Abstractions to 8.0.0 at which point Microsoft.Bcl.AsyncInterfaces comes transitively for free. I would prefer to do this, but figured that might be too controversial right before a release.

Let's do this after this upcoming release, maybe?

austindrenski:
> > An even cleaner solution would be to just bump our minimum required version of Microsoft.Extensions.Logging.Abstractions to 8.0.0 at which point Microsoft.Bcl.AsyncInterfaces comes transitively for free. I would prefer to do this, but figured that might be too controversial right before a release.
> 
> Let's do this after this upcoming release, maybe?

Yeah, that's sounds fair. Plus, it'll dovetail nicely with #189 which also shouldn't land until after the upcoming release is finalized.

PR #190, opened 2024-01-17
chore: revert breaking setProvider

toddbaert:
This PR makes the breaking change [here](https://github.com/open-feature/dotnet-sdk/pull/158#discussion_r1378120046) non breaking, by adding a new method instead. This means our pending release will no longer be breaking.

This will allow us to release all the pending .NET work, without merging https://github.com/open-feature/dotnet-sdk/pull/184. I like the idea of https://github.com/open-feature/dotnet-sdk/pull/184, but I think I'd like us all to take some time to discuss exactly what we expect the `cancellationToken` to do in all circumstances (what should the SDK do on SDK methods like `init()/shutdown()`?, and what should providers do?).

I think buys us some without sacrificing too much, and we can release a 2.0.0 in the coming weeks.

Interested to hear the opinion of others.

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -44,7 +44,7 @@ namespace OpenFeature
         [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
         public void SetProvider(FeatureProvider featureProvider)
         {
-            this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
+            this._eventExecutor.RegisterDefaultFeatureProvider(featureProvider);
             _ = this._repository.SetProvider(featureProvider, this.GetContext());
         }
 
@@ -68,7 +68,7 @@ namespace OpenFeature
         [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
         public void SetProvider(string clientName, FeatureProvider featureProvider)
         {
-            this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
+            this._eventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
             _ = this._repository.SetProvider(clientName, featureProvider, this.GetContext());
         }
 
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -47,7 +47,7 @@ namespace OpenFeature
         public void SetProvider(FeatureProvider featureProvider)
         {
             this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
-            this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
+            _ = this._repository.SetProvider(featureProvider, this.GetContext());
         }
 
         /// <summary>
@@ -71,7 +71,7 @@ namespace OpenFeature
         public void SetProvider(string clientName, FeatureProvider featureProvider)
         {
             this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
-            this._repository.SetProvider(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
+            _ = this._repository.SetProvider(clientName, featureProvider, this.GetContext());
         }
 
         /// <summary>
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
@@ -38,17 +39,40 @@ namespace OpenFeature
         private Api() { }
 
         /// <summary>
-        /// Sets the feature provider. In order to wait for the provider to be set, and initialization to complete,
+        /// Sets the default feature provider to given clientName without awaiting its initialization.
+        /// </summary>
+        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
+        /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
+        [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
+        public void SetProvider(FeatureProvider featureProvider)
+        {
+            this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
+            this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
         /// await the returned task.
         /// </summary>
         /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProvider(FeatureProvider featureProvider)
+        public async Task SetProviderAsync(FeatureProvider featureProvider)
         {
             this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
             await this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Sets the feature provider to given clientName without awaiting its initialization.
+        /// </summary>
+        /// <param name="clientName">Name of client</param>
+        /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
+        [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
+        public void SetProvider(string clientName, FeatureProvider featureProvider)
+        {
+            this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
+            this._repository.SetProvider(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
+        }
 
         /// <summary>
         /// Sets the feature provider to given clientName. In order to wait for the provider to be set, and
@@ -56,7 +80,7 @@ namespace OpenFeature
         /// </summary>
         /// <param name="clientName">Name of client</param>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProvider(string clientName, FeatureProvider featureProvider)
+        public async Task SetProviderAsync(string clientName, FeatureProvider featureProvider)
         {
             this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
             await this._repository.SetProvider(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
--- a/test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
+++ b/test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
@@ -42,7 +42,7 @@ namespace OpenFeature.E2ETests
         {
             _scenarioContext = scenarioContext;
             var flagdProvider = new FlagdProvider();
-            Api.Instance.SetProvider(flagdProvider).Wait();
+            Api.Instance.SetProviderAsync(flagdProvider).Wait();
             client = Api.Instance.GetClient();
         }
 
--- a/test/OpenFeature.Tests/ClearOpenFeatureInstanceFixture.cs
+++ b/test/OpenFeature.Tests/ClearOpenFeatureInstanceFixture.cs
@@ -7,7 +7,7 @@ namespace OpenFeature.Tests
         {
             Api.Instance.SetContext(null);
             Api.Instance.ClearHooks();
-            Api.Instance.SetProvider(new NoOpFeatureProvider()).Wait();
+            Api.Instance.SetProviderAsync(new NoOpFeatureProvider()).Wait();
         }
     }
 }
--- a/test/OpenFeature.Tests/OpenFeatureEventTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureEventTests.cs
@@ -143,7 +143,9 @@ namespace OpenFeature.Tests
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
             var testProvider = new TestProvider();
+            #pragma warning disable CS0618 // Type or member is obsolete
             Api.Instance.SetProvider(testProvider);
+            #pragma warning restore CS0618 // Type or member is obsolete
 
             Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
--- a/test/OpenFeature.Tests/OpenFeatureEventTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureEventTests.cs
@@ -136,7 +136,7 @@ namespace OpenFeature.Tests
         [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
         [Specification("5.3.1", "If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers MUST run.")]
         [Specification("5.3.3", "Handlers attached after the provider is already in the associated state, MUST run immediately.")]
-        public async void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
+        public async Task API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
         {
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
--- a/test/OpenFeature.Tests/OpenFeatureEventTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureEventTests.cs
@@ -137,7 +137,7 @@ namespace OpenFeature.Tests
         [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
         [Specification("5.3.1", "If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers MUST run.")]
         [Specification("5.3.3", "Handlers attached after the provider is already in the associated state, MUST run immediately.")]
-        public void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
+        public async void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
         {
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
--- a/test/OpenFeature.Tests/OpenFeatureEventTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureEventTests.cs
@@ -121,13 +121,12 @@ namespace OpenFeature.Tests
 
             Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
-            Thread.Sleep(1000);
-            eventHandler
+            await Utils.AssertUntilAsync(_ => eventHandler
                 .Received()
                 .Invoke(
                     Arg.Is<ProviderEventPayload>(
                         payload => payload.ProviderName == testProvider.GetMetadata().Name && payload.Type == ProviderEventTypes.ProviderReady
-                    ));
+                    )));
         }
 
         [Fact]
--- a/test/OpenFeature.Tests/OpenFeatureEventTests.cs
+++ b/test/OpenFeature.Tests/OpenFeatureEventTests.cs
@@ -70,7 +70,7 @@ namespace OpenFeature.Tests
             Api.Instance.AddHandler(ProviderEventTypes.ProviderStale, eventHandler);
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             testProvider.SendEvent(ProviderEventTypes.ProviderConfigurationChanged);
             testProvider.SendEvent(ProviderEventTypes.ProviderError);
@@ -118,7 +118,32 @@ namespace OpenFeature.Tests
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
+
+            Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
+
+            Thread.Sleep(1000);
+            eventHandler
+                .Received()
+                .Invoke(
+                    Arg.Is<ProviderEventPayload>(
+                        payload => payload.ProviderName == testProvider.GetMetadata().Name && payload.Type == ProviderEventTypes.ProviderReady
+                    ));
+        }
+
+        [Fact]
+        [Specification("5.1.2", "When a `provider` signals the occurrence of a particular `event`, the associated `client` and `API` event handlers MUST run.")]
+        [Specification("5.2.2", "The `API` MUST provide a function for associating `handler functions` with a particular `provider event type`.")]
+        [Specification("5.2.3", "The `event details` MUST contain the `provider name` associated with the event.")]
+        [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
+        [Specification("5.3.1", "If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers MUST run.")]
+        [Specification("5.3.3", "Handlers attached after the provider is already in the associated state, MUST run immediately.")]
+        public void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
+        {
+            var eventHandler = Substitute.For<EventHandlerDelegate>();
+
+            var testProvider = new TestProvider();
+            Api.Instance.SetProvider(testProvider);
 
             Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
@@ -143,7 +168,7 @@ namespace OpenFeature.Tests
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             testProvider.SetStatus(ProviderStatus.Error);
 
@@ -169,7 +194,7 @@ namespace OpenFeature.Tests
             var eventHandler = Substitute.For<EventHandlerDelegate>();
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             testProvider.SetStatus(ProviderStatus.Stale);
 
@@ -198,12 +223,12 @@ namespace OpenFeature.Tests
             Api.Instance.AddHandler(ProviderEventTypes.ProviderConfigurationChanged, eventHandler);
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             testProvider.SendEvent(ProviderEventTypes.ProviderConfigurationChanged);
 
             var newTestProvider = new TestProvider();
-            await Api.Instance.SetProvider(newTestProvider);
+            await Api.Instance.SetProviderAsync(newTestProvider);
 
             newTestProvider.SendEvent(ProviderEventTypes.ProviderConfigurationChanged);
 
@@ -224,13 +249,13 @@ namespace OpenFeature.Tests
             Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             Thread.Sleep(1000);
             Api.Instance.RemoveHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
             var newTestProvider = new TestProvider();
-            await Api.Instance.SetProvider(newTestProvider);
+            await Api.Instance.SetProviderAsync(newTestProvider);
 
             eventHandler.Received(1).Invoke(Arg.Is<ProviderEventPayload>(payload => payload.ProviderName == testProvider.GetMetadata().Name));
         }
@@ -255,7 +280,7 @@ namespace OpenFeature.Tests
             Api.Instance.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
             var testProvider = new TestProvider(fixture.Create<string>());
-            await Api.Instance.SetProvider(testProvider);
+            await Api.Instance.SetProviderAsync(testProvider);
 
             failingEventHandler.Received().Invoke(Arg.Is<ProviderEventPayload>(payload => payload.ProviderName == testProvider.GetMetadata().Name));
             eventHandler.Received().Invoke(Arg.Is<ProviderEventPayload>(payload => payload.ProviderName == testProvider.GetMetadata().Name));
@@ -274,7 +299,7 @@ namespace OpenFeature.Tests
             var myClient = Api.Instance.GetClient(fixture.Create<string>());
 
             var testProvider = new TestProvider();
-            await Api.Instance.SetProvider(myClient.GetMetadata().Name, testProvider);
+            await Api.Instance.SetProviderAsync(myClient.GetMetadata().Name, testProvider);
 
             myClient.AddHandler(ProviderEventTypes.ProviderReady, eventHandler);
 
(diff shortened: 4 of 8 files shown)


src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
+        /// Sets the feature provider to given clientName without awaiting its initialization.
+        /// </summary>
+        /// <param name="clientName">Name of client</param>
+        /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
+        [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
+        public void SetProvider(string clientName, FeatureProvider featureProvider)
+        {
+            this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
+            this._repository.SetProvider(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
+        }
```
austindrenski, line 75:
```suggestion
        public void SetProvider(string clientName, FeatureProvider featureProvider) => this.SetProviderAsync(clientName, featureProvider).AsTask().GetAwaiter().GetResult();
```

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
+
+        /// <summary>
+        /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
         /// await the returned task.
         /// </summary>
         /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProvider(FeatureProvider featureProvider)
+        public async Task SetProviderAsync(FeatureProvider featureProvider)
```
austindrenski, line 59: While we're revisiting this...

I have some work planned/started (see: #185) to go through and `ValueTask`-ize some (all?) of our public async methods that have a ~~reasonable~~ vague likelihood of either completing synchronously, and/or being likely to be implemented/plugged into from contrib/user code in ways that could benefit from some non-allocating completions.

```suggestion
        public ValueTask SetProviderAsync(FeatureProvider featureProvider)
```

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
         {
             this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
             await this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
```
austindrenski, line 62: and then until #185 updates `ProviderRepository.SetProvider(...)` to return `ValueTask`, we can just wrap its result ourselves:

```suggestion
            return new ValueTask(this._repository.SetProvider(featureProvider, this.GetContext()));
```

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
+        }
 
         /// <summary>
         /// Sets the feature provider to given clientName. In order to wait for the provider to be set, and
         /// initialization to complete, await the returned task.
         /// </summary>
         /// <param name="clientName">Name of client</param>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProvider(string clientName, FeatureProvider featureProvider)
+        public async Task SetProviderAsync(string clientName, FeatureProvider featureProvider)
```
austindrenski, line 83:
```suggestion
        public ValueTask SetProviderAsync(string clientName, FeatureProvider featureProvider)
```

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
         {
             this.EventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
             await this._repository.SetProvider(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
```
austindrenski, line 86:
```suggestion
            return new ValueTask(this._repository.SetProvider(clientName, featureProvider, this.GetContext()));
```

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
```diff
@@ -42,7 +42,7 @@ public EvaluationStepDefinitions(ScenarioContext scenarioContext)
         {
             _scenarioContext = scenarioContext;
             var flagdProvider = new FlagdProvider();
-            Api.Instance.SetProvider(flagdProvider).Wait();
+            Api.Instance.SetProviderAsync(flagdProvider).Wait();
```
austindrenski, line 45:
```suggestion
            Api.Instance.SetProviderAsync(flagdProvider).AsTask().GetAwaiter().GetResult();
```

test/OpenFeature.Tests/ClearOpenFeatureInstanceFixture.cs
```diff
@@ -7,7 +7,7 @@ public ClearOpenFeatureInstanceFixture()
         {
             Api.Instance.SetContext(null);
             Api.Instance.ClearHooks();
-            Api.Instance.SetProvider(new NoOpFeatureProvider()).Wait();
+            Api.Instance.SetProviderAsync(new NoOpFeatureProvider()).Wait();
```
austindrenski, line 10:
```suggestion
            Api.Instance.SetProviderAsync(new NoOpFeatureProvider()).AsTask().GetAwaiter().GetResult();

```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -103,8 +103,8 @@ public void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_P
         {
             var openFeature = Api.Instance;
 
-            openFeature.SetProvider(new NoOpFeatureProvider());
-            openFeature.SetProvider(TestProvider.DefaultName, new TestProvider());
+            openFeature.SetProviderAsync(new NoOpFeatureProvider());
+            openFeature.SetProviderAsync(TestProvider.DefaultName, new TestProvider());
```
austindrenski, line 107: I think these tests should probably be made async. I've been seeing some flaky test failures and pet theory is un-awaited tasks are the culprit.

- [ ] `OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider`
- [ ] `OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided`
- [ ] `OpenFeature_Should_Assign_Provider_To_Existing_Client`
- [ ] `OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance`
- [ ] `OpenFeature_Should_Get_Metadata`
- [ ] `OpenFeature_Should_Allow_Multiple_Client_Mapping`

austindrenski (changes requested):
Nice, should provide a gentler migration path downstream.

Couple of suggestions, including a while-we're-at-it related to #185 🤞

lukas-reining:
> I like the idea of https://github.com/open-feature/dotnet-sdk/pull/184, but I think I'd like us all to take some time to discuss exactly what we expect the cancellationToken to do in all circumstances (what should the SDK do on SDK methods like init()/shutdown()?, and what should providers do?).

I like the idea @toddbaert, and I think we could also add them later without even having a breaking change.
I think especially for getting flags we should later add cancellationTokens but maybe as you said with first discussing all the possible side effects.

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
+        /// Sets the default feature provider to given clientName without awaiting its initialization.
+        /// </summary>
+        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
+        /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
+        [Obsolete("Will be removed in later versions; use SetProviderAsync, which can be awaited")]
+        public void SetProvider(FeatureProvider featureProvider)
+        {
+            this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
+            this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
+        }
```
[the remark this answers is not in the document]
toddbaert: Nice catch. Yes, this was just copy/paste - the PR was more of a "thinking out loud" experiment to resolve our release conundrum.

src/OpenFeature/Api.cs
```diff
@@ -38,25 +39,48 @@ public sealed class Api : IEventBus
...
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProvider(FeatureProvider featureProvider)
+        public async Task SetProviderAsync(FeatureProvider featureProvider)
         {
             this.EventExecutor.RegisterDefaultFeatureProvider(featureProvider);
             await this._repository.SetProvider(featureProvider, this.GetContext()).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Sets the feature provider to given clientName without awaiting its initialization.
```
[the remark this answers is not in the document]
toddbaert: > I feel like it would be safer/more consistent with prior behavior to block on these rather than let the tasks go unobserved.

I'm not sure I agree. The purpose of the SDK's events is to be able to attach handlers, one of which is "PROVIDER_READY". Some implementations have non blocking provider settings, and then use an event to understand when to evaluate flags.

test/OpenFeature.Tests/OpenFeatureTests.cs
toddbaert: I also saw some flaky tests, but I think they are all to do with the new events functionality, and potentially related to the fact we have some handlers we don't clean up.

src/OpenFeature/Api.cs
toddbaert: I can see the benefit of using `ValueTask` for evaluations, but I'm not sure I do here. From the .NET 8 [API docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-8.0):

> the default choice for any asynchronous method that does not return a result should be to return a [Task](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-8.0). **Only if performance analysis proves it worthwhile should a ValueTask be used instead of a [Task](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-8.0)**.

Setting a provider is not on a "hot path" and it's potentially used in more complex logic where the additional utilities of a plain Task might be helpful.

toddbaert: I'm interested in what @benjiro and @kinyoklion think about using a `ValueTask` here.

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
austindrenski: Okay, that's interesting, and sounds like I may have misunderstood some intent with the async init changes (still wrapping my head around the wider spec).

Reframing the scenario I was thinking about before:

In #181, I'm registering an implementation of `IHostedLifecycleService` with the idea being to:
1. resolve:
   - `Api` from the container (i.e. currently a singleton registration of `Api.Instance`)
   - `IEnumerable<FeatureProvider>` (i.e. all registrations of type `FeatureProvider`)
2. perform async setup on the `Api` instance when the `IHost` calls `IHostedLifecycleService.StartingAsync(...)` which is the earliest hook we have inside of the `IHost` (i.e. before any calls are made to any `IHostedService.StartAsync(...)`)

Then as a consumer of `OpenFeature.Extensions.Hosting`, I can inject `IFeatureClient` into some `IHostedService` and feel confident knowing that the SDK is fully configured by the time the `IHost` calls my `IHostedService.StartAsync(...)` method.

Now this is somewhat moot for #181 since it has access to the new `Api.SetProviderAsync(...)` method and can thus choose to wait for the async init to complete, but #181 is based on some internal code I wrote, and if this change were to ship _without_ blocking, I have this itchy feeling that my first `IHostedService` could (theoretically?) end up using its injected `IFeatureClient` before the intended `FeatureProvider` has been added to the underlying `ProviderRepository._featureProviders`.

Again, I'm not arguing in favor of blocking (especially if the spec is ambivalent), just thinking out loud about how this _might_ surprise me in my existing code if I weren't aware of the change.

Given that our plan is to ship this sync-over-unawaited-async method at the same time as we're exposing the new async method, I'll just plan to migrate my existing usages to the async overloads when I upgrade (barring #181 also shipping at the same time), so no big deal, but that itchy feeling makes me think that we might want to call this out as a _we-dont-consider-this-breaking-but-take-note-before-you-ugrade_ in the release notes.

Or maybe at least suggest to contrib providers that if/when they opt-into async init, that they call it out as a breaking change in _their_ release notes. _(The `dotnet/efcore` team uses a `providers-beware` label for things like this where the breaking change may or may not impact consumers based on how the provider uses the feature before/after the change in question.)_

### Related snippets

https://github.com/open-feature/dotnet-sdk/blob/1efe09da3948d5dfd7fd9f1c7a040fc5c2cbe833/src/OpenFeature/OpenFeatureClient.cs#L39-L48

https://github.com/open-feature/dotnet-sdk/blob/1efe09da3948d5dfd7fd9f1c7a040fc5c2cbe833/src/OpenFeature/ProviderRepository.cs#L251-L261

https://github.com/open-feature/dotnet-sdk/blob/2eb7335eb3843de93dd5b0b05745ff0eacf58c16/src/OpenFeature.Extensions.Hosting/Internal/OpenFeatureHostedService.cs#L11-L36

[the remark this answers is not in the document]
toddbaert: > Given that our plan is to ship this sync-over-unawaited-async method at the same time as we're exposing the new async method, I'll just plan to migrate my existing usages to the async overloads when I upgrade (barring https://github.com/open-feature/dotnet-sdk/pull/181 also shipping at the same time), **_so no big deal, but that itchy feeling makes me think that we might want to call this out as a we-dont-consider-this-breaking-but-take-note-before-you-ugrade in the release notes._**

Yes, absolutely. The thing to note though that might make you feel better, is that since we haven't released init/shutdown yet, there really aren't semantics around provider readiness in the wild yet in .NET (it's conformant to a very old OpenFeature spec); there's basically no existing provider lifecycle defined. As of now, providers are for all practical purposes assumed to be ready when you set them.

Or maybe at least suggest to contrib providers that if/when they opt-into async init, that they call it out as a breaking change in their release notes. (The dotnet/efcore team uses a providers-beware label for things like this where the breaking change may or may not impact consumers based on how the provider uses the feature before/after the change in question.)

100%.

I would suggest that all providers in our contribs would be upgraded to require at minimum this new release of the SDK, and consider that a breaking change so that it's captured in their release notes. We could do that with a single PR in the contribs (providers hosted elsewhere should take note, and we can reach out to those responsible for them).

src/OpenFeature/Api.cs
askpt: Just to add: https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/

Honestly, I don't see much of a gain here. I don't see setting a provider to be a high-performance bottleneck, especially ideally you set the provider during the application startup.

austindrenski: > I can see the benefit of using ValueTask for flag evaluations and provider resolvers, but I'm not sure I do here. From the .NET 8 [API docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-8.0):
>
>> [...]
>
> Setting a provider is not on a "hot path" and it's potentially used in more complex logic where the additional utilities of a plain Task might be helpful.

That's a good point, especially considering that, while I'm a fan of supporting multi-providers, the most common steady-state (i.e. outside of slow, in-place migrations) will most likely see apps with a singular provider.

I'm happy to hear we're open to `ValueTask` on other, more obvious, hot-paths, so I work with the stance that `we don't consider async init a hot path`, if that's where the consensus ends up.

The only context I want to add is that I read `[...] default choice for any [...]` as more so speaking caution to consumer apps (e.g. don't make every method in your MVC web app return `ValueTask`) vs to library/SDK authors seeking to be as low-impact as possible.

So, if I _were_ going to make the argument in favor of `ValueTask` in our async init logic, it would be that these paths might not be hot, but in most uses, they'll probably precede all other app logic, and being lightweight on hosting startup can be nice in some (arguably niche) use cases (e.g. cold start lambdas, CLI apps, etc).

_(If anyone actually wants takes this^ "nice use cases" argument seriously, we could write benchmarks to see if it even matters, but tbh, I'm not sure I'm even taking it that seriously after having written it out lol.)_

kinyoklion: `ValueTask` does have merit, and I would be open to it for some demonstrable cases, I don't think this would be a case considerable improved by it. Considering it is a matter of allocations, and that `ValueTask` isn't flatly fewer allocations, it would make sense for high frequency use cases where we suspect they will usually be able to complete synchronously (if they are actually consistently async, then it probably doesn't help). This seems reasonable for evaluations where most of the time, most providers, would be satisfying the request from cache.

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
toddbaert: I've removed the `.ConfigureAwait(false);` and added a discard to avoid [CS4014](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014).

src/OpenFeature/Api.cs
toddbaert: Same as https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1457845424

src/OpenFeature/Api.cs
toddbaert: See thread [here](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456538431).

src/OpenFeature/Api.cs
toddbaert: See thread [here](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456538431).

src/OpenFeature/Api.cs
toddbaert: See thread [here](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456538431).

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
toddbaert: See thread [here](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456538431).

test/OpenFeature.Tests/OpenFeatureTests.cs
toddbaert: I will make these async though, there's no reason not to, you're right. They probably should have been when `SetProvider` was made async.

I _think_ the test flakiness is coming from elsewhere as I said above, but let's see. :+1:

test/OpenFeature.Tests/ClearOpenFeatureInstanceFixture.cs
toddbaert: See thread [here](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456538431).

toddbaert:
@austindrenski I've accepted a few suggestions, and noted that others, which I did not accept are corollaries to [this discussion](https://github.com/open-feature/dotnet-sdk/pull/190#discussion_r1456537287) and don't apply if we go with the route proposed there.

test/OpenFeature.Tests/OpenFeatureTests.cs
toddbaert: @austindrenski you may have been right about the test flakiness being due to missed awaits in tests after we made `SetProvider` async. I've run the tests a few times and they seem stable so far :sunglasses: .

austindrenski: 🍻🍻🍻🍻

austindrenski (approved):
LGTM 🚢🚢🚢🚢

austindrenski:
@toddbaert wrote (https://github.com/open-feature/dotnet-sdk/pull/190#issuecomment-1899033139):

> Based on the slack discussion, it sounds like we'd like to go the route of merging this to release a non-breaking version, and then subsequently working on a breaking release with a bunch of async changes, including moving some methods over to `ValueTask` as you proposed.

Sounds good, but would like to flag #186 for consideration before we cut the non-breaking release.

I opened #188 with a proposed solution, but want to call out that I drafted it with the assumption that the next release would be a breaking release, so it would help to get some fresh eyes on it.

The downside to not fixing #186 before the next release is that shutdown/restart will remain broken, but since its already broken, it's not a showstopper if we need to punt.

askpt (approved):
LGTM! Unfortunately, our test coverage decreased due to the new methods being introduced.

toddbaert:
> @toddbaert wrote ([#190 (comment)](https://github.com/open-feature/dotnet-sdk/pull/190#issuecomment-1899033139)):
> 
> > Based on the slack discussion, it sounds like we'd like to go the route of merging this to release a non-breaking version, and then subsequently working on a breaking release with a bunch of async changes, including moving some methods over to `ValueTask` as you proposed.
> 
> Sounds good, but would like to flag #186 for consideration before we cut the non-breaking release.
> 
> I opened #188 with a proposed solution, but want to call out that I drafted it with the assumption that the next release would be a breaking release, so it would help to get some fresh eyes on it.
> 
> The downside to not fixing #186 before the next release is that shutdown/restart will remain broken, but since its already broken, it's not a showstopper if we need to punt.

I will take a look, this is the sort of thing I'd like to have resolved in this release. :pray:

test/OpenFeature.Tests/OpenFeatureTests.cs
toddbaert: ugh nope, I got it again. I'm fairly confident I've actually fixed it here: https://github.com/open-feature/dotnet-sdk/pull/194

austindrenski (approved):

test/OpenFeature.Tests/OpenFeatureEventTests.cs
```diff
@@ -137,7 +137,7 @@ public async Task API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_
         [Specification("5.2.4", "The `handler function` MUST accept a `event details` parameter.")]
         [Specification("5.3.1", "If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers MUST run.")]
         [Specification("5.3.3", "Handlers attached after the provider is already in the associated state, MUST run immediately.")]
-        public void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
+        public async void API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
```
austindrenski, line 139: Ope, careful, in the past xUnit has gotten weird around `async void`. 
```suggestion
        public async Task API_Level_Event_Handlers_Should_Be_Informed_About_Ready_State_After_Registering_Provider_Ready_Sync()
```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -91,12 +91,12 @@ public async Task OpenFeature_Should_Support_Shutdown()
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()
+        public async void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()
```
austindrenski, line 94:
```suggestion
        public async Task OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()

```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -107,11 +107,11 @@ public void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_P
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
+        public async void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
```
austindrenski, line 110:
```suggestion
        public async Task OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()

```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -120,26 +120,26 @@ public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Assign_Provider_To_Existing_Client()
+        public async void OpenFeature_Should_Assign_Provider_To_Existing_Client()
```
austindrenski, line 123:
```suggestion
        public async Task OpenFeature_Should_Assign_Provider_To_Existing_Client()

```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -120,26 +120,26 @@ public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
...
+            await openFeature.SetProviderAsync(name, new TestProvider()).ConfigureAwait(true);
+            await openFeature.SetProviderAsync(name, new NoOpFeatureProvider()).ConfigureAwait(true);
 
             openFeature.GetProviderMetadata(name).Name.Should().Be(NoOpProvider.NoOpProviderName);
         }
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()
+        public async void OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()
```
austindrenski, line 136:
```suggestion
        public async Task OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()

```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -226,12 +226,12 @@ public void Should_Always_Have_Provider()
         }
 
         [Fact]
-        public void OpenFeature_Should_Allow_Multiple_Client_Mapping()
+        public async void OpenFeature_Should_Allow_Multiple_Client_Mapping()
```
austindrenski, line 229:
```suggestion
        public async Task OpenFeature_Should_Allow_Multiple_Client_Mapping()

```

test/OpenFeature.Tests/ProviderRepositoryTests.cs
```diff
@@ -16,24 +16,24 @@ namespace OpenFeature.Tests
     public class ProviderRepositoryTests
     {
         [Fact]
-        public void Default_Provider_Is_Set_Without_Await()
+        public async void Default_Provider_Is_Set_Without_Await()
```
austindrenski, line 19:
```suggestion
        public async Task Default_Provider_Is_Set_Without_Await()
```

test/OpenFeature.Tests/ProviderRepositoryTests.cs
```diff
@@ -182,24 +182,24 @@ public async Task AfterError_Is_Called_For_Shutdown_That_Throws()
         }
 
         [Fact]
-        public void Named_Provider_Provider_Is_Set_Without_Await()
+        public async void Named_Provider_Provider_Is_Set_Without_Await()
```
austindrenski, line 185:
```suggestion
        public async Task Named_Provider_Provider_Is_Set_Without_Await()
```

test/OpenFeature.Tests/ProviderRepositoryTests.cs
```diff
@@ -182,24 +182,24 @@ public async Task AfterError_Is_Called_For_Shutdown_That_Throws()
...
             var provider = new NoOpFeatureProvider();
             var context = new EvaluationContextBuilder().Build();
-            repository.SetProvider("the-name", provider, context);
+            await repository.SetProvider("the-name", provider, context);
             Assert.Equal(provider, repository.GetProvider("the-name"));
         }
 
         [Fact]
-        public void AfterSet_Is_Invoked_For_Setting_Named_Provider()
+        public async void AfterSet_Is_Invoked_For_Setting_Named_Provider()
```
austindrenski, line 195:
```suggestion
        public async Task AfterSet_Is_Invoked_For_Setting_Named_Provider()
```

austindrenski (changes requested):

toddbaert:
There was some fun pseudo-conflicts on this branch after recent merges, but I think we should be good now.

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -91,12 +91,13 @@ public async Task OpenFeature_Should_Support_Shutdown()
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()
+        public async Task OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()
+
```
austindrenski, line 95:
```suggestion
        public async Task OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_Provider()
```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -107,11 +108,12 @@ public void OpenFeature_Should_Not_Change_Named_Providers_When_Setting_Default_P
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
+        public async Task OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
+
```
austindrenski, line 112:
```suggestion
        public async Task OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -120,26 +122,28 @@ public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Assign_Provider_To_Existing_Client()
+        public async Task OpenFeature_Should_Assign_Provider_To_Existing_Client()
+
```
austindrenski, line 126:
```suggestion
        public async Task OpenFeature_Should_Assign_Provider_To_Existing_Client()
```

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -226,12 +230,13 @@ public void Should_Always_Have_Provider()
         }
 
         [Fact]
-        public void OpenFeature_Should_Allow_Multiple_Client_Mapping()
+        public async Task OpenFeature_Should_Allow_Multiple_Client_Mapping()
+
```
austindrenski, line 234:
```suggestion
        public async Task OpenFeature_Should_Allow_Multiple_Client_Mapping()
```

austindrenski (changes requested):
Little follow-up cleanup because GitHub got weird applying those suggestions

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -120,26 +120,27 @@ public void OpenFeature_Should_Set_Default_Provide_When_No_Name_Provided()
...
+            await openFeature.SetProviderAsync(name, new NoOpFeatureProvider()).ConfigureAwait(true);
 
             openFeature.GetProviderMetadata(name).Name.Should().Be(NoOpProvider.NoOpProviderName);
         }
 
         [Fact]
         [Specification("1.1.3", "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.")]
-        public void OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()
+        public async Task OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()
+
```
austindrenski, line 137: idk why it keeps doing this, but if it does it again we can just leave it for later cleanup...

```suggestion
        public async Task OpenFeature_Should_Allow_Multiple_Client_Names_Of_Same_Instance()
```

austindrenski (approved):
One more cleanup because the UI is clearly having some issues this morning, but otherwise LGTM (again)

toddbaert:
> One more cleanup because the UI is clearly having some issues this morning, but otherwise LGTM (again)

At least we can be reasonably satisfied the tests are stable now, they've run like 30 times this morning due to this PR :joy:

austindrenski (approved):

beeme1mr (approved):

PR #196, opened 2024-01-22
chore: Sync ci.yml with contrib repo

austindrenski:
Follows #173 and open-feature/dotnet-sdk-contrib#134 to synchronize `ci.yml` between both repos.

```console
$ diff dotnet-sdk{,-contrib}/.github/workflows/ci.yml


```

____

Story here is a little convoluted, but basically this repo already had the linux and windows tests in one workflow, so for #173 I was able to just condition the new `packaging` job on both `unit-tests-linux` and `unit-tests-windows` completing successfully.

However, prior to open-feature/dotnet-sdk-contrib#134, the contrib repo had separate workflow files for its respective linux and windows tests, so as part of open-feature/dotnet-sdk-contrib#134, I merged those separate workflows into one workflow and, at the same time, rewrote them to use a job matrix.

For PR consistency, I almost included the same job matrix update as part of #173, but held off in the spirit of doing one thing at a time. So, with that, this PR now syncs the two `ci.yml` files so both will exactly match _after_ open-feature/dotnet-sdk-contrib#134 is merged (anticipated sometime today).

Additionally, this PR also includes the `nuget.config` and `CONTRIBUTING.md` updates from open-feature/dotnet-sdk-contrib#134. Arguably, these may not be needed by this repo, but I'm including them here for (1) consistency, (2) to promote visibility, and (3) to help advertise that these are available if required and/or helpful to future contributors/bug reporters.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Test
+name: CI
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


austindrenski:
Looking for one more approval so we can merge this and address the code coverage failure @benjiro spotted via #219 🙏

askpt (approved):

PR #202, opened 2024-01-23
build: Update OpenFeature.sln

askpt:
## This PR

- This PR fixes an issue related to opening the project in VSCode.

### Related Issues

See log output here: 
```
Failed at 'Open a solution with environment service' (5ms) with error: Error: An item with the same key has already been added. Key: 90e7ead3-251e-4490-af78-e758e33518e5
Failed at 'Open a solution' (8ms) with error: SolutionOpenError: Run into errors while opening the solution: Error: An item with the same key has already been added. Key: 90e7ead3-251e-4490-af78-e758e33518e5
Failed to open solution SolutionOpenError: Run into errors while opening the solution: Error: An item with the same key has already been added. Key: 90e7ead3-251e-4490-af78-e758e33518e5
```

### How to test

Just open VSCode and check it's working again.

### DIFF
--- a/OpenFeature.sln
+++ b/OpenFeature.sln
@@ -36,7 +36,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenFeature.Tests", "test\O
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenFeature.Benchmarks", "test\OpenFeature.Benchmarks\OpenFeature.Benchmarks.csproj", "{90E7EAD3-251E-4490-AF78-E758E33518E5}"
 EndProject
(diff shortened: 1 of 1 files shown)


austindrenski (changes requested):
I guess someone must have originally added this manually, then?

@askpt can you run these commands locally, and assuming they do what I expect, check in the results?

```
git reset @~ --hard
dotnet sln remove test/OpenFeature.E2ETests/OpenFeature.E2ETests.csproj
dotnet sln add test/OpenFeature.E2ETests/OpenFeature.E2ETests.csproj
```

toddbaert (approved):
I noticed this too during a review, but was able to work around it by disabling an extension. I was going to dig into it but assumed it was just me. Thanks.

PR #219, opened 2024-01-27
fix: Remove reference to NETFramework.ReferenceAssemblies

benjiro:
## This PR
Since turning on CPM the code coverage build has broken, see [here](https://github.com/open-feature/dotnet-sdk/actions/runs/7676368996/job/20923831827?pr=217)

Believe we don't need to reference this anymore as its being implicitly included

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,6 @@
   <ItemGroup Label="src">
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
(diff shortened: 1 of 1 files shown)


austindrenski:
@benjiro I'm not sure why `dotnet test` is reporting an error when including the reference assemblies in that report workflow.

AFAIK those reference assemblies don't come for free with the .NET SDK on Linux, which is why we have to include them manually. I know that was at least true in the past, so I guess it might have changed, but idk...

Any chance it's something about codecover being included?

austindrenski:
Wait scratch that, looks like this workflow might just have gotten messed up?

It looks like it's referencing a matrix variable without defining the strategy matrix:

austindrenski:
Going to try addressing this as part of #196 which is already cleaning up some other matrixes

Issue #165, opened 2023-12-21
[FEATURE] add in-memory provider to SDK


toddbaert:
Add `InMemoryProvider` to SDK. See [spec](https://openfeature.dev/specification/appendix-a#in-memory-provider).

Tasks:

- add provider
- add tests
- replace e2e tests [here](https://github.com/open-feature/dotnet-sdk/tree/main/test/OpenFeature.E2ETests) with `InMemoryProvider`.

see [Java equivalent](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java).

Parent issue: https://github.com/open-feature/dotnet-sdk/issues/126

CommCody:
fwiw @toddbaert I'm looking at this, limited bandwidth, but interested in completing this

toddbaert:
Hey @CommCody ! I've assigned you.

It's no problem if this takes some time. I'll keep it assigned to you for now.

NeerajGartia21:
@CommCody are you working on this?
This issue looks interesting and I will be happy to collaborate with you.

CommCody:
Mildly, I do have some work on my fork. My time is limited however, so I'm not very actively working on it.

NeerajGartia21:
Can I continue from where you left?

toddbaert:
> Can I continue from where you left?

@NeerajGartia21 if @CommCody has a partial implementation, maybe you'd be better off working on the Java reconnect tests, or on the Error enhancements in JS you recently commented on.

NeerajGartia21:
Cool, that sounds good.

CommCody:
Hi @toddbaert I was updating my code to further implement this issue, but now I can't build the main branch anymore. the build fails with:
> E:\openfeature-dotnet-sdk\test\OpenFeature.Tests\OpenFeature.Tests.csproj : error NU1009: The packages Microsoft.NETFramework.ReferenceAssemblies are implicitly referenced. You do not typically need to refer
ence them from your project or in your central package versions management file. For more information, see https://aka.ms/sdkimplicitrefs
> E:\openfeature-dotnet-sdk\src\OpenFeature\OpenFeature.csproj : error NU1009: The packages Microsoft.NETFramework.ReferenceAssemblies are implicitly referenced. You do not typically need to reference them fro 
[...]

PR #222, opened 2024-01-29
fix: Fix NU1009 reference assembly warning

austindrenski:
Very silly, had this right in my internal projects and decided to "simplify" it when I reimplemented here 🤦🤦🤦

tldr; NuGet analyzers run over the `Directory.Packages.props` file without validating whether a given `PackageVersion` is used in a given TFM, so the analyzer pops even though the build is happy. I'm still a little confused as to why I can't repro locally, but probably caching, so "un-simplifying" my earlier oops and all should hopefully work as intended going forward.

See: #165, #196, #219, #220

H/T @CommCody


beeme1mr (approved):

askpt (approved):

austindrenski:
@CommCody please pull the latest `main` and let us know if you have any further build issues. Thanks again for reporting!

CommCody:
@austindrenski I can confirm main builds for me now

PR #223, opened 2024-01-30
feat: Flag metadata

### LINKED ISSUE #192: [FEATURE] Implement flag metadata
### Requirements

Check https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a

askpt:
## Flag Metadata

- Adds support for flag metadata

### Related Issues

Fixes #192 

### Notes

- Currently, an invalid flag evaluation (cast) will throw an exception. I was thinking of returning  `null` and logging it as a warning. For that effort, I need a log source (#189).

### Follow-up Tasks

- https://github.com/open-feature/dotnet-sdk/issues/234


src/OpenFeature/Model/FlagMetadata.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+#nullable enable
+namespace OpenFeature.Model;
+
+public sealed class FlagMetadata
```
toddbaert, line 8: It would be nice to have a more generic metadata class that we could use for `EventMetadata` as well: https://github.com/open-feature/dotnet-sdk/blob/7eebcdda123f9a432a8462d918b7454a26d3e389/src/OpenFeature/Model/ProviderEvents.cs#L39C43-L39C56

The difficulty is that the obvious generic name `Metadata` is already taken.

toddbaert:
@askpt `FlagMetadata` and `EventMetadata` are shaped the same in the SDK:

https://openfeature.dev/specification/types#event-metadata
https://openfeature.dev/specification/types#flag-metadata

I think for consistency, that means we should use the same type for these, which I guess would mean it should be a `Dictionary<string, object>()`, unless there's some way we can bring them in line with your new implementation in a non-breaking way.

askpt:
> @askpt `FlagMetadata` and `EventMetadata` are shaped the same in the SDK:
> 
> https://openfeature.dev/specification/types#event-metadata https://openfeature.dev/specification/types#flag-metadata
> 
> I think for consistency, that means we should use the same type for these, which I guess would mean it should be a `Dictionary<string, object>()`, unless there's some way we can bring them in line with your new implementation in a non-breaking way... maybe with extension methods?

I added a TODO to change the dictionary to use the `BaseMetadata` type. This would be a breaking change.

toddbaert:
@askpt merged the recent changes to remove the flagd dep and now e2e looks good!

src/OpenFeature/Model/ProviderEvents.cs
```diff
@@ -36,6 +36,7 @@ public class ProviderEventPayload
         /// <summary>
         /// Metadata information for the event.
         /// </summary>
+        // TODO: This needs to be changed to a EventMetadata object
```
kinyoklion, line 39: I assume we are deferring this as it would be breaking?

If so, we may want to consider how we accumulate and act on such pending breaks?

askpt: Yes. This would bring breaking changes. As part of the massive breaking changes coming, I created this issue to track this effort: https://github.com/open-feature/dotnet-sdk/issues/234

I've been thinking about creating a milestone to track all the breaking changes so we can safely understand what needs to be done.

kinyoklion: Something like a milestone seems like a reasonable idea to me.

kinyoklion (approved):

toddbaert (approved):

PR #231, opened 2024-02-07
fix: Add targeting key

roelofb:
Fixes #230

PR #232, opened 2024-02-08
feat: implement in-memory provider

toddbaert:
Implements in-memory provider as per spec, updates gherkin to use spec version, removes flagd deps.

Thanks to @CommCody

I could have added more unit tests for the in-memory provider, but the e2e suite tests most of it. I'll create a "good first issue" for others to add more unit tests, but I'd like to get this merged sooner than later because I believe it's blocking [this](https://github.com/open-feature/dotnet-sdk/pull/223) and [this](https://github.com/open-feature/dotnet-sdk/pull/231)

Resolves: https://github.com/open-feature/dotnet-sdk/issues/165


test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
```diff
@@ -41,15 +42,14 @@ public class EvaluationStepDefinitions
...
 
-        [Given(@"a provider is registered with cache disabled")]
-        public void Givenaproviderisregisteredwithcachedisabled()
+        [Given(@"a provider is registered")]
+        public void GivenAProviderIsRegistered()
         {
-
+            var memProvider = new InMemoryProvider(e2eFlagConfig);
+            Api.Instance.SetProviderAsync(memProvider).Wait();
+            client = Api.Instance.GetClient();
```
toddbaert, line 52: All the tests now use the in-memory provider (configuration is at the end of this file).

src/OpenFeature/Providers/Memory/Flag.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace OpenFeature.Providers.Memory
+{
+    /// <summary>
+    /// Flag representation for the in-memory provider.
+    /// </summary>
+    public class Flag
+    {
+
+    }
```
askpt, line 15: Why do we need this empty class? Might be an `interface` or `abstract class`?

src/OpenFeature/Providers/Memory/Flag.cs
```diff
@@ -0,0 +1,78 @@
...
+                    );
+                }
+                else
+                {
+                    throw new GeneralException($"variant {this.DefaultVariant} not found");
+                }
+            }
+            else
+            {
+                string variant = this.ContextEvaluator.Invoke(evaluationContext);
```
askpt, line 60:
```suggestion
                var variant = this.ContextEvaluator.Invoke(evaluationContext);
```

In dotnet we love `var` 🤣

src/OpenFeature/Providers/Memory/Flag.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// </summary>
+    public class Flag
+    {
+
+    }
+
+    /// <summary>
+    /// Flag representation for the in-memory provider.
+    /// </summary>
+    public class Flag<T> : Flag
```
askpt, line 20:
```suggestion
    public sealed class Flag<T> : Flag
```

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+using OpenFeature.Model;
+
+namespace OpenFeature.Providers.Memory
+{
+    /// <summary>
+    /// The in memory provider.
+    /// Useful for testing and demonstration purposes.
+    /// </summary>
+    /// <seealso href="https://openfeature.dev/specification/appendix-a#in-memory-provider">In Memory Provider specification</seealso>
+    public class InMemoryProvider : FeatureProvider
```
askpt, line 17:
```suggestion
    public sealed class InMemoryProvider : FeatureProvider
```

src/OpenFeature/Providers/Memory/Flag.cs
```diff
@@ -0,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using OpenFeature.Constant;
+using OpenFeature.Error;
+using OpenFeature.Model;
+
```
askpt, line 6:
```suggestion

#nullable enable
```

Let's force the nullable validations/checks in the new class files so it would be easier later on to enable for the entire project.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
```diff
@@ -0,0 +1,129 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using OpenFeature.Constant;
+using OpenFeature.Error;
+using OpenFeature.Model;
+
```
askpt, line 9:
```suggestion

#nullable enable
```

Let's force the nullable validations/checks in the new class files so it would be easier later on to enable for the entire project.

askpt (commented):
Added a few minor comments. It seems the unit test coverage is failing due to the lack of testing when exceptions happen.

src/OpenFeature/Providers/Memory/Flag.cs
toddbaert: Good call. We just need a non-generic base but I can use an interface.

src/OpenFeature/Providers/Memory/Flag.cs
toddbaert: pardon my Java.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
toddbaert: I think I'd rather not seal this class unless you have a good reason for us to...

I've extended in-memory providers to create artificial delays for our React SDK testing (to test suspense), extended it to support "fake" events, etc. I think it's nice to be able to do this.

src/OpenFeature/Providers/Memory/Flag.cs
toddbaert: great idea.

toddbaert: This IMMEDIATELY helped me find at least 2 possible null pointer errors. Wow.

src/OpenFeature/Providers/Memory/Flag.cs
toddbaert: Made interface.

toddbaert:
@askpt Thanks for the review. Please check again when you have time: I added tests and applied all but one of your suggestions.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
askpt: Agreed! That's a valid point.

askpt (approved):

askpt:
> @askpt Thanks for the review. Please check again when you have time: I added tests and applied all but one of your suggestions.

@toddbaert LGTM! There are some checks still failing

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+        private ResolutionDetails<T> Resolve<T>(string flagKey, T defaultValue, EvaluationContext? context)
+        {
+            if (!this._flags.TryGetValue(flagKey, out var flag))
+            {
+                throw new FlagNotFoundException($"flag {flag} not found");
+            }
+            else
+            {
+                if (typeof(Flag<T>).Equals(flag.GetType()))
```
kinyoklion, line 126: So, this is maybe fine for an in-memory provider, but in general it gives me pause. In many typical cases the underlying transport has "numbers". So JSON, for instance, everything is basically a double. People then ask for that double as an integer or double.

kinyoklion: A person who is using this as a short-cut to an in-house provider will just need to be mindful.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+                {
+                    return ((Flag<T>)flag).Evaluate(flagKey, defaultValue, context);
+                }
+                else
+                {
+                    throw new TypeMismatchException($"flag {flag} not found");
```
kinyoklion, line 132: This looks to be the wrong message, copy paste from the `FlagNotFoundException`.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
toddbaert: Great point. I've seen this issue elsewhere. Any suggestions? It's possible people will look at this implementation as a reference. Is there anything that can be done that's more elegant than checking specifically if these are any kind of numbers and casting?

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
toddbaert: Nice catch.

src/OpenFeature/Providers/Memory/InMemoryProvider.cs
kinyoklion: I am not sure of a simple way to handle it. The way we handle it in LD SDKs is basically to always store doubles and then the int/double method use both use double. The int method truncates. Our data type is from JSON, so we don't lose fidelity compared to the source. Someone who actually supports ints this would be a limitation.

I am not sure I see an easy way to handle it here other than to have special handling for numbers.

Get the flag, check if it is the requested numeric type, if it isn't check if it is the other one, then do the conversion. Maybe it could be somewhat clean with switching on the type.

toddbaert: If you agree it's sufficient, I think I'll just add a comment explaining the shortcoming of this check, since the in-memory provider is is not a production tool. This will at least be edifying for anyone looking at the implementation for reference.

toddbaert: done:

```
// This check returns False if a floating point flag is evaluated as an integer flag, and vice-versa.
// In a production provider, such behavior is probably not desirable; consider supporting conversion.
```

kinyoklion (approved):

PR #245, opened 2024-03-10
ci: Generate SBOM

### LINKED ISSUE #159: Generate SBOMs for .NET components
We have SBOMs currently for Java and Go. We could use them here as well. I recommend this utility: https://github.com/marketplace/actions/cyclonedx-net-generate-sbom (we're using the clyclonedx format elsewhere and it's popular).

Definition of done:

- SBOMs generated and attached to release artifact in GH, or otherwise made publicly available (for every release)
- only includes dependencies of module in question (not of repo)

jenshenneberg:
## This PR
Generates Software Bill of Materials (SBOM) as described in #159. Once https://github.com/NuGet/Home/issues/12497 is implemented, the SBOM file(s) should be embedded in the published nuget packages. Until then, I've added the SBOM as an asset under the release. 

### Known issue
The SBOM file lists the dependences for all target frameworks combined. Once the above [NuGet ](https://github.com/NuGet/Home/issues/12497)issue is implemented, it should be changed, so there is one sbom created for each target framework with only the applicable references included.

### Related Issues
Fixes #159

### How to test
Unfortunately, this is somewhat cumbersome to test, as the logic in question only kicks in upon a release from the main branch. I've tested it myself this way:
- Create new fork of this repo
- Merge this branch to main in the new repo
- Create a release in the new repo


toddbaert:
Wow thanks @jenshenneberg ! I will review this carefully tomorrow.

.github/workflows/release.yml
```diff
@@ -34,13 +41,32 @@ jobs:
...
         run: dotnet pack --no-restore
 
       - name: Publish to Nuget
-        if: ${{ steps.release.outputs.releases_created }}
         run: dotnet nuget push "src/**/*.nupkg" --api-key "${{ secrets.NUGET_TOKEN }}" --source https://api.nuget.org/v3/index.json
+
+  sbom:
+    runs-on: ubuntu-latest
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created }}
```
beeme1mr, line 55: It may be worth enabling [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) until we've seen this job run successfully.

beeme1mr (approved):
Looks good but we may want to preview the CI from failing until it has run successfully.

.github/workflows/release.yml
jenshenneberg: It should not block the release even if it fails, but it's a good point regardless to keep the noise down in case of a failure. I'll add your suggestion.

jenshenneberg: I've added `continue-on-error` now 😃

beeme1mr (approved):
Awesome, thank you! 🎉

toddbaert (approved):
Ya this looks good to me, thanks for also extracting the conditions.

I'll merge tomorrow unless I hear objections from maintainers.

benjiro (approved):

askpt (approved):

PR #253, opened 2024-03-28
chore!: Enable nullable reference types

### LINKED ISSUE #208: Enable Nullable Reference Types
### Requirements

As part of C# 8.0, we have access to a Nullable Reference Types (NRT) feature. This allows the have a more explicit flow of nulls in the codebase.

> The NRT feature holds you to a higher standard on how you deal with nulls, and as such it issues new warnings on existing code. So that those warnings (however useful) don’t break you, the feature must be explicitly enabled in your code before it starts complaining. Once you do that on existing code, you have work to do to make that code null-safe and satisfy the compiler that you did.

Check: https://devblogs.microsoft.com/dotnet/embracing-nullable-reference-types/

### Possible breaking changes
This might bring some potential breaking changes for the providers and customers that consume our library. I think this should be released with a breaking change alert.

### Notes
To ease the migration, the new code created can use a preprocessor directive. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#nullable-context

### Potential tasks
```[tasklist]
### Tasks
- [x] Add nullable enable to the common.props
- [x] Fix the warnings related to nulls that arise from it
- [x] Review all public API surfaces
```

askpt:
## Enable nullable reference types

- This PR enables the nullable validation and treats warnings as errors.

### Related Issues

Fixes #208 

### Notes

This PR turns on the nullable checks for the dotnet checks. This gives us a better and safer codebase since it checks for potential null exceptions.

### Breaking changes
While this PR won't require changes to the exposed API, it might show some errors to the clients consuming it.

### DIFF
--- a/src/OpenFeature/ProviderRepository.cs
+++ b/src/OpenFeature/ProviderRepository.cs
@@ -63,7 +63,7 @@ namespace OpenFeature
         /// </param>
         /// <param name="afterShutdown">called after a provider is shutdown, can be used to remove event handlers</param>
         public async Task SetProvider(
-            FeatureProvider featureProvider,
+            FeatureProvider? featureProvider,
             EvaluationContext context,
             Action<FeatureProvider>? afterSet = null,
             Action<FeatureProvider>? afterInitialization = null,
@@ -105,10 +105,10 @@ namespace OpenFeature
         }
 
         private static async Task InitProvider(
-            FeatureProvider newProvider,
+            FeatureProvider? newProvider,
             EvaluationContext context,
-            Action<FeatureProvider> afterInitialization,
-            Action<FeatureProvider, Exception> afterError)
+            Action<FeatureProvider>? afterInitialization,
+            Action<FeatureProvider, Exception>? afterError)
         {
             if (newProvider == null)
             {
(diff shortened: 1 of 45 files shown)


beeme1mr (approved):

src/OpenFeature/ProviderRepository.cs
```diff
@@ -256,9 +261,9 @@ public FeatureProvider GetProvider()
             }
         }
 
-        public FeatureProvider GetProvider(string clientName)
+        public FeatureProvider GetProvider(string? clientName)
         {
-            if (string.IsNullOrEmpty(clientName))
+            if (clientName == null || string.IsNullOrEmpty(clientName))
```
kinyoklion, line 266: Looks like maybe this check got added when there was already a `IsNullOrEmpty` check. Or the intent here is unclear.

askpt: It is a strange change, but only the newer dotnet SDK has the proper null check.
See the error here:
Screenshot 2024-04-03 at 08 40 06
Screenshot 2024-04-03 at 08 38 47

kinyoklion: Ok. Apparently .netstandard and .net framework don't have nullable annotations for IsNullOrEmpty.

In cases like this I think we need a comment to indicate that. Otherwise it is something everyone will want to optimize away

askpt: I added a target framework conditional, so it is way more explicit why that null check needs to be there. This way, the next developer will know why! :)

I hope it addresses your concerns!

kinyoklion (approved):

PR #256, opened 2024-04-08
fix: Add missing error message when an error occurred

### LINKED ISSUE #255: [BUG] Missing error message when unknown exception is thrown
### Observed behavior

Before the version v1.5.0 when we were receiving a `GENERAL` error we had the error message associated to the error.

Since the v1.5.0 we are not receiving the error message anymore.

### Expected Behavior

Receive the error message associated with the error.

### Steps to reproduce

_No response_

thomaspoignant:
## This PR
- Add missing error message when an error occurred

### Related Issues
Fixes #255

### DIFF
--- a/src/OpenFeature/OpenFeatureClient.cs
+++ b/src/OpenFeature/OpenFeatureClient.cs
@@ -262,7 +262,7 @@ namespace OpenFeature
             {
                 this._logger.LogError(ex, "Error while evaluating flag {FlagKey}", flagKey);
                 var errorCode = ex is InvalidCastException ? ErrorType.TypeMismatch : ErrorType.General;
-                evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, errorCode, Reason.Error, string.Empty);
+                evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, errorCode, Reason.Error, string.Empty, ex.Message);
                 await this.TriggerErrorHooks(allHooksReversed, hookContext, ex, options).ConfigureAwait(false);
             }
             finally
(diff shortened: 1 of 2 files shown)


kinyoklion (approved):
Code coverage issue seems unrelated to this. 

Change looks good.

thomaspoignant:
> It would be nice to have a unit test for this condition.

I haven't worked on the SDK, and my knowledge of dotnet is limited so not sure how to trigger the exception in the test.
Any help is welcome.

kinyoklion:
> > It would be nice to have a unit test for this condition.
> 
> I haven't worked on the SDK, and my knowledge of dotnet is limited so not sure how to trigger the exception in the test. Any help is welcome.

Let me take a look.

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -262,7 +262,7 @@ public void AddHooks(IEnumerable<Hook> hooks)
             {
                 this._logger.LogError(ex, "Error while evaluating flag {FlagKey}", flagKey);
                 var errorCode = ex is InvalidCastException ? ErrorType.TypeMismatch : ErrorType.General;
```
kinyoklion, line 264: It looks like this line helps find how we could test this.

There is a test `OpenFeatureClient_Should_Return_DefaultValue_When_Type_Mismatch` which triggers this condition, and now it will have a message. 

```
            var evaluationDetails = await client.GetObjectDetails(flagName, defaultValue);
            evaluationDetails.ErrorMessage.Should().Be(new InvalidCastException().Message);
```

thomaspoignant: Oh, thanks it works 🙌 

Coverage is still not working, but it is not related to anything introduced by this PR.

askpt (approved):
LGTM. Is the coverage not working because of a missing token? I thought that if we kept the v3, it would never require 🤔

benjiro (approved):

PR #260, opened 2024-04-10
ci: Change dotnet formatter

askpt:
## This PR

- This PR changes the code formatter we currently use to the new built-in in the SDK. See https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format

### Notes
- We should now use this tool since it is now shipped as part of the dotnet SDK, removing the need to install an external one.
- As pointed out in the chat, this build runs in parallel and is relatively fast. I will enable all PRs instead of those focusing on changes to .cs files.

### DIFF
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -22,9 +22,8 @@ jobs:
 
     - name: Setup .NET SDK
       uses: actions/setup-dotnet@v4
-
-    - name: Install format tool
-      run: dotnet tool install -g dotnet-format
+      with:
+          dotnet-version: 8.0.x
 
     - name: dotnet format
-      run: dotnet-format --folder --check
+      run: dotnet format --verify-no-changes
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnet-format.yml
```diff
@@ -2,29 +2,22 @@ name: dotnet format
 
 on:
   push:
-    branches: [ main ]
-    paths:
-    - '**.cs'
-    - '.editorconfig'
```
beeme1mr, line 8: Why was this removed?

beeme1mr (approved):

.github/workflows/dotnet-format.yml
askpt: I removed it because this action runs relatively fast and in parallel. If there is any change to the action, we can see if it fails.

beeme1mr: Fair enough, thanks.

kinyoklion (approved):

PR #268, opened 2024-05-01
feat!: add CancellationTokens, ValueTasks hooks

toddbaert:
This PR is a combination of https://github.com/open-feature/dotnet-sdk/pull/184 and https://github.com/open-feature/dotnet-sdk/pull/185. Changes include:

- adding cancellation tokens
  - in all cases where async operations include side-effects (`setProviderAsync`, `InitializeAsync`, I've specified in the in-line doc that the cancellation token's purpose is to cancel such side-effects - so setting a provider and canceling that operation still results in that provider's being set, but async side-effect should be cancelled. I'm interested in feedback here, I think we need to consider the semantics around this... I suppose the alternative would be to always ensure any state changes only occur after async side-effects, if they weren't cancelled beforehand. 
- adding "Async" suffix to all async methods
- remove deprecated sync `SetProvider` methods 
- Using `ValueTask` for hook methods
  - I've decided against converting all `Tasks` to `ValueTasks`, from the [official .NET docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-8.0):
  > the default choice for any asynchronous method that does not return a result should be to return a [Task](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-8.0). Only if performance analysis proves it worthwhile should a ValueTask be used instead of a [Task](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-8.0). 
  - I think for hooks, `ValueTask` especially makes sense since often hooks are synchronous, in fact async hooks are probably the less likely variant.
  - I've kept the resolver methods as `Task`, but there could be an argument for making them `ValueTask`, since some providers resolve asynchronously. 
  - I'm still a bit dubious on the entire idea of `ValueTask`, so I'm really interested in feedback here
- associated test updates

UPDATE:

After chewing on this for a night, I'm starting to feel:
- We should simply remove cancellation tokens from Init/Shutdown. We can always add them later, which would be non-breaking. I think the value is low and the complexity is potentially high.
- ValueTask is only a good idea for hooks, because:
  - Hooks will very often be synchronous under the hood
  - We (SDK authors) await the hooks, not consumer code, so we can be careful of the potential pitfalls of ValueTask. I think everywhere else we should stick to Task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-<<<<<<< HEAD
 <!-- markdownlint-disable MD033 -->
 <!-- x-hide-in-docs-start -->
 <!-- NuGet doesn't support most HTML tags. Disabling dark mode support until https://github.com/NuGet/NuGetGallery/issues/8644 is resolved. -->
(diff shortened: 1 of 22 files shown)


src/OpenFeature/EventExecutor.cs
```diff
@@ -10,7 +10,9 @@
 
 namespace OpenFeature
 {
-    internal sealed partial class EventExecutor : IAsyncDisposable
+    internal delegate Task ShutdownDelegate(CancellationToken cancellationToken);
```
askpt, line 13: `IAsyncDisposable` was removed. I would encourage you to keep it since disposing of it will trigger the provider's shutdown.

src/OpenFeature/Model/EvaluationContext.cs
```diff
@@ -24,6 +24,16 @@ internal EvaluationContext(string? targetingKey, Structure content)
...
+        /// <summary>
+        /// Internal constructor used by the builder.
+        /// </summary>
+        /// <param name="content">The content of the context.</param>
+        internal EvaluationContext(Structure content)
+        {
+            this.TargetingKey = string.Empty;
+            this._structure = content;
+        }
+
```
askpt, line 36: Is this an intended change? It seems out of scope.

askpt (commented):
I left a few comments. It is a big diff, but most of it is just renaming.

Overall it looks good but I added some of my concerns.

src/OpenFeature/Api.cs
```diff
@@ -37,55 +37,30 @@ public sealed class Api : IEventBus
...
         /// <param name="clientName">Name of client</param>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProviderAsync(string clientName, FeatureProvider featureProvider)
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to cancel any async side effects.</param>
+        public async Task SetProviderAsync(string clientName, FeatureProvider featureProvider, CancellationToken cancellationToken = default)
         {
-            if (string.IsNullOrWhiteSpace(clientName))
-            {
-                throw new ArgumentNullException(nameof(clientName));
-            }
```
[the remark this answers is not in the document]
toddbaert: This wasn't intentional, it must be the result of a conflict I messed up.

src/OpenFeature/EventExecutor.cs
toddbaert: It didn't seem to me that there was simple way of leveraging IDisposableAsync with a cancellation token, so I removed this on favor of just shutdown so we can propagate the token. 

This is an internal class only used by us in a single location, so I didn't think that ability to use "using" for cleanup was that valuable.

README.md
```diff
@@ -1,320 +1,320 @@
...
+### Usage
+
+```csharp
+public async Task Example()
+{
+    // Register your feature flag provider
+    await Api.Instance.SetProvider(new InMemoryProvider());
+
+    // Create a new client
+    FeatureClient client = Api.Instance.GetClient();
```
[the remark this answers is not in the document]
toddbaert: Nice catch.

src/OpenFeature/Api.cs
```diff
@@ -248,18 +223,17 @@ public EvaluationContext GetContext()
         /// Once shut down is complete, API is reset and ready to use again.
         /// </para>
         /// </summary>
-        public async Task Shutdown()
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to cancel any async side effects.</param>
+        public async Task ShutdownAsync(CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
toddbaert: Ya, I think this is an important discussion. I opened this PR to try to get to the bottom of this. See conversation on slack: https://cloud-native.slack.com/archives/C06EDTWU25B/p1714590325481829?thread_ts=1714590325.481829&cid=C06EDTWU25B

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
toddbaert: Fixed

src/OpenFeature/Model/EvaluationContext.cs
toddbaert: Reverted, this was unintentional. Nice catch.

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
toddbaert: I'd be happy to remove cancellation tokens from `SetProviderAsync` and `ShutdownAsync` entirely, in which case it would be better to implement `IDisposableAsync`.

src/OpenFeature/EventExecutor.cs
toddbaert: Relevant: https://github.com/open-feature/dotnet-sdk/pull/268#discussion_r1587609537

toddbaert:
After chewing on this for a night, I'm starting to feel:

We should simply remove cancellation tokens from Init/Shutdown. We can always add them later, which would be non-breaking. I think the value is low and the complexity is potentially high.
ValueTask is only a good idea for hooks, because:
Hooks will very often be synchronous under the hood
-We (SDK authors) await the hooks, not consumer code, so we can be careful of the potential pitfalls of ValueTask.
I think everywhere else we should stick to Task.

src/OpenFeature/EventExecutor.cs
askpt: I prefer to keep the `Dispose`.

I would do something like this:
```
public ValueTask DisposeAsync() => new(this.Shutdown(CancellationToken.None));
```

When disposing of the objects, we shouldn't have any cancellation tokens. 

What are your thoughts?

toddbaert: > public ValueTask DisposeAsync() => new(this.Shutdown(CancellationToken.None));

Yes, that works syntactically, but it doesn't propagate the cancellation token from the original method the user called (`.ShutdownAsync`) so there's no point to involving the cancellation at all. In this case, we might as well remove the cancellation token entirely from this part of the API: https://github.com/open-feature/dotnet-sdk/pull/268#issuecomment-2090536154

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
toddbaert: Let's wait to see what others say, but it seems to me that we can just remove CTs from init/shutdown, and go back to `IDisposable`. Let me know if you disagree.

src/OpenFeature/EventExecutor.cs
toddbaert: I've removed the cancellation and made this `IAsyncDisposable` again.

src/OpenFeature/Api.cs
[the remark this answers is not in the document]
toddbaert: I've removed cancellation from `ShutdownAsync`

kinyoklion (commented):
I think that we should probably have tests which have a dummy async provider that blocks evaluations until cancelled, then we should start an evaluation and cancel it to make sure all the steps happen. Maybe something similar with a hook implementation.

toddbaert:
> I think that we should probably have tests which have a dummy async provider that blocks evaluations until cancelled, then we should start an evaluation and cancel it to make sure all the steps happen. Maybe something similar with a hook implementation.

That's a good idea. Will add.

toddbaert:
> > I think that we should probably have tests which have a dummy async provider that blocks evaluations until cancelled, then we should start an evaluation and cancel it to make sure all the steps happen. Maybe something similar with a hook implementation.
> 
> That's a good idea. Will add.

I've added tests for cancellation (evaluation/resolution and hooks): https://github.com/open-feature/dotnet-sdk/pull/268/commits/036d7a45c5627c4fb17ce7ab7208cd52ca19b7ca

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
```diff
@@ -39,17 +39,17 @@ public EvaluationStepDefinitions(ScenarioContext scenarioContext)
         }
 
         [Given(@"a provider is registered")]
-        public void GivenAProviderIsRegistered()
+        public async void GivenAProviderIsRegistered()
```
askpt, line 42:
```suggestion
        public async Task GivenAProviderIsRegistered()
```

test/OpenFeature.Tests/Providers/Memory/InMemoryProviderTests.cs
```diff
@@ -111,7 +111,7 @@ public InMemoryProviderTests()
         [Fact]
         public async void GetBoolean_ShouldEvaluateWithReasonAndVariant()
```
askpt, line 112:
```suggestion
        public async Task GetBoolean_ShouldEvaluateWithReasonAndVariant()
```

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
askpt, line 1: There are a lot of changes here where we are not awaiting the get operation.

test/OpenFeature.Tests/Providers/Memory/InMemoryProviderTests.cs
askpt, line 1: This class has a lot of `async void`, which we should avoid—I added a suggestion to explain my point.

askpt (approved):
I left a few suggestions around the tests.
We should `await` the async operations, and we should not use `async void` but `async Task`.

askpt (commented):
Reseting my approval. Sorry! 😃

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
toddbaert: Thanks, nice catches! I will update.

toddbaert: Actually, most of these aren't awaited, but the values are being assigned to Tasks, with the result called on.

Basically the e2e tests run synchronously, which IMO is fine, perhaps even better because of their external dependencies.

test/OpenFeature.Tests/Providers/Memory/InMemoryProviderTests.cs
toddbaert: Corrected!

kinyoklion (approved):

askpt (approved):

Issue #250, opened 2024-03-15
[FEATURE] Make provider interface "stateless", SDK maintains provider state


toddbaert:
#### Stateless providers

Providers no longer maintain their own state: the state for each provider is maintained in the SDK automatically, and updated according to the success/failures of lifecycle methods (init/shutdown) or events emitted from providers spontaneously.

```[tasklist]
### Tasks
- [ ] [Requirement 1.7.1](https://openfeature.dev/specification/sections/flag-evaluation#requirement-171): The client MUST define a provider status accessor which indicates the readiness of the associated provider, with possible values NOT_READY, READY, STALE, ERROR, or FATAL.
- [ ] [Requirement 1.7.3](https://openfeature.dev/specification/sections/flag-evaluation#requirement-173): The client's provider status accessor MUST indicate READY if the initialize function of the associated provider terminates normally.
- [ ] [Requirement 1.7.4](https://openfeature.dev/specification/sections/flag-evaluation#requirement-174): The client's provider status accessor MUST indicate ERROR if the initialize function of the associated provider terminates abnormally.
- [ ] [Requirement 1.7.5](https://openfeature.dev/specification/sections/flag-evaluation#requirement-175): The client's provider status accessor MUST indicate FATAL if the initialize function of the associated provider terminates abnormally and indicates error code PROVIDER_FATAL.
- [ ] [Requirement 1.7.6](https://openfeature.dev/specification/sections/flag-evaluation#requirement-176): The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in NOT_READY.
- [ ] [Requirement 1.7.7](https://openfeature.dev/specification/sections/flag-evaluation#requirement-177): The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in FATAL.
- [ ] [Requirement 1.7.8](https://openfeature.dev/specification/sections/flag-evaluation#requirement-178): Implementations SHOULD propagate the error code returned from any provider lifecycle methods.
- [ ] [Requirement 5.1.5](https://openfeature.dev/specification/sections/events#requirement-515): PROVIDER_ERROR events SHOULD populate the provider event details's error code field.
- [ ] [Requirement 5.3.5](https://openfeature.dev/specification/sections/events#requirement-535): If the provider emits an event, the value of the client's provider status MUST be updated accordingly.
```

toddbaert:
I've started on this.

PR #276, opened 2024-06-12
feat!: internally maintain provider status

toddbaert:
This PR implements a few things from spec 0.8.0:

- implements internal provider status (already implemented in JS)
  - the provider no longer updates its status to READY/ERROR, etc after init (the SDK does this automatically)
  - the provider's state is updated according to the last event it fired
- adds `PROVIDER_FATAL` error and code
- adds "short circuit" feature when evaluations are skipped if provider is `NOT_READY` or `FATAL`
- removes some deprecations that were making the work harder since we already have pending breaking changes.

Fixes: https://github.com/open-feature/dotnet-sdk/issues/250

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -46,7 +46,7 @@ namespace OpenFeature
         public async Task SetProviderAsync(FeatureProvider featureProvider)
         {
             this._eventExecutor.RegisterDefaultFeatureProvider(featureProvider);
-            await this._repository.SetProviderAsync(featureProvider, this.GetContext(), afterInitialization, afterError).ConfigureAwait(false);
+            await this._repository.SetProviderAsync(featureProvider, this.GetContext(), AfterInitialization, AfterError).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -62,7 +62,7 @@ namespace OpenFeature
                 throw new ArgumentNullException(nameof(clientName));
             }
             this._eventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
-            await this._repository.SetProviderAsync(clientName, featureProvider, this.GetContext(), afterInitialization, afterError).ConfigureAwait(false);
+            await this._repository.SetProviderAsync(clientName, featureProvider, this.GetContext(), AfterInitialization, AfterError).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -258,6 +258,7 @@ namespace OpenFeature
         public void SetLogger(ILogger logger)
         {
             this._eventExecutor.SetLogger(logger);
+            this._repository.SetLogger(logger);
         }
 
         internal void AddClientHandler(string client, ProviderEventTypes eventType, EventHandlerDelegate handler)
@@ -269,8 +270,7 @@ namespace OpenFeature
         /// <summary>
         /// Update the provider state to READY and emit an READY after successful init.
         /// </summary>
-        private void AfterInitialization(FeatureProvider provider)
-
+        private async Task AfterInitialization(FeatureProvider provider)
         {
             provider.Status = ProviderStatus.Ready;
             var eventPayload = new ProviderEventPayload
@@ -280,13 +280,13 @@ namespace OpenFeature
                 ProviderName = provider.GetMetadata().Name,
             };
 
-            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
+            await this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload }).ConfigureAwait(false);
         }
 
         /// <summary>
         /// Update the provider state to ERROR and emit an ERROR after failed init.
         /// </summary>
-        private void AfterError(FeatureProvider provider, Exception ex)
+        private async Task AfterError(FeatureProvider provider, Exception ex)
 
         {
             provider.Status = typeof(ProviderFatalException) == ex.GetType() ? ProviderStatus.Fatal : ProviderStatus.Error;
@@ -297,7 +297,7 @@ namespace OpenFeature
                 ProviderName = provider.GetMetadata()?.Name,
             };
 
-            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
+            await this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload }).ConfigureAwait(false);
         }
     }
 }
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -269,7 +269,6 @@ namespace OpenFeature
 
         /// <summary>
         /// Update the provider state to READY and emit a READY event after successful init.
-
         /// </summary>
         private async Task AfterInitialization(FeatureProvider provider)
         {
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -269,7 +269,8 @@ namespace OpenFeature
         /// <summary>
         /// Update the provider state to READY and emit an READY after successful init.
         /// </summary>
-        private void afterInitialization(FeatureProvider provider)
+        private void AfterInitialization(FeatureProvider provider)
+
         {
             provider.Status = ProviderStatus.Ready;
             var eventPayload = new ProviderEventPayload
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -286,7 +286,8 @@ namespace OpenFeature
         /// <summary>
         /// Update the provider state to ERROR and emit an ERROR after failed init.
         /// </summary>
-        private void afterError(FeatureProvider provider, Exception ex)
+        private void AfterError(FeatureProvider provider, Exception ex)
+
         {
             provider.Status = typeof(ProviderFatalException) == ex.GetType() ? ProviderStatus.Fatal : ProviderStatus.Error;
             var eventPayload = new ProviderEventPayload
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using OpenFeature.Constant;
+using OpenFeature.Error;
 using OpenFeature.Model;
 
 namespace OpenFeature
@@ -37,7 +38,7 @@ namespace OpenFeature
         private Api() { }
 
         /// <summary>
-        /// Sets the feature provider. In order to wait for the provider to be set, and initialization to complete,
+        /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
         /// await the returned task.
         /// </summary>
         /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
@@ -45,9 +46,8 @@ namespace OpenFeature
         public async Task SetProviderAsync(FeatureProvider featureProvider)
         {
             this._eventExecutor.RegisterDefaultFeatureProvider(featureProvider);
-            await this._repository.SetProviderAsync(featureProvider, this.GetContext()).ConfigureAwait(false);
+            await this._repository.SetProviderAsync(featureProvider, this.GetContext(), afterInitialization, afterError).ConfigureAwait(false);
         }
-
 
         /// <summary>
         /// Sets the feature provider to given clientName. In order to wait for the provider to be set, and
@@ -62,7 +62,7 @@ namespace OpenFeature
                 throw new ArgumentNullException(nameof(clientName));
             }
             this._eventExecutor.RegisterClientFeatureProvider(clientName, featureProvider);
-            await this._repository.SetProviderAsync(clientName, featureProvider, this.GetContext()).ConfigureAwait(false);
+            await this._repository.SetProviderAsync(clientName, featureProvider, this.GetContext(), afterInitialization, afterError).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -121,7 +121,7 @@ namespace OpenFeature
         /// <returns><see cref="FeatureClient"/></returns>
         public FeatureClient GetClient(string? name = null, string? version = null, ILogger? logger = null,
             EvaluationContext? context = null) =>
-            new FeatureClient(name, version, logger, context);
+            new FeatureClient(() => _repository.GetProvider(name), name, version, logger, context);
 
         /// <summary>
         /// Appends list of hooks to global hooks list
@@ -265,5 +265,37 @@ namespace OpenFeature
 
         internal void RemoveClientHandler(string client, ProviderEventTypes eventType, EventHandlerDelegate handler)
             => this._eventExecutor.RemoveClientHandler(client, eventType, handler);
+
+        /// <summary>
+        /// Update the provider state to READY and emit an READY after successful init.
+        /// </summary>
+        private void afterInitialization(FeatureProvider provider)
+        {
+            provider.Status = ProviderStatus.Ready;
+            var eventPayload = new ProviderEventPayload
+            {
+                Type = ProviderEventTypes.ProviderReady,
+                Message = "Provider initialization complete",
+                ProviderName = provider.GetMetadata().Name,
+            };
+
+            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
+        }
+
+        /// <summary>
+        /// Update the provider state to ERROR and emit an ERROR after failed init.
+        /// </summary>
+        private void afterError(FeatureProvider provider, Exception ex)
+        {
+            provider.Status = typeof(ProviderFatalException) == ex.GetType() ? ProviderStatus.Fatal : ProviderStatus.Error;
+            var eventPayload = new ProviderEventPayload
+            {
+                Type = ProviderEventTypes.ProviderError,
+                Message = $"Provider initialization error: {ex?.Message}",
+                ProviderName = provider.GetMetadata()?.Name,
+            };
+
+            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
+        }
     }
 }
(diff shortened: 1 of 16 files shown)


src/OpenFeature/Api.cs
```diff
@@ -36,18 +37,6 @@ public sealed class Api : IEventBus
         static Api() { }
         private Api() { }
 
-        /// <summary>
```
toddbaert, line 39: Removed since we already have pending breaking changes.

src/OpenFeature/EventExecutor.cs
```diff
@@ -232,6 +232,7 @@ private async void ProcessFeatureProviderEventsAsync(object? providerRef)
                 switch (item)
                 {
                     case ProviderEventPayload eventPayload:
+                        this.UpdateProviderStatus(typedProviderRef, eventPayload);
```
toddbaert, line 235: This is the only thing I don't like about this implementation - the EventExecutor probably should be altering the provider state. We could instead pass a delegate which would be defined in the API, which might be better separation of concerns.

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -68,16 +72,18 @@ public void SetContext(EvaluationContext? context)
...
         /// Initializes a new instance of the <see cref="FeatureClient"/> class.
         /// </summary>
+        /// <param name="providerAccessor">Function to retrieve current provider</param>
         /// <param name="name">Name of client <see cref="ClientMetadata"/></param>
         /// <param name="version">Version of client <see cref="ClientMetadata"/></param>
         /// <param name="logger">Logger used by client</param>
         /// <param name="context">Context given to this client</param>
         /// <exception cref="ArgumentNullException">Throws if any of the required parameters are null</exception>
-        public FeatureClient(string? name, string? version, ILogger? logger = null, EvaluationContext? context = null)
+        internal FeatureClient(Func<FeatureProvider> providerAccessor, string? name, string? version, ILogger? logger = null, EvaluationContext? context = null)
```
toddbaert, line 81: This constructor should not be public, `OpenFeature.GetClient()` should be used.

src/OpenFeature/ProviderRepository.cs
```diff
@@ -44,31 +44,18 @@ public async ValueTask DisposeAsync()
...
         /// initialization
         /// </param>
-        /// <param name="afterShutdown">called after a provider is shutdown, can be used to remove event handlers</param>
         public async Task SetProvider(
             FeatureProvider? featureProvider,
             EvaluationContext context,
-            Action<FeatureProvider>? afterSet = null,
-            Action<FeatureProvider>? afterInitialization = null,
-            Action<FeatureProvider, Exception>? afterError = null,
-            Action<FeatureProvider>? afterShutdown = null)
```
toddbaert, line 71: `afterSet` and `afterShutdown` were only used in tests. Removing them for simplicity.

src/OpenFeature/FeatureProvider.cs
```diff
@@ -88,34 +90,24 @@ public abstract class FeatureProvider
...
-        /// <returns>The current <see cref="ProviderStatus"/></returns>
-        /// <remarks>
-        /// If a provider does not override this method, then its status will be assumed to be
-        /// <see cref="ProviderStatus.Ready"/>. If a provider implements this method, and supports initialization,
-        /// then it should start in the <see cref="ProviderStatus.NotReady"/>status . If the status is
-        /// <see cref="ProviderStatus.NotReady"/>, then the Api will call the <see cref="Initialize" /> when the
-        /// provider is set.
-        /// </remarks>
-        public virtual ProviderStatus GetStatus() => ProviderStatus.Ready;
+        internal virtual ProviderStatus Status { get; set; } = ProviderStatus.NotReady;
```
toddbaert, line 97: This is now assembly-internal.

src/OpenFeature/FeatureProvider.cs
```diff
@@ -1,9 +1,11 @@
 using System.Collections.Immutable;
+using System.Runtime.CompilerServices;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using OpenFeature.Constant;
 using OpenFeature.Model;
 
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] // required to allow NSubstitute mocking of internal methods
```
toddbaert, line 8: This allows us to mock the internal `ProviderStatus` in tests

kinyoklion:
Maybe more related to the spec than to this specific implementation. But I did find it somewhat hard to handle the runtime status after initialization status when the SDK automatically implements ready. 

I think it makes it easy for simple providers, and somewhat cumbersome for more complex providers. I would actually prefer to be able to opt out. 

Basically an underlying SDK may already emit these events, but we have to connect and translate these events conditionally around initialization. So after initialization if a provider fails we want to emit that event, but if during initialization it fails we need to not emit that event from the provider because the SDK also will emit it.

Example of attempting to reconcile this in a python provider: https://github.com/launchdarkly/openfeature-python-server/blob/243a8dc515bcd16fe8c7d5a0913fbdfe6e073aa0/ld_openfeature/provider.py#L46

toddbaert:
> So after initialization if a provider fails we want to emit that event, but if during initialization it fails we need to not emit that event from the provider because the SDK also will emit it.

I can see the issue here. We have some similar logic in flagd since it also supports automatic reconnection. I think for the majority of providers though, such events around reconnection etc, are not supported. It does indeed seem to be something featured in the most rigorous implementations.

I'm open to some kind of `opt-out` for emitting events on READY, but I'm not sure if the ROI is there. :thinking: I suppose it would be fairly easy to implement. :thinking:

kinyoklion:
> > So after initialization if a provider fails we want to emit that event, but if during initialization it fails we need to not emit that event from the provider because the SDK also will emit it.
> 
> I can see the issue here. We have some similar logic in flagd since it also supports automatic reconnection. I think for the majority of providers though, such events around reconnection etc, are not supported; it does indeed seem to be something featured in the most rigorous implementations.
> 
> I'm open to some kind of `opt-out` for emitting events on READY, but I'm not sure if the ROI is there. 🤔 I suppose it would be fairly easy to implement. 🤔

I think an opt-out would work. 

I've not thought about it extensively, but another option may be to have automatic events be part of a base class that more complex providers could choose not to use instead of it being in the API. Though that path isn't the most sustainable unless you have a language that supports more "mixin" type classes. So maybe not.

Maybe also we can leave it as is and think if the situation can be improved. Overall I feel like getting initialization correct has become more complex, so maybe the abstraction needs adjusted some.

toddbaert:
> > > So after initialization if a provider fails we want to emit that event, but if during initialization it fails we need to not emit that event from the provider because the SDK also will emit it.
> > 
> > 
> > I can see the issue here. We have some similar logic in flagd since it also supports automatic reconnection. I think for the majority of providers though, such events around reconnection etc, are not supported; it does indeed seem to be something featured in the most rigorous implementations.
> > I'm open to some kind of `opt-out` for emitting events on READY, but I'm not sure if the ROI is there. 🤔 I suppose it would be fairly easy to implement. 🤔
> 
> I think an opt-out would work.
> 
> I've not thought about it extensively, but another option may be to have automatic events be part of a base class that more complex providers could choose not to use instead of it being in the API. Though that path isn't the most sustainable unless you have a language that supports more "mixin" type classes. So maybe not.
> 
> Maybe also we can leave it as is and think if the situation can be improved. Overall I feel like getting initialization correct has become more complex, so maybe the abstraction needs adjusted some.

I thought a bit about an opt-out. I think we could add some fields/metadata to the provider to signal this and possibly other behaviors. Perhaps we could open an new issue to discuss this in the spec.

src/OpenFeature/EventExecutor.cs
Kavindu-Dodan: Can't we keep provider status in the `EventExecutor` and avoid any accessor from the provider itself ? `EventExecutor` already maintains a provider dictionary [1]

[1] - https://github.com/open-feature/dotnet-sdk/blob/626ee1046bc9b1c0bfb3bbbce0fee2b7d7d5b45a/src/OpenFeature/EventExecutor.cs#L18

src/OpenFeature/FeatureProvider.cs
Kavindu-Dodan: related - https://github.com/open-feature/dotnet-sdk/pull/276/files#r1643102758

toddbaert:
I had to rebase this and resolve conflicts on the recently merged concellationToken/Async chnage.

src/OpenFeature/EventExecutor.cs
toddbaert: Perhaps, but the client needs to expose that field, (`client.ProviderStatus`) so we'd still need to do some awkward plumbing. Plus, I think the event executor should really just be about executing events, not changing this status, I think.

Kavindu-Dodan: That's true. For dot net implementation, we have virtual method to hide the status field. But for other langs, I think it'll be tricky.

Kavindu-Dodan: Anyway the other workaround is to introduce a state maintainer (similar to event handler) at SDK level

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -246,6 +252,16 @@ public void AddHooks(IEnumerable<Hook> hooks)
             {
                 var contextFromHooks = await this.TriggerBeforeHooksAsync(allHooks, hookContext, options, cancellationToken).ConfigureAwait(false);
 
+                // short circuit evaluation entirely if provider is in a bad state
```
kinyoklion, line 255: This is probably another comment outside the scope of this PR, but similar to the provider events from init this creates some potentially interesting problems. 

An SDK can fail to initialize, but be capable of handling evaluations. (The LD SDK, for instance, will always evaluate the flag and does other work associated with that evaluation even if it returns a default value.)

When the provider is bypassed like this it means that the analytic events associated with those evaluations are also bypassed. (Because a failure to be able to retrieve configuration is not 100% correlated with not being able to make other connections).

src/OpenFeature/EventExecutor.cs
kinyoklion: I think the provider being the storage for the status, but not really being the owner or maintainer, is going to be awkward regardless where the updates are done.

You have the provider and then you have inferred state of the provider based on its behavior. Maybe this could be extended to other state tracking, but for now probably it is just the provider state.

You also want this state to be tied to the same lifetime as the provider, even if it is externalized.

One approach would be to wrap each provider in a "stateful provider". (Though it still has some awkward complexity as you can register a provider more than once, and you want to share the state.)

src/OpenFeature/Api.cs
```diff
@@ -265,5 +265,37 @@ internal void AddClientHandler(string client, ProviderEventTypes eventType, Even
 
         internal void RemoveClientHandler(string client, ProviderEventTypes eventType, EventHandlerDelegate handler)
             => this._eventExecutor.RemoveClientHandler(client, eventType, handler);
+
+        /// <summary>
+        /// Update the provider state to READY and emit an READY after successful init.
+        /// </summary>
+        private void afterInitialization(FeatureProvider provider)
```
askpt, line 272: Could you please rename it?
```suggestion
        private void AfterInitialization(FeatureProvider provider)

```

src/OpenFeature/Api.cs
```diff
@@ -265,5 +265,37 @@ internal void AddClientHandler(string client, ProviderEventTypes eventType, Even
...
+        {
+            provider.Status = ProviderStatus.Ready;
+            var eventPayload = new ProviderEventPayload
+            {
+                Type = ProviderEventTypes.ProviderReady,
+                Message = "Provider initialization complete",
+                ProviderName = provider.GetMetadata().Name,
+            };
+
+            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
```
askpt, line 282: I'm not a fan of this line, to be honest. We have an async method here that has yet to be awaited. The `WriteAsync` returns a `ValueTask`, which can lead to performance degradation in this case. Check: https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca2012

src/OpenFeature/Api.cs
```diff
@@ -265,5 +265,37 @@ internal void AddClientHandler(string client, ProviderEventTypes eventType, Even
...
+        {
+            provider.Status = typeof(ProviderFatalException) == ex.GetType() ? ProviderStatus.Fatal : ProviderStatus.Error;
+            var eventPayload = new ProviderEventPayload
+            {
+                Type = ProviderEventTypes.ProviderError,
+                Message = $"Provider initialization error: {ex?.Message}",
+                ProviderName = provider.GetMetadata()?.Name,
+            };
+
+            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
```
askpt, line 298: Same as the previous comment about `WriteAsync`.

src/OpenFeature/Api.cs
```diff
@@ -265,5 +265,37 @@ internal void AddClientHandler(string client, ProviderEventTypes eventType, Even
...
+                ProviderName = provider.GetMetadata().Name,
+            };
+
+            this._eventExecutor.EventChannel.Writer.WriteAsync(new Event { Provider = provider, EventPayload = eventPayload });
+        }
+
+        /// <summary>
+        /// Update the provider state to ERROR and emit an ERROR after failed init.
+        /// </summary>
+        private void afterError(FeatureProvider provider, Exception ex)
```
askpt, line 288: Please rename it.
```suggestion
        private void AfterError(FeatureProvider provider, Exception ex)

```

src/OpenFeature/Model/ProviderEvents.cs
```diff
@@ -28,6 +28,11 @@ public class ProviderEventPayload
         /// </summary>
         public string? Message { get; set; }
 
+        /// <summary>
+        /// Optional error associated with the event. 
+        /// </summary>
+        public ErrorType? errorType { get; set; }
```
askpt, line 34: Please rename it. 
```suggestion
        public ErrorType? ErrorType { get; set; }
```

src/OpenFeature/ProviderRepository.cs
```diff
@@ -242,11 +206,9 @@ await InitProviderAsync(this._defaultProvider, context, afterInitialization, aft
             try
             {
                 await targetProvider.ShutdownAsync().ConfigureAwait(false);
-                afterShutdown?.Invoke(targetProvider);
             }
-            catch (Exception ex)
+            catch (Exception)
```
askpt, line 210: Is there any reason why we decided to remove the exception here?

askpt (commented):
I added a few comments and concerns about the `ValueTask`.

src/OpenFeature/OpenFeatureClient.cs
toddbaert: I agree fully, this is one of the reasons we don't bypass the `ERROR` state though. We only bypass evaluations if the provider has not been initialized (`NOT_READY`) or if it's in `FATAL` state.

In both these situations, I don't think analytics events could even be reliably transmitted anyway, since we likely haven't connected or there's a credential error, or something the like.

src/OpenFeature/Api.cs
toddbaert: I'm not sure... 

From your article:

> Attempts to consume a ValueTask multiple times or to directly access one's result before it's known to be completed may result in an exception or corruption.

This makes sense, and I'm aware of the pitfalls here.

> Ignoring such a ValueTask is likely an indication of a functional bug and may degrade performance.

I'm not sure if this is saying that not awaiting a task at all (using it in a "fire and forget" way) is a problem. I can't see how it would be, but I could be missing something... I can use a combination of `WaitToWriteAsync` and `TryWrite`, but I don't think the end result is much different.

src/OpenFeature/Api.cs
toddbaert: Ugh sorry, my Java is showing.

src/OpenFeature/EventExecutor.cs
toddbaert: > One approach would be to wrap each provider in a "stateful provider". (Though it still has some awkward complexity as you can register a provider more than once, and you want to share the state.)

This is exactly what I did in JS (where there's no `internal` equivalent, and it felt like a lot of trouble. I think this is a valid usage for an internal prop, but if there's a consensus I can make changes. It just felt like a poor ROI in here for this single simple field.

src/OpenFeature/ProviderRepository.cs
toddbaert: It's now unused. We before passed an optional lambda that wasn't ever being set, except in tests.

I've added some logging here instead.

src/OpenFeature/OpenFeatureClient.cs
kinyoklion: This would also apply to basic non-blocking cases though, correct? Set the provider without waiting and start evaluating flags? Typically I would expect a vendor to cache events and then send them once the provider is ready.

kinyoklion: Regardless though. It is fine for this PR.

src/OpenFeature/Api.cs
```diff
@@ -258,12 +258,46 @@ public void RemoveHandler(ProviderEventTypes type, EventHandlerDelegate handler)
...
         }
 
         internal void AddClientHandler(string client, ProviderEventTypes eventType, EventHandlerDelegate handler)
             => this._eventExecutor.AddClientHandler(client, eventType, handler);
 
         internal void RemoveClientHandler(string client, ProviderEventTypes eventType, EventHandlerDelegate handler)
             => this._eventExecutor.RemoveClientHandler(client, eventType, handler);
+
+        /// <summary>
+        /// Update the provider state to READY and emit an READY after successful init.
```
kinyoklion, line 271:
```suggestion
        /// Update the provider state to READY and emit a READY event after successful init.

```
nit

src/OpenFeature/Api.cs
askpt: In this case, the `ValueTask` is not used again, so the code will not directly cause any issues. However, I am more worried if we need to reuse the value in the future. Thanks for the change 😄

askpt (approved):
LGTM. I don't have any more concerns about this PR. Thanks @toddbaert

kinyoklion (approved):

benjiro (approved):

Issue #235, opened 2024-02-14
[FEATURE] Expose TargetingKey with all value exporters 


Kavindu-Dodan:
### Requirements

Similar to the Java SDK [1], Dotnet SDK `EvaluationContext` should include targetingKey with bulk exporters. 

Currently `EvaluationContext ` exposes `AsDictionary()` [2] as well as `GetEnumerator()`[3], but they do not include the `targetingKey`, which is stored in a dedicated field. 

**AC**

Include the targetingKey in bulk exporters and refactor of the property to match with this change

[1] - https://github.com/open-feature/java-sdk/issues/801
[2] - https://github.com/open-feature/dotnet-sdk/blob/v1.4.1/src/OpenFeature/Model/EvaluationContext.cs#L76
[3] - https://github.com/open-feature/dotnet-sdk/blob/v1.4.1/src/OpenFeature/Model/EvaluationContext.cs#L90

kinyoklion:
The addition of the `TargetingKey` property has presented some problems for the LaunchDarkly provider. We had supported using attributes with the name "key" (convenient coming from an LD context) as well as "targetingKey". The `TargetingKey` not being in the map just meant we basically were unaware of it.

I wonder if we should have just made it short-hand for "targetingKey" as a member of the underlying structure. If we just export it, then we have to use some kind of name that will shadow whatever we pick. (Likely "targetingKey").

Instead we could adjust the general `Set` method to have restrictions on "targetingKey". We do something similar in LD contexts where if you call the set function with a built-in attribute we do our best, and then log if we cannot handle it. 

Similar situation: https://github.com/launchdarkly/dotnet-sdk-common/blob/d248d74b29fb0d1e27771c5810b8277fd256e266/src/LaunchDarkly.CommonSdk/ContextBuilder.cs#L272

toddbaert:
@kinyoklion I think I understand what you're suggesting, but I want to be sure... Are you suggesting that you think the targeting key setter/getter should just retrieve a "special" property from the map (probably `targetingKey`?

I think that's fine, if so. I think the real problem is just that people will expect the `AsDictionary()` and `GetEnumerator()` to convey a targeting key, _somehow_ if it's set.

kinyoklion:
@toddbaert Yeah, I am suggesting that the key of "targetingKey" and the setter/getter for targetingKey should operate on the same value. 

Which either requires a specially interacted with field or a special entry in the map with specialized logic for the set/get generic mechanisms.

[...]

PR #287, opened 2024-07-26
feat: back targetingKey with internal map

toddbaert:
Use the internal dictionary for the `targetingKey`.

This is non-breaking from a compiler perspective. It could result in some behavioral changes, but IMO they are largely desirable.

Fixes: https://github.com/open-feature/dotnet-sdk/issues/235

### DIFF
--- a/src/OpenFeature/Model/EvaluationContext.cs
+++ b/src/OpenFeature/Model/EvaluationContext.cs
@@ -14,7 +14,8 @@ namespace OpenFeature.Model
         /// <summary>
         /// The index for the "targeting key" property when the EvaluationContext is serialized or expressed as a dictionary.
         /// </summary>
-        public readonly static string TargetingKeyIndex = "targetingKey";
+        internal const string TargetingKeyIndex = "targetingKey";
+
 
         private readonly Structure _structure;
 
--- a/src/OpenFeature/Model/EvaluationContext.cs
+++ b/src/OpenFeature/Model/EvaluationContext.cs
@@ -11,18 +11,18 @@ namespace OpenFeature.Model
     /// <seealso href="https://github.com/open-feature/spec/blob/v0.5.2/specification/sections/03-evaluation-context.md">Evaluation context</seealso>
     public sealed class EvaluationContext
     {
+        /// <summary>
+        /// The index for the "targeting key" property when the EvaluationContext is serialized or expressed as a dictionary.
+        /// </summary>
+        public readonly static string TargetingKeyIndex = "targetingKey";
+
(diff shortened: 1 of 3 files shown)


src/OpenFeature/Model/EvaluationContext.cs
```diff
@@ -11,26 +11,29 @@ namespace OpenFeature.Model
...
+
         private readonly Structure _structure;
 
         /// <summary>
         /// Internal constructor used by the builder.
         /// </summary>
-        /// <param name="targetingKey">The targeting key</param>
-        /// <param name="content">The content of the context.</param>
-        internal EvaluationContext(string? targetingKey, Structure content)
+        /// <param name="content"></param>
```
toddbaert, line 24: This is only used internally, so not a breaking change.

toddbaert:
@Kavindu-Dodan adding you since you opened the issue and implemented it in Java.

src/OpenFeature/Model/EvaluationContext.cs
```diff
@@ -11,26 +11,29 @@ namespace OpenFeature.Model
     /// <seealso href="https://github.com/open-feature/spec/blob/v0.5.2/specification/sections/03-evaluation-context.md">Evaluation context</seealso>
     public sealed class EvaluationContext
     {
+        /// <summary>
+        /// The index for the "targeting key" property when the EvaluationContext is serialized or expressed as a dictionary.
+        /// </summary>
+        public readonly static string TargetingKeyIndex = "targetingKey";
```
askpt, line 17:
```suggestion
        public const string TargetingKeyIndex = "targetingKey";

```

Can we make it `const` here for some performance gains? Do we want this to be exposed, or can it be `internal`?

askpt (commented):
Minor comment around visibility and `const`. The rest seems fine

src/OpenFeature/Model/EvaluationContext.cs
benjiro: Yea i don't think we need to expose this publicly

toddbaert: I was going back and forth on this, so I am fine to not expose it.

askpt: Yup. Let's make `internal` :)

toddbaert: Done

askpt (approved):
LGTM! 👍

beeme1mr (approved):

PR #294, opened 2024-08-12
feat!: domain instead of client name

toddbaert:
Uses "domain" terminology instead of "client name / named client".

Fixes: https://github.com/open-feature/dotnet-sdk/issues/249

I believe with this, we are able to release a 2.0

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -50,19 +50,19 @@ namespace OpenFeature
         }
 
         /// <summary>
-        /// Sets the feature provider to given clientName. In order to wait for the provider to be set, and
+        /// Binds the feature provider to the given domain. In order to wait for the provider to be set, and
         /// initialization to complete, await the returned task.
         /// </summary>
-        /// <param name="clientName">Name of client</param>
+        /// <param name="domain">An identifier which logically binds clients with providers</param>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
-        public async Task SetProviderAsync(string clientName, FeatureProvider featureProvider)
+        public async Task SetProviderAsync(string domain, FeatureProvider featureProvider)
         {
-            if (string.IsNullOrWhiteSpace(clientName))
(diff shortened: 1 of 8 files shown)


src/OpenFeature/Api.cs
```diff
@@ -82,14 +82,14 @@ public FeatureProvider GetProvider()
         }
 
         /// <summary>
-        /// Gets the feature provider with given clientName
+        /// Gets the feature provider with given domain
         /// </summary>
-        /// <param name="clientName">Name of client</param>
-        /// <returns>A provider associated with the given clientName, if clientName is empty or doesn't
+        /// <param name="domain">Name of client</param>
```
askpt, line 87: We are renaming from `clientName` to `domain`. Does it make sense to change the description accordingly?
```suggestion
        /// <param name="domain">An identifier which logically binds clients with providers</param>

```

askpt (commented):
I added some minor comments around the comments for consistency. Other than that, it looks good to me.

src/OpenFeature/Api.cs
```diff
@@ -104,12 +104,12 @@ public FeatureProvider GetProvider(string clientName)
         public Metadata? GetProviderMetadata() => this.GetProvider().GetMetadata();
 
         /// <summary>
-        /// Gets providers metadata assigned to the given clientName. If the clientName has no provider
+        /// Gets providers metadata assigned to the given domain. If the domain has no provider
         /// assigned to it the default provider will be returned
         /// </summary>
-        /// <param name="clientName">Name of client</param>
+        /// <param name="domain">Name of client</param>
```
askpt, line 110: Same as https://github.com/open-feature/dotnet-sdk/pull/294/files#r1715020674

```suggestion
        /// <param name="domain">An identifier which logically binds clients with providers</param>

```

src/OpenFeature/Api.cs
toddbaert: Totally. I missed this one.

askpt (approved):

kinyoklion (approved):

PR #312, opened 2024-11-11
feat: Implement transaction context

### LINKED ISSUE #243: [FEATURE] Implement transaction context

askpt:
## This PR

- adds this new feature

### Related Issues

Fixes #243

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -22,6 +22,8 @@ namespace OpenFeature
         private EventExecutor _eventExecutor = new EventExecutor();
         private ProviderRepository _repository = new ProviderRepository();
         private readonly ConcurrentStack<Hook> _hooks = new ConcurrentStack<Hook>();
+        private ITransactionContextPropagator? _transactionContextPropagator;
+        private object _transactionContextPropagatorLock = new();
 
         /// The reader/writer locks are not disposed because the singleton instance should never be disposed.
         private readonly ReaderWriterLockSlim _evaluationContextLock = new ReaderWriterLockSlim();
@@ -47,6 +49,7 @@ namespace OpenFeature
         {
             this._eventExecutor.RegisterDefaultFeatureProvider(featureProvider);
             await this._repository.SetProviderAsync(featureProvider, this.GetContext(), this.AfterInitialization, this.AfterError).ConfigureAwait(false);
+
         }
 
         /// <summary>
@@ -85,7 +88,6 @@ namespace OpenFeature
         /// Gets the feature provider with given domain
         /// </summary>
         /// <param name="domain">An identifier which logically binds clients with providers</param>
-
         /// <returns>A provider associated with the given domain, if domain is empty or doesn't
         /// have a corresponding provider the default provider will be returned</returns>
         public FeatureProvider GetProvider(string domain)
@@ -109,7 +111,6 @@ namespace OpenFeature
         /// assigned to it the default provider will be returned
         /// </summary>
         /// <param name="domain">An identifier which logically binds clients with providers</param>
-
         /// <returns>Metadata assigned to provider</returns>
         public Metadata? GetProviderMetadata(string domain) => this.GetProvider(domain).GetMetadata();
 
@@ -216,6 +217,64 @@ namespace OpenFeature
             {
                 this._evaluationContextLock.ExitReadLock();
             }
+        }
+
+        /// <summary>
+        /// Return the transaction context propagator.
+        /// </summary>
+        /// <returns><see cref="ITransactionContextPropagator"/>the registered transaction context propagator</returns>
+        public ITransactionContextPropagator? GetTransactionContextPropagator()
+        {
+            return this._transactionContextPropagator;
+        }
+
+        /// <summary>
+        /// Sets the transaction context propagator.
+        /// </summary>
+        /// <param name="transactionContextPropagator">the transaction context propagator to be registered</param>
+        /// <exception cref="ArgumentNullException">Transaction context propagator cannot be null</exception>
+        public void SetTransactionContextPropagator(ITransactionContextPropagator transactionContextPropagator)
+        {
+            if (transactionContextPropagator == null)
+            {
+                throw new ArgumentNullException(nameof(transactionContextPropagator),
+                    "Transaction context propagator cannot be null");
+            }
+
+            lock (this._transactionContextPropagatorLock)
+            {
+                this._transactionContextPropagator = transactionContextPropagator;
+            }
+        }
+
+        /// <summary>
+        /// Returns the currently defined transaction context using the registered transaction context propagator.
+        /// </summary>
+        /// <returns><see cref="EvaluationContext"/>The current transaction context</returns>
+        public EvaluationContext? GetTransactionContext()
+        {
+            return this._transactionContextPropagator?.GetTransactionContext();
+        }
+
+        /// <summary>
+        /// Sets the transaction context using the registered transaction context propagator.
+        /// </summary>
+        /// <param name="evaluationContext">The <see cref="EvaluationContext"/> to set</param>
+        /// <exception cref="InvalidOperationException">Transaction context propagator is not set.</exception>
+        /// <exception cref="ArgumentNullException">Evaluation context cannot be null</exception>
+        public void SetTransactionContext(EvaluationContext evaluationContext)
+        {
+            if (evaluationContext == null)
+            {
+                throw new ArgumentNullException(nameof(evaluationContext), "Evaluation context cannot be null");
+            }
+
+            if (this._transactionContextPropagator == null)
+            {
+                throw new InvalidOperationException("Transaction context propagator is not set");
+            }
+
+            this._transactionContextPropagator?.SetTransactionContext(evaluationContext);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


askpt:
> Nice start! Left some thoughts and references to how we are doing it in the other languages. 

Thanks for the review @lukas-reining. I based the implementation on the JS client. I'll look into more detail on the Java version instead.

src/OpenFeature/Api.cs
```diff
@@ -218,6 +219,59 @@ public EvaluationContext GetContext()
             }
         }
 
+        /// <summary>
+        /// Return the transaction context propagator.
+        /// </summary>
+        /// <returns><see cref="ITransactionContextPropagator"/>the registered transaction context propagator</returns>
+        public ITransactionContextPropagator GetTransactionContextPropagator()
```
toddbaert, line 226: I don't think we need this accessor, if we need it for tests that's fine, but let's make it internal in that case.

toddbaert:
This looks right to me, though unless there's a good reason, I think we should remove [this](https://github.com/open-feature/dotnet-sdk/pull/312#discussion_r1900109544).

I also [pushed a small change](https://github.com/open-feature/dotnet-sdk/pull/312/commits/f461152cc1ea78b55fb7d8b6b393bc5b8fb842aa) to add docs to the README. @askpt please check them and make sure they make sense to you (it's a similar example to other SDKs).

src/OpenFeature/Api.cs
askpt: I followed the [java-sdk](https://github.com/open-feature/java-sdk/blob/f1817d8fef585f957de1cfb9222b03cb591ed2e9/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java#L139) implementation. I am happy to change it to `internal` if it is the expected behaviour.

Let me know if I should remove or mark it as internal.

toddbaert: Ah, I didn't realize we implemented it there. I'm not of the opinion it's a great idea, but since it's also there it's not a blocker for me (it's not in the spec).

I can go either way... do you think we should have it? 

Approving now.

toddbaert (approved):
Approving, but highlighting this thread: https://github.com/open-feature/dotnet-sdk/pull/312#discussion_r1900109544

src/OpenFeature/Api.cs
askpt: Does anyone outside of the client **need** to access the Transaction Context Propagator? Likely not... 

I prefer to keep it `internal` so we expose what is necessary. But I don't like different language implementations going in different directions, and I don't know what the best approach is.

toddbaert: > I prefer to keep it internal so we expose what is necessary.

I agree.

> But I don't like different language implementations going in different directions, and I don't know what the best approach is.

Of course I generally agree, that's why we have a spec, but there's always going to be some deviations and extras. If anything, we should probably deprecate the Java function and remove it eventually. I don't see it in JS either.

askpt: Change visibility here: https://github.com/open-feature/dotnet-sdk/pull/312/commits/bf4dbb4ff1d0a399aeb9e4d424be755d9dc05920

src/OpenFeature/AsyncLocalTransactionContextPropagator.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading;
+using OpenFeature.Model;
+
+namespace OpenFeature;
+
+/// <summary>
+///  It is a no-op implementation of <see cref="ITransactionContextPropagator"/>
```
chrfwow, line 7: I would not consider this a No-Op implementation. Maybe this line should be moved to the `NoOpTransactionContextPropagator`?

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -221,6 +221,7 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
             evaluationContextBuilder.Merge(evaluationContext);
             evaluationContextBuilder.Merge(this.GetContext());
             evaluationContextBuilder.Merge(context);
+            evaluationContextBuilder.Merge(Api.Instance.GetTransactionContext());
```
chrfwow, line 224: According to Requirement 3.2.3, the order should be `API (global; lowest precedence) - transaction - client - invocation - before hooks (highest precedence)`, so I believe `evaluationContextBuilder.Merge(context);` should be the last merge call.
I think we should alos have a test to verify the correct merge order.

test/OpenFeature.Tests/AsyncLocalTransactionContextPropagatorTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        // Assert
+        Assert.Equal(EvaluationContext.Empty, context);
+    }
+
+    [Fact]
+    public void SetTransactionContext_SetsAndGetsContextCorrectly()
+    {
+        // Arrange
+        var propagator = new AsyncLocalTransactionContextPropagator();
+        var evaluationContext = EvaluationContext.Empty;
```
chrfwow, line 26: Maybe we should use a different value here, so that we can be sure that we actually get the value we stored in the propagator, and not just the default empty value

test/OpenFeature.Tests/OpenFeatureTests.cs
```diff
@@ -244,5 +245,74 @@ public async Task OpenFeature_Should_Allow_Multiple_Client_Mapping()
...
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => api.SetTransactionContext(null!));
+        }
+
+        [Fact]
+        public void SetTransactionContext_ShouldSetTransactionContext_WhenValidEvaluationContextIsProvided()
+        {
+            // Arrange
+            var api = Api.Instance;
+            var evaluationContext = EvaluationContext.Empty;
```
chrfwow, line 288: Same as previous comment

README.md
```diff
@@ -234,6 +235,28 @@ The OpenFeature API provides a close function to perform a cleanup of all regist
...
+### Transaction Context Propagation
+
+Transaction context is a container for transaction-specific evaluation context (e.g. user id, user agent, IP).
+Transaction context can be set where specific data is available (e.g. an auth service or request handler) and by using the transaction context propagator it will automatically be applied to all flag evaluations within a transaction (e.g. a request or thread).
+By default, the `NoOpTransactionContextPropagator` is used, which doesn't store anything.
+To register a [AsyncLocal](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1) context propagator, you can use the `SetTransactionContextPropagator` method as shown below.
+
+```csharp
+// registering the AsyncLocalTransactionContextPropagator
+Api.Instance.SetTransactionContextPropagator(new AsyncLocalTransactionContextPropagator());
```
chrfwow, line 247: Out of curiosity, why don't we set a AsyncLocalTransactionContextPropagator as default value? Then the user would not have to set one manually first, which would remove one potential error source.

src/OpenFeature/AsyncLocalTransactionContextPropagator.cs
askpt: Thanks! Fixed with: https://github.com/open-feature/dotnet-sdk/pull/312/commits/7a6099712fb6d7a5b96866330973292839058e57

test/OpenFeature.Tests/OpenFeatureTests.cs
askpt: Changed: https://github.com/open-feature/dotnet-sdk/pull/312/commits/1769be823c4a57d1e58b0241c4c5f02f226567fb

test/OpenFeature.Tests/AsyncLocalTransactionContextPropagatorTests.cs
askpt: Changed: https://github.com/open-feature/dotnet-sdk/pull/312/commits/1769be823c4a57d1e58b0241c4c5f02f226567fb

src/OpenFeature/OpenFeatureClient.cs
askpt: Good catch! Updated: https://github.com/open-feature/dotnet-sdk/pull/312/commits/59ddca73db5da8b4522ac62be3231f30ae704ba6

@toddbaert please review this part again 👍

README.md
askpt: I followed the Java SDK implementation on this: https://github.com/open-feature/java-sdk/blob/f1817d8fef585f957de1cfb9222b03cb591ed2e9/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java#L31

chrfwow (approved):
Looks good

PR #317, opened 2024-11-15
feat: add dotnet 9 support, rm dotnet 6

askpt:
## This PR

- Updates all dotnet packages
- Updates the Target Framework
- Updates all GH actions
- Removes net6.0
- Updates AutoFixture to a beta version

### Notes

- .net6 is now EOF by Microsoft, so we should remove it. See https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core for reference
- I needed to use a beta version of AutoFixture since, with .net9, it is more strict in package vulnerabilities, and a library version used by AutoFixture has vulnerabilities. See https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.12 for extra Nuget information and https://github.com/AutoFixture/AutoFixture/issues/1481 for AutoFixture


toddbaert:
LGTM, ready to approve a non-draft version.

src/OpenFeature/OpenFeature.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0;net8.0;net462</TargetFrameworks>
```
toddbaert, line 4: @askpt sorry, I didn't notice this before... this is removing `netstandard2.0`, which IIUC means we won't have support for some older platforms, specifically I'm thinking about support for full .NET framework 4.x.x. Am I correct on this? If so, is that really what we want?

askpt: Hey @toddbaert! I blame the tools! 

The `netstandard2.0` TFM is still there. It just moved to 3rd instead of 1st.

toddbaert: :facepalm: Ha! I'm sorry. Disregard then. I'm merging this.

PR #323, opened 2024-11-25
feat: Support Returning Error Resolutions from Providers

### LINKED ISSUE #307: [FEATURE] Support Returning Error Resolutions from Providers
### Requirements

As of right now, the OpenFeature .NET SDK doesn't properly handle returning error resolutions; when a provider returns an erroneous evaluation by setting the errotType, the `error hooks` do not run, and for many purposes, the evaluation appears to have succeeded.

In other SDKs, we've made changes to check the resolution details and if the errorType is set (or anything but `ErrorType.None` in the case of this SDK) we then behave similarly to if something was thrown. [We don't currently do this in the .NET SDK](https://github.com/open-feature/dotnet-sdk/blob/c471c062cf70d78b67f597f468c62dbfbf0674d2/src/OpenFeature/OpenFeatureClient.cs#L268-L281). This isn't ideal, since exceptions are generally bad for control flow and are expensive.

To resolve this:

- after evaluation, check if the `errorType` is not `ErrorType.None`, and if not, behave as if the provider threw an Exception by:
  - running the error hooks
  - setting the error message (the description of the `ErrorType` would be fine)
  - NOT running the after hooks
- add associated tests

chrfwow:
## This PR
- When a Flag evaluation fails without throwing an exception, the registered error hooks are invoked, and the after hooks are not invoked. 

### Related Issues

Fixes #307 

### How to test
`test/OpenFeature.Tests/OpenFeatureClientTests#When_Error_Is_Returned_From_Provider_Should_Not_Run_After_Hook_But_Error_Hook` verifies the behaviour

### DIFF
--- a/src/OpenFeature/OpenFeatureClient.cs
+++ b/src/OpenFeature/OpenFeatureClient.cs
@@ -263,7 +263,23 @@ namespace OpenFeature
                     (await resolveValueDelegate.Invoke(flagKey, defaultValue, contextFromHooks.EvaluationContext, cancellationToken).ConfigureAwait(false))
                     .ToFlagEvaluationDetails();
 
-                await this.TriggerAfterHooksAsync(allHooksReversed, hookContext, evaluation, options, cancellationToken).ConfigureAwait(false);
+                if (evaluation.ErrorType == ErrorType.None)
+                {
+                    await this.TriggerAfterHooksAsync(
+                        allHooksReversed,
+                        hookContext,
+                        evaluation,
+                        options,
+                        cancellationToken
+                    ).ConfigureAwait(false);
+                }
+                else
+                {
+                    var exception = new FeatureProviderException(evaluation.ErrorType, evaluation.ErrorMessage);
+                    this.FlagEvaluationErrorWithDescription(flagKey, evaluation.ErrorType.GetDescription(), exception);
+                    await this.TriggerErrorHooksAsync(allHooksReversed, hookContext, exception, options, cancellationToken)
+                        .ConfigureAwait(false);
+                }
             }
             catch (FeatureProviderException ex)
             {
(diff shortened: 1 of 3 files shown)


src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -263,7 +263,23 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
...
+                        options,
+                        cancellationToken
+                    ).ConfigureAwait(false);
+                }
+                else
+                {
+                    var exception = new FeatureProviderException(evaluation.ErrorType, evaluation.ErrorMessage);
+                    this.FlagEvaluationErrorWithDescription(flagKey, evaluation.ErrorType.GetDescription(), exception);
+                    await this.TriggerErrorHooksAsync(allHooksReversed, hookContext, exception, options, cancellationToken)
+                        .ConfigureAwait(false);
```
askpt, line 281: Do we not need to assign the `evaluation` variable here? We do it when we get an exception, and at this stage, we don't do it when there is an error at the provider error.

askpt (commented):
Small question. Other than that it seems fine 👍

src/OpenFeature/OpenFeatureClient.cs
chrfwow: I'm not sure I understand your question. The `evaluation` variable is set in line 262, and the exception on this line produces no new information

askpt: I meant that we override the `evaluation` variable with some "custom" values when there is an exception. In this case, we are just returning the result without overriding. Is this the expected behaviour?

If it is, I am happy to approve it. 👍

chrfwow: That's a good question. I assumed that the initial value for evaluation will contain the proper data. Maybe someone else knows what to do here?

aepfli: Yes that is right, the information should be already contained in the evaluation details, as we do it in java and js

aepfli (approved):

askpt (approved):

toddbaert (approved):
Looks great!

PR #338, opened 2024-12-31
feat: Move OTEL hooks to the SDK

### LINKED ISSUE #175: [FEATURE] Promote OTel hooks from contrib to in-the-box
### Requirements

Following open-feature/dotnet-sdk-contrib#132, the `OpenFeature.Contrib.Hooks.Otel` package now only depends on `OpenFeature` and `OpenTelemetry.Api`, the latter of which in turn has just a single dependency on `System.Diagnostics.DiagnosticSource`.

I'm proposing that we obsolete the `OpenFeature.Contrib.Hooks.Otel` package and incorporate its contained code into the main `OpenFeature` project to provide first-class telemetry support in line with other popular projects in the .NET ecosystem.

## Is this a good idea?

From [`open-telemetry/opentelemetry-dotnet`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/extending-the-sdk/README.md#instrumentation-library) ___(emphasis added)___:

> The [inspiration of the OpenTelemetry project](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#instrumentation-libraries) is ___to make every library observable out of the box by having them call OpenTelemetry API directly.___ However, many libraries will not have such integration, and as such there is a need for a separate library which would inject such calls, using mechanisms such as wrapping interfaces, subscribing to library-specific callbacks, or translating existing telemetry into OpenTelemetry model.

From [`open-telemetry/opentelemetry-dotnet`](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#introduction) ___(emphasis added)___:

> Application developers and library authors use OpenTelemetry API to instrument their application/library. ___The API only surfaces necessary abstractions to instrument an application/library.___ It does not address concerns like how telemetry is exported to a specific telemetry backend, how to sample the telemetry, etc. The API consists of [Tracing API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md), [Logging API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/README.md), [Metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md), [Context and Propagation API](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/context), and a set of [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/trace.md).

## What if we think `OpenTelemetry.Api` is just too scary?

I personally have a lot of faith in the `open-telemetry/opentelemetry-dotnet` team, and thus place a lot of trust in their promise to maintain `OpenTelemetry.Api` as a lightweight abstractions package, but for good measure, I do want to call out that we could accomplish this proposed in-the-box'ing _without_ taking a dependency on `OpenTelemetry.Api`.

To do this we would instead take a direct dependency on `System.Diagnostics.DiagnosticSource`, and then reimplement-the-wheel for [`ActivityExtensions.RecordException(...)`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d0620e6f0f24bc4aa89143aaeddfac7a4fffb2aa/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L59-L105), [`ActivityExtensions.SetStatus(...)`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d0620e6f0f24bc4aa89143aaeddfac7a4fffb2aa/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L22-L57), etc.

Again, I'm not endorsing this, but it's an option (e.g. among many other reasons, it would be exceptionally peevish if our impl drifts from the upstream impl and exceptions recorded by another library have different formats in Datadog/Dynatrace/etc).

___

See: open-feature/dotnet-sdk-contrib#132

askpt:
## Move OTEL hooks to the SDK

This pull request introduces telemetry enhancements to the OpenFeature .NET SDK by adding new hooks for tracing and metrics, updating dependencies, and providing examples and tests. The most significant changes include the addition of the `TraceEnricherHook` and `MetricsHook` classes, integration of OpenTelemetry in the ASP.NET Core sample, and updates to dependencies to support telemetry features.

### Telemetry Enhancements

* **Trace Enricher Hook**: Added `TraceEnricherHook` to enrich telemetry traces with feature flag evaluation details, including tags and events for tracing purposes. This hook integrates with the current `Activity` and supports error handling.
* **Metrics Hook**: Introduced `MetricsHook` for capturing metrics such as evaluation requests, successes, errors, and active evaluations. Metrics are collected using OpenTelemetry's `Meter` API. [[1]](diffhunk://#diff-912b71a06f9a65012af403b04269f68f3cb9ee580d0dd62e1d6afc99bd433d31R1-R100) [[2]](diffhunk://#diff-9c7590e55694b19483ea6d802e6a1bb34418366a8f66d6f35b57ae81adb5bf17R1-R16)

### Dependency Updates

* **OpenTelemetry Dependencies**: Added `OpenTelemetry.Extensions.Hosting`, `OpenTelemetry.Instrumentation.AspNetCore`, and `OpenTelemetry.Exporter.OpenTelemetryProtocol` to the ASP.NET Core sample project to support telemetry features. [[1]](diffhunk://#diff-40299025547d304b834a53acda1eb9a8a6f49f2ec679b6bcebd3449211497c56R3-R18) [[2]](diffhunk://#diff-711ea17cbdebe419375c7684c8c39a1423d2bebcf8976ddd7bdd78deaab65b21R12)
* **DiagnosticSource Dependency**: Included `System.Diagnostics.DiagnosticSource` in the main project and centralized dependency management. [[1]](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R19) [[2]](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R33-R34)

### Documentation and Examples

* **README Updates**: Added documentation for `TraceEnricherHook` and `MetricsHook`, including detailed descriptions, examples, and usage instructions for integrating these hooks with OpenTelemetry.
* **ASP.NET Core Sample**: Updated the sample application to demonstrate the use of `TraceEnricherHook` and `MetricsHook` with OpenTelemetry tracing and metrics. [[1]](diffhunk://#diff-41550b31d77b5898b38a3280f8ffbc5d2531fc4d4884079ebf3c5e953a85075dR6-R32) [[2]](diffhunk://#diff-40299025547d304b834a53acda1eb9a8a6f49f2ec679b6bcebd3449211497c56R3-R18)

### Testing

* **MetricsHook Tests**: Added unit tests for `MetricsHook` to verify metrics collection during different stages of feature flag evaluation (e.g., `BeforeAsync`, `AfterAsync`, `ErrorAsync`, `FinallyAsync`).
### Related Issues

Fixes #175 

### Notes
* In this PR I made some changes to the `samples` application. This allow us to see the metrics and traces in any OTEL tool.

Check the screenshots below:

Screenshot 2025-06-25 at 18 58 59

### Follow-up Tasks

- We should remove any reference in the other repository and mark the Nuget package as deprecated.


beeme1mr:
Thanks for starting this PR but I think we should wait until [this spec PR](https://github.com/open-feature/spec/pull/287) is merged before moving this hook to the SDK.

src/OpenFeature/Hooks/TracingHook.cs
```diff
@@ -0,0 +1,54 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using OpenFeature.Model;
+
+namespace OpenFeature.Hooks;
+
+/// <summary>
+/// Stub.
```
kinyoklion, line 10: This doc doesn't appear correct.

askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/338/commits/345a1648714e8fb4eaea986082f47d8ab91ff0c1

askpt:
@toddbaert, @beeme1mr, @kinyoklion, @kylejuliandev, @arttonoyan and @austindrenski

I merged the OTEL semcov from #397 and replaced the similar attributes. Please have a look and let me know what you think.

README.md
```diff
@@ -510,6 +510,121 @@ services.AddOpenFeature(builder =>
...
+using OpenTelemetry;
+using OpenTelemetry.Trace;
+
+namespace OpenFeatureTestApp
+{
+    class Hello {
+        static void Main(string[] args) {
+
+			// set up the OpenTelemetry OTLP exporter
+			var tracerProvider = Sdk.CreateTracerProviderBuilder()
```
kylejuliandev, line 538: I think the whitespace here looks off, I think there are some tabs instead of spaces causing these lines to be further right than expected

README.md
```diff
@@ -510,6 +510,121 @@ services.AddOpenFeature(builder =>
...
+After running this example, you will be able to see the traces, including the events sent by the hook in your Jaeger UI.
+
+### Metrics Hook
+
+For this hook to function correctly a global `MeterProvider` must be set.
+`MetricsHook` performs metric collection by tapping into various hook stages.
+
+Below are the metrics extracted by this hook and dimensions they carry:
+
+| Metric key                             | Description                     | Unit         | Dimensions                          |
```
kylejuliandev, line 575: Are these documented anyway else beside the README? I've seen projects, like the [OpenTelemetry.Instrumentation.Runtime](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/eddf6c55126a980beda4ff535a9e4c228c4fec29/src/OpenTelemetry.Instrumentation.Runtime#metrics), have a seperate readme file with a more thorough breakdown of metrics being measured

src/OpenFeature/Hooks/MetricsHook.cs
```diff
@@ -0,0 +1,105 @@
...
+    private readonly Counter<long> _evaluationRequestCounter;
+    private readonly Counter<long> _evaluationSuccessCounter;
+    private readonly Counter<long> _evaluationErrorCounter;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MetricsHook"/> class.
+    /// </summary>
+    public MetricsHook()
+    {
+        var meter = new Meter(InstrumentationName, InstrumentationVersion);
```
kylejuliandev, line 34: If someone adds a hook at a client level we may end up newing up lots of Meters, I think it may be safer to create this once as a static field

```csharp
private static readonly Meter Meter = new(InstrumentationName, InstrumentationVersion);
```

kylejuliandev: I think the same should happen with the counters

src/OpenFeature/Hooks/MetricsHook.cs
```diff
@@ -0,0 +1,105 @@
...
+
+
+    /// <inheritdoc/>
+    public override ValueTask AfterAsync<T>(HookContext<T> context, FlagEvaluationDetails<T> details, IReadOnlyDictionary<string, object>? hints = null, CancellationToken cancellationToken = default)
+    {
+        var tagList = new TagList
+        {
+            { TelemetryConstants.Key, context.FlagKey },
+            { TelemetryConstants.Provider, context.ProviderMetadata.Name },
+            { TelemetryConstants.Variant, details.Variant ?? details.Value?.ToString() },
```
kylejuliandev, line 65: I'm guessing variant should never be null, but is it safe to fallback to the `details.Value?.ToString()`? I don't _think_ this would have high cardinality

beeme1mr: Cardinality is the largest concern here. It probably wouldn't be an issue but we've historically tried to avoid using value as a metric dimension. I'd advising leaving it out for now since it's much easier to add dimensions than remove them later.

src/OpenFeature/Hooks/MetricsHook.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/338/commits/bc38b0c58a894149433eecb6c794e03677b2ea5c

The counters don't need to be static. Internally, they do a `TryAdd`. Did a couple of tests, and the behaviour is the one expected; the meter only creates one counter.

README.md
askpt: Nice catch. Fixed with: https://github.com/open-feature/dotnet-sdk/pull/338/commits/cd9811e48e4d0b111c35f0119aee613a538381b2

README.md
askpt: I suggest breaking the README because it is getting quite big. If everyone agrees, I will do it in a later PR. cc: @beeme1mr and @toddbaert

src/OpenFeature/Hooks/MetricsHook.cs
askpt: I agree with both. I did a carbon copy of the OTEL metric implementation, but we can make improvements since we are trying to move.
Cardinality is a big issue here. @beeme1mr, do you want me to:
1. Allow "nulls"
2. Do not add `Variant` tag when there is no value
3. Remove the `Variant` tag altogether

README.md
beeme1mr: I personally don't mind large READMEs (within reason). Of course, breaking them out is an option, but it won't happen automagically but will be included in the openfeature.dev.

kinyoklion (approved):

src/OpenFeature/Hooks/MetricsHook.cs
kinyoklion: I'd tentatively say to remove it, to unblock the PR, and then it would be easy to add back once we have sorted the details.

askpt: Removed with https://github.com/open-feature/dotnet-sdk/pull/338/commits/904fd44e350c3d6dc0f74923a2a280950f2dd6cd

kylejuliandev (approved):

src/OpenFeature/Hooks/MetricsHook.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Diagnostics.Metrics;
+using System.Reflection;
+using OpenFeature.Constant;
+using OpenFeature.Model;
+using OpenFeature.Telemetry;
+
+namespace OpenFeature.Hooks;
+
+/// <summary>
+/// Represents a hook for capturing metrics related to flag evaluations.
```
beeme1mr, line 11: It might be a good idea to mark this as experimental because metrics haven't been officially spec'd

src/OpenFeature/Hooks/TracingHook.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <summary>
+    /// Adds tags and events to the current <see cref="Activity"/> for tracing purposes.
+    /// </summary>
+    /// <typeparam name="T">The type of the flag value being evaluated.</typeparam>
+    /// <param name="context">The hook context containing metadata about the evaluation.</param>
+    /// <param name="details">Details about the flag evaluation including the key, value, and variant.</param>
+    /// <param name="hints">Optional dictionary of hints that can modify hook behavior.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A completed <see cref="ValueTask"/> representing the asynchronous operation.</returns>
+    public override ValueTask AfterAsync<T>(HookContext<T> context, FlagEvaluationDetails<T> details,
```
beeme1mr: Can this move to the "finally" stage? It should contain all the data you need there to handle both the success and failure use cases.

src/OpenFeature/Hooks/TracingHook.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        Activity.Current?
+            .SetTag(TelemetryConstants.Key, details.FlagKey)
+            .SetTag(TelemetryConstants.Variant, details.Variant)
+            .SetTag(TelemetryConstants.Provider, context.ProviderMetadata.Name)
+            .AddEvent(new ActivityEvent("feature_flag", tags: new ActivityTagsCollection
+            {
+                [TelemetryConstants.Key] = details.FlagKey,
+                [TelemetryConstants.Variant] = details.Variant,
+                [TelemetryConstants.Provider] = context.ProviderMetadata.Name
```
beeme1mr: This seems to be missing many of the attributes we defined.

https://openfeature.dev/specification/appendix-d

kinyoklion: I think it would help to clarify the scope of this PR. Is it "moving the existing implementation" or is it "moving the existing implementation and updating it to the latest conventions". Or are we only moving hooks if they have been updated to the latest?

kinyoklion: It does potentially make sense to couple them for "in the box" support.

askpt: I initially suggested moving the project as it is and updating it along the way. I waited for the spec release to avoid breaking changes, but I’d like to close this PR since it’s been hanging for a while.

toddbaert (commented):
This seems good to me, but I want to make sure we're not putting anything into the SDK that we'll need to change soon. I know there's been some thrash with the sem-conv recently... @beeme1mr is there anything here you worry that we might need to change?

We might want to consider annotating this as experimental either way.

beeme1mr:
> I am waiting for feedback on how to proceed. If there are potential convention changes, we should mark them as experimental.

I'd really like to trace hook to match the latest spec. It's now marked as a release candidate and unlikely to change in a breaking way. Please also consider using the `finally` stage.

src/OpenFeature/Hooks/TracingHook.cs
askpt: @beeme1mr I added a few more that fit here. I think I am happy to release this as it is now and add more properties as we see fit. As mentioned, this is a move from the previous contrib repository, not a complete rewrite. Fixed with https://github.com/open-feature/dotnet-sdk/pull/338/commits/02003197816abcf7b9f47c95e5f918f6dbce00dc and https://github.com/open-feature/dotnet-sdk/pull/338/commits/ed543ddf34df0120425aeccec90170b170f4de4a

src/OpenFeature/Hooks/MetricsHook.cs
askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/338/commits/53092328388311359c3355c36c05ca5879ff0ca5

src/OpenFeature/Hooks/TracingHook.cs
askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/338/commits/4359a37954b8fcac29aeb95f204ed2ffb7bd9c0b

src/OpenFeature/Hooks/TracingHook.cs
```diff
@@ -0,0 +1,78 @@
...
+        };
+        if (!string.IsNullOrWhiteSpace(error.Message))
+        {
+            tagsCollection.Add(TracingConstants.AttributeExceptionMessage, error.Message);
+        }
+
+        Activity.Current?.AddEvent(new ActivityEvent(TracingConstants.AttributeExceptionEventName, default, tagsCollection));
+#endif
+        return base.ErrorAsync(context, error, hints, cancellationToken);
+    }
```
beeme1mr: Won't this overlap with the finally stage? In the JavaScript implementation, all the telemetry was captured in finally.

askpt:
@beeme1mr, @toddbaert, @kinyoklion, and @kylejuliandev, I made a couple of changes since your last review. Please look to see if they make sense.

beeme1mr (approved):

Issue #356, opened 2025-01-31
CI: Setup Merge Queues


beeme1mr:
Steps:
- Update repo config
- Add CLA workaround

beeme1mr:
FYI @toddbaert @askpt

PR #357, opened 2025-01-31
ci: add dco workaround

beeme1mr:
## This PR

- add dco workaround

### Related Issues

Relates to #356

### Notes

The DCO bot doesn't support merge queues. This is a workaround that we're using in other repos.


askpt (approved):
LGTM.

PR #387, opened 2025-02-24
feat: Add support for hook data.

kinyoklion:
## This PR

Adds support for hook data.
https://github.com/open-feature/spec/pull/273

### Notes

I realized that the 4.6.1 section of the spec wasn't consistent with the expected usage. Basically it over-specifies the typing of the hook data matching that of the evaluation context. That is one possible approach, it would just mean a bit more work on the part of the hook implementers. 

In the earlier example in the spec I put a `Span` in the hook data:
```
  public Optional<EvaluationContext> before(HookContext context, HookHints hints) {
    SpanBuilder builder = tracer.spanBuilder('sample')
    .setParent(Context.current().with(Span.current()));
    Span span = builder.startSpan()
    context.hookData.set("span", span);
  }
  public void after(HookContext context, FlagEvaluationDetails details, HookHints hints) {
    // Only accessible by this hook for this specific evaluation.
    Object value = context.hookData.get("span");
    if (value instanceof Span) {
      Span span = (Span) value;
      span.end();
    }
  }
```

This is only possible if the hook data allows specification of any `object` instead of being limited to the immutable types of a context. For hooks hook data this is safe because only the hook mutating the data will have access to that data. Additionally the execution of the hook will be in sequence with the evaluation (likely in a single thread).

The alternative would be to store data in the hook, and use the hook data to know when to remove it.
Something like this:
```
  public Optional<EvaluationContext> before(HookContext context, HookHints hints) {
    SpanBuilder builder = tracer.spanBuilder('sample')
    .setParent(Context.current().with(Span.current()));
    Span span = builder.startSpan()

    String storageId = Uuid();
    this.tmpData.set(storageId, span);
    context.hookData.set("span", storageId);
  }
  public void after(HookContext context, FlagEvaluationDetails details, HookHints hints) {
    // Only accessible by this hook for this specific evaluation.
    Object value = context.hookData.get("span");
    if (value) {
      String id = value.AsString();
      Span span= this.tmpData.get(id);
      span.end();
    }
  }
```


src/OpenFeature/HookRunner.cs
kinyoklion, line 1: Executing hooks seems complex enough to not directly be in the client.

src/OpenFeature/Model/HookContext.cs
```diff
@@ -10,20 +10,22 @@ namespace OpenFeature.Model
     /// <seealso href="https://github.com/open-feature/spec/blob/v0.5.2/specification/sections/04-hooks.md#41-hook-context"/>
     public sealed class HookContext<T>
     {
+        private readonly SharedHookContext<T> _shared;
```
kinyoklion, line 13: Given the combination of evaluation context updates with before hooks, and mutable hook data per hook-instance, the number of potential copies increases. I moved most of the hook context into this shared component. Then the per instance contexts are just referencing the same data.

This also made it easier to not instantiate a extra HookData. It could be moved back, and the first instance used for the first hook as an alternative to this approach.

src/OpenFeature/SharedHookContext.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public FlagValueType FlagValueType { get; } = flagValueType;
+
+        public ClientMetadata ClientMetadata { get; } =
+            clientMetadata ?? throw new ArgumentNullException(nameof(clientMetadata));
+
+        public Metadata ProviderMetadata { get; } =
+            providerMetadata ?? throw new ArgumentNullException(nameof(providerMetadata));
+
+        public HookContext<T> ToHookContext(EvaluationContext? evaluationContext)
```
kinyoklion, line 26: Convenient for using the shared context as a template and instantiating the per-instance data.

beeme1mr:
Hey @kinyoklion, thanks for kicking this off. I'll cut some GitHub issues for other languages once this PR is ready so we can use this as a reference.

src/OpenFeature/HookRunner.cs
```diff
@@ -0,0 +1,173 @@
...
+        private readonly ILogger _logger;
+
+        /// <summary>
+        /// Construct a hook runner instance. Each instance should be used for the execution of a single evaluation.
+        /// </summary>
+        /// <param name="hooks">
+        /// The hooks for the evaluation, these should be in the correct order for the before evaluation stage
+        /// </param>
+        /// <param name="evaluationContext">
+        /// The initial evaluation context, this can be updates as the hooks execute
```
chrfwow, line 32:
```suggestion
        /// The initial evaluation context, this can be updated as the hooks execute
```

src/OpenFeature/HookData.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <para>
+    /// This collection is intended for use only during the execution of individual hook stages, a reference
+    /// to the collection should not be retained.
+    /// </para>
+    /// <para>
+    /// This collection is not thread-safe.
+    /// </para>
+    /// </summary>
+    /// <seealso href="https://github.com/open-feature/spec/blob/main/specification/sections/04-hooks.md#46-hook-data"/>
+    public sealed class HookData
```
chrfwow, line 23: Not a blocker for me, but since this is a wrapper around a dict, maybe we could add a custom [indexer](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers)

chrfwow (approved):

test/OpenFeature.Tests/OpenFeatureHookTests.cs
```diff
@@ -394,7 +491,7 @@ public async Task Error_Hook_Should_Be_Executed_Even_If_Abnormal_Termination()
             hook1.BeforeAsync(Arg.Any<HookContext<bool>>(), null).Returns(EvaluationContext.Empty);
             hook2.BeforeAsync(Arg.Any<HookContext<bool>>(), null).Returns(EvaluationContext.Empty);
             featureProvider1.ResolveBooleanValueAsync(Arg.Any<string>(), Arg.Any<bool>(), Arg.Any<EvaluationContext>()).Throws(new Exception());
-            hook2.ErrorAsync(Arg.Any<HookContext<bool>>(), Arg.Any<Exception>(), null).Returns(new ValueTask());
+            hook2.ErrorAsync(Arg.Any<HookContext<bool>>(), Arg.Any<Exception>(), null).Throws(new Exception());
```
kinyoklion, line 494: The test description is for an `error` hook abnormally terminating, but it didn't have an error hook abnormally terminate. As a result the catch block in the error hook was untested.  Based on the ordering and test description I think this hook was supposed to throw.

src/OpenFeature/HookData.cs
kinyoklion: Added.

askpt (approved):

toddbaert:
> This is only possible if the hook data allows specification of any object instead of being limited to the immutable types of a context. For hooks hook data this is safe because only the hook mutating the data will have access to that data. Additionally the execution of the hook will be in sequence with the evaluation (likely in a single thread).

@kinyoklion I agree with your assertion that allowing any object is relatively safe, and certainly more ergonomic. I think we should adjust the spec.

test/OpenFeature.Tests/OpenFeatureHookTests.cs
toddbaert: Yes I think you're right.

test/OpenFeature.Tests/HookDataTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        Assert.Equal(5, asDictionary.Count);
+        Assert.Equal(true, asDictionary["bool"]);
+        Assert.Equal(1.2, asDictionary["double"]);
+        Assert.Equal(1.2f, asDictionary["float"]);
+        Assert.Equal(1, asDictionary["int"]);
+        Assert.Equal("string", asDictionary["string"]);
+    }
+
+    [Fact]
+    public void HookData_Get_Should_Throw_When_Key_Not_Found()
```
toddbaert, line 123: This particular behavior (throw if we can't retrieve a value) isn't specified. I'm not sure it needs to be, and I think I lean towards not specifying it, but I'm just calling it out here in case others disagree.

cc @chrfwow @askpt @beeme1mr

test/OpenFeature.Tests/OpenFeatureHookTests.cs
```diff
@@ -151,6 +164,90 @@ await client.GetBooleanValueAsync("test", false, EvaluationContext.Empty,
...
+            "`Hook data` **MUST** must be created before the first `stage` invoked in a hook for a specific evaluation and propagated between each `stage` of the hook. The hook data is not shared between different hooks.")]
+        public async Task HookData_Must_Be_Unique_Per_Hook()
+        {
+            var hook1 = Substitute.For<Hook>();
+            var hook2 = Substitute.For<Hook>();
+
+            hook1.BeforeAsync(Arg.Any<HookContext<bool>>(), Arg.Any<ImmutableDictionary<string, object>>())
+                .Returns(EvaluationContext.Empty).AndDoes(info =>
+                {
+                    info.Arg<HookContext<bool>>().Data.Set("hook-1-value-a", true);
```
toddbaert, line 209: If I'm reading correctly, I think we might be missing an assertion here wherein we'd set some duplicated property (a common key'd prop) on both hook-data instances, and ensure they don't "interfere" with each-other (I think that's implied in 4.3.2).

Please correct me if I'm misunderstanding.

test/OpenFeature.Tests/HookDataTests.cs
askpt: This implementation follows the dotnet behaviour for the other data structures. If we don't want to throw an exception, I would add a new method called `TryGet(string key, out object value)`. That would be the "safe way" to retrieve a value.

toddbaert: Ya that makes sense @askpt . I'm satisfied.

src/OpenFeature/HookData.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Set the key to the given value.
+        /// </summary>
+        /// <param name="key">The key for the value</param>
+        /// <param name="value">The value to set</param>
+        public void Set(string key, object value)
+        {
+            this._data[key] = value;
+        }
```
WeihanLi, line 34: Maybe we could return `HookData` to allow `hookData.Set("a", "aaa").Set("b", "bbb")` fluent set expression

```suggestion
        public HookData Set(string key, object value)
        {
            this._data[key] = value;
            return this;
        }
```

askpt: It is a great suggestion. Thanks for raising it 👍

kinyoklion: Added.

beeme1mr (approved):
Looks good to me! 

@kinyoklion could you please mentioning the new functionality briefly in the hook section of the readme and update the spec based on [this comment](https://github.com/open-feature/dotnet-sdk/pull/387#issuecomment-2744152718)?

kinyoklion:
> Looks good to me!
> 
> @kinyoklion could you please mentioning the new functionality briefly in the hook section of the readme and update the spec based on [this comment](https://github.com/open-feature/dotnet-sdk/pull/387#issuecomment-2744152718)?

https://github.com/open-feature/spec/pull/307

Also I added an example of a "timing" hook using hook data to the readme.

test/OpenFeature.Tests/OpenFeatureHookTests.cs
kinyoklion: Added an assertion using the same name for data in two different hooks with two different values.

PR #393, opened 2025-02-28
refactor!: improve EventExecutor

askpt:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523

### DIFF
--- a/src/OpenFeature/EventExecutor.cs
+++ b/src/OpenFeature/EventExecutor.cs
@@ -10,8 +10,6 @@ using OpenFeature.Model;
 
 namespace OpenFeature
 {
-    internal delegate Task ShutdownDelegate(CancellationToken cancellationToken);
-
     internal sealed partial class EventExecutor : IAsyncDisposable
     {
         private readonly object _lockObj = new object();
@@ -28,8 +26,7 @@ namespace OpenFeature
         public EventExecutor()
         {
             this._logger = NullLogger<EventExecutor>.Instance;
-            var eventProcessing = new Thread(this.ProcessEventAsync);
-            eventProcessing.Start();
+            Task.Run(this.ProcessEventAsync);
         }
 
         public ValueTask DisposeAsync() => new(this.ShutdownAsync());
@@ -234,7 +231,7 @@ namespace OpenFeature
                 switch (item)
                 {
                     case ProviderEventPayload eventPayload:
-                        this.UpdateProviderStatus(typedProviderRef, eventPayload);
+                        UpdateProviderStatus(typedProviderRef, eventPayload);
                         await this.EventChannel.Writer.WriteAsync(new Event { Provider = typedProviderRef, EventPayload = eventPayload }).ConfigureAwait(false);
                         break;
                 }
@@ -242,12 +239,14 @@ namespace OpenFeature
         }
 
         // Method to process events
-        private async void ProcessEventAsync()
+        private async Task ProcessEventAsync()
         {
             while (await this.EventChannel.Reader.WaitToReadAsync().ConfigureAwait(false))
             {
                 if (!this.EventChannel.Reader.TryRead(out var item))
+                {
                     continue;
+                }
 
                 switch (item)
                 {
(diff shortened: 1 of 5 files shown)


src/OpenFeature/EventExecutor.cs
```diff
@@ -186,42 +180,37 @@ private void EmitOnRegistration(FeatureProvider? provider, ProviderEventTypes ev
             }
             var status = provider.Status;
 
-            var message = "";
-            if (status == ProviderStatus.Ready && eventType == ProviderEventTypes.ProviderReady)
-            {
-                message = "Provider is ready";
-            }
-            else if (status == ProviderStatus.Error && eventType == ProviderEventTypes.ProviderError)
+            var message = status switch
```
toddbaert, line 183: nice. :sunglasses:

src/OpenFeature/EventExecutor.cs
```diff
@@ -234,82 +223,92 @@ private async void ProcessFeatureProviderEventsAsync(object? providerRef)
...
+            if (e.EventPayload?.Type != null && this._apiHandlers.TryGetValue(e.EventPayload.Type, out var eventHandlers))
+            {
+                foreach (var eventHandler in eventHandlers)
+                {
+                    this.InvokeEventHandler(eventHandler, e);
+                }
+            }
+        }
+
+        private void ProcessClientHandlers(Event e)
```
toddbaert, line 268: You removed a comment, but I guess the method name makes it clear what's happening now, so I'm fine with it.

toddbaert (approved):
Good improvements, thanks.

src/OpenFeature/EventExecutor.cs
```diff
@@ -10,26 +9,23 @@
...
+        private readonly Dictionary<string, Dictionary<ProviderEventTypes, List<EventHandlerDelegate>>> _clientHandlers = [];
 
         private ILogger _logger;
 
         public EventExecutor()
         {
             this._logger = NullLogger<EventExecutor>.Instance;
-            var eventProcessing = new Thread(this.ProcessEventAsync);
-            eventProcessing.Start();
+            Task.Run(this.ProcessEventAsync);
```
toddbaert, line 28: I should have caught this earlier.

toddbaert:
@askpt could you explain the breaking change? How could the event channel be null if we set it? I must be missing something.

I guess this would force providers to do a null check, right? I guess this would only break providers? I won't have any impact on consumers will it? I don't think we want to release a v3 for this.

askpt:
> @askpt could you explain the breaking change? How could the event channel be null if we set it? I must be missing something.
> 
> I guess this would force providers to do a null check, right? I guess this would only break providers? I won't have any impact on consumers will it? I don't think we want to release a v3 for this.

@toddbaert If you check test `OpenFeature_Should_Shutdown_Unused_Provider` you will see the test will fail if you don't do a null check in the method `StartListeningAndShutdownOld`, particularly on the line `oldProvider.GetEventChannel()?.Writer.Complete();`. It seems there is the possibility for the `EventChannel` to be null in this case, so it needs to be marked as a `nullable`. This is why I needed to mark it as null and the potential breaking change (even though it would be at compile time).

askpt:
> > > @askpt could you explain the breaking change? How could the event channel be null if we set it? I must be missing something.
> > > I guess this would force providers to do a null check, right? I guess this would only break providers? I won't have any impact on consumers will it? I don't think we want to release a v3 for this.
> > 
> > 
> > @toddbaert If you check test `OpenFeature_Should_Shutdown_Unused_Provider` you will see the test will fail if you don't do a null check in the method `StartListeningAndShutdownOld`, particularly on the line `oldProvider.GetEventChannel()?.Writer.Complete();`. It seems there is the possibility for the `EventChannel` to be null in this case, so it needs to be marked as a `nullable`. This is why I needed to mark it as null and the potential breaking change (even though it would be at compile time).
> 
> Isn't that just because we use [Substitute.For in that test](https://github.com/open-feature/dotnet-sdk/blob/9b6feab50085ee7dfcca190fe42f583c072ae50d/test/OpenFeature.Tests/OpenFeatureTests.cs#L47-L55) to create these providers?
> 
> What if we remove the `virtual` from the `GetEventChannel` accessor though? So it cannot be overridden... that might be a way to guarantee it cannot be null in any case, right? I think that technically still might be breaking (in case somebody overrode it) but it seems more correct and more ergonomic, and also "less breaking".

@toddbaert That is a good point. Let's do this. I removed the breaking change part, and we can discuss it on another topic. This PR is already complex enough. Plus, I need to investigate it better to be able to reply to your points! 👍

chrfwow (approved):

Issue #358, opened 2025-01-31
[BUG] EventExecutor exception


askpt:
### Description

We are currently experiencing some flaky unit tests that we should tackle. Example output:
```
Unhandled exception. System.Threading.Channels.ChannelClosedException: The channel has been closed.
   at OpenFeature.EventExecutor.ProcessFeatureProviderEventsAsync(Object providerRef) in /_/src/OpenFeature/EventExecutor.cs:line 238
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.Threading.Channels.ChannelClosedException: The channel has been closed.
   at OpenFeature.EventExecutor.ProcessFeatureProviderEventsAsync(Object providerRef) in /_/src/OpenFeature/EventExecutor.cs:line 238
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()


Test Run Aborted.
Total tests: Unknown
     Passed: 90
 Total time: 2.8570 Seconds
     6>_VSTestConsole:
         MSB4181: The "VSTestTask" task returned false but did not log an error.
     6>Done Building Project "/home/runner/work/dotnet-sdk/dotnet-sdk/test/OpenFeature.Tests/OpenFeature.Tests.csproj" (VSTest target(s)) -- FAILED.
```

As part of the debugging, we can try using the `dotnet test—- blame` to understand what is wrong.

### Notes
- https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test
- https://github.com/open-feature/dotnet-sdk/actions/runs/13074052418/job/36481801840

kinyoklion:
This looks to be a bug in the EventExecutor rather than just a flaky test. Writing to a channel will throw if the channel has been closed.

There is a purple note about it here: https://learn.microsoft.com/en-us/dotnet/core/extensions/channels#bounding-strategies

There isn't any synchronization between shutdown and the event processing loop.

```
            while (await reader.WaitToReadAsync().ConfigureAwait(false))
            {
                if (!reader.TryRead(out var item))
[...]

PR #401, opened 2025-03-13
fix: Remove virtual GetEventChannel from FeatureProvider

askpt:
## This PR

This pull request includes changes to the `EventChannel` handling in the `OpenFeature` library to improve code consistency and reliability.

Improvements to `EventChannel` handling:

* [`src/OpenFeature/EventExecutor.cs`](diffhunk://#diff-f563baadcf750bb66efaea76d7ec4783320e6efdc45c468effb531c948a2292cL150-R150): Removed unnecessary null-conditional operator when completing the `Writer` of the event channel.
* [`src/OpenFeature/FeatureProvider.cs`](diffhunk://#diff-96ebc8fc507d0a19d55b9a5cb57b72a0e8058e09f31ee7d0b39e99b00c5029d8L143-R143): Made the `GetEventChannel` method non-virtual to ensure consistent behavior across different implementations.

### Related Issues

Related to #358

### DIFF
--- a/src/OpenFeature/EventExecutor.cs
+++ b/src/OpenFeature/EventExecutor.cs
@@ -147,7 +147,7 @@ namespace OpenFeature
             if (oldProvider != null && !this.IsProviderBound(oldProvider))
             {
                 this._activeSubscriptions.Remove(oldProvider);
-                oldProvider.GetEventChannel()?.Writer.Complete();
+                oldProvider.GetEventChannel().Writer.Complete();
             }
         }
 
--- a/src/OpenFeature/FeatureProvider.cs
+++ b/src/OpenFeature/FeatureProvider.cs
@@ -140,7 +140,7 @@ namespace OpenFeature
         /// Returns the event channel of the provider.
         /// </summary>
         /// <returns>The event channel of the provider</returns>
-        public virtual Channel<object> GetEventChannel() => this.EventChannel;
+        public Channel<object> GetEventChannel() => this.EventChannel;
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


src/OpenFeature/EventExecutor.cs
```diff
@@ -147,7 +147,7 @@ private void StartListeningAndShutdownOld(FeatureProvider newProvider, FeaturePr
             if (oldProvider != null && !this.IsProviderBound(oldProvider))
             {
                 this._activeSubscriptions.Remove(oldProvider);
-                oldProvider.GetEventChannel()?.Writer.Complete();
```
toddbaert, line 150: I think we should do [this](https://github.com/open-feature/dotnet-sdk/pull/393#issuecomment-2715674476), or keep this check.

it's unlikely, but still technically possible that somebody overrides the `GetEventChannel` and returns null here unless we remove the virtual (which is technically breaking, but not in a way I think we need to worry about).

askpt: @toddbaert Actually, it is impossible to override and return `null` if you have the `nullable` setting enabled. But I agree with you, I prefer to remove virtual.

toddbaert: OK. I think we should not mark this breaking. Even if it technically is, in a very strict sense, I truly doubt anyone has overridden this, and if they did, they did something very wrong and risky. I'm going to remove the `!`.

toddbaert (approved):

kinyoklion (approved):

PR #422, opened 2025-04-09
test: Add E2E Steps for contextMerging.feature tests

### LINKED ISSUE #399: Add tests for contextMerging.feature
### Description
As part of the spec update, we need to add new tests. Until we add new spec tests, we won’t be able to merge #395.

### Notes
Check PR: https://github.com/open-feature/dotnet-sdk/pull/395

kylejuliandev:
## This PR

- Add new Reqnroll steps for contextMerging.feature ahead of merging #395 

### Related Issues

Fixes #399

### Notes

I was following the  [requirement 3.2.3](https://openfeature.dev/specification/sections/evaluation-context#requirement-323)

I'm assuming with these additions that "Before Hook" is the same as "Invocation" in the dotnet-sdk. If this is not the case then please let me know! 

In order to fetch the merged context I use a test hook with a function that can set the EvaluationContext on the state. There are probably better abstractions to use

I haven't updated the spec submodule in these changes either, although I can add that if needed

Let me know if you have any concerns or feedback 🏗️


askpt:
Thanks for your contribution, @kylejuliandev.
We need to complete two bits before we can merge this PR:
- Fix #423 (my fault, I will have a look later)
- Merge #395 into this branch to ensure the tests pass correctly.

kylejuliandev:
I've merged #395 into this branch now. Looks like CI builds pass (minus the nuget packaging due to attestations you mentioned)

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
```diff
@@ -57,6 +60,56 @@ public void GivenAString_FlagWithKeyAndADefaultValue(string key, string defaultT
         this.State.Flag = flagState;
     }
 
+    [Given("a stable provider with retrievable context is registered")]
+    public void GivenAStableProviderWithRetrievableContextIsRegistered()
+    {
+        var memProvider = new InMemoryProvider(E2EFlagConfig);
+        Api.Instance.SetProviderAsync(memProvider).Wait();
```
askpt, line 67: Why not `await` here? I noticed the step `GivenAStableProvider` is also not `awaiting`. Would you mind changing it there?

askpt (commented):
LGTM. I only have a small suggestion about the provider initialization.

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
kylejuliandev: Fixed - I can't remember why I didn't use await in the first place

askpt (approved):

beeme1mr (approved):

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
```diff
@@ -82,6 +135,57 @@ public async Task WhenTheFlagWasEvaluatedWithDetails()
                 break;
         }
     }
+    private void InitialiseContext(string level, EvaluationContext context)
```
chrfwow, line 138:
```suggestion
    private void InitializeContext(string level, EvaluationContext context)
```

test/OpenFeature.E2ETests/Steps/ContextMergingPrecedenceStepDefinitions.cs
```diff
@@ -0,0 +1,33 @@
...
+public class ContextMergingPrecedenceStepDefinitions : BaseStepDefinitions
+{
+    public ContextMergingPrecedenceStepDefinitions(State state) : base(state)
+    {
+    }
+
+    [When("Some flag was evaluated")]
+    public async Task WhenSomeFlagWasEvaluated()
+    {
+        this.State.Flag = new FlagState("boolean-flag", "true".ToString(), FlagType.Boolean);
```
chrfwow, line 19: why do we need the `.ToString()`?

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
```diff
@@ -82,6 +135,57 @@ public async Task WhenTheFlagWasEvaluatedWithDetails()
...
+                    {
+                        this.State.InvocationEvaluationContext = EvaluationContext.Builder()
+                            .Merge(context)
+                            .Merge(this.State.InvocationEvaluationContext!)
+                            .Build();
+                    }
+                    else
+                    {
+                        this.State.InvocationEvaluationContext = context;
+                    }
```
chrfwow, line 181: This context should be added in a Before Hook. In Java this looks like this:
```java
if ("Before Hooks".equals(level)) {
            state.client.addHooks(new Hook() {
                @Override
                public Optional<EvaluationContext> before(HookContext ctx, Map hints) {
                    return Optional.of(context);
                }
            });
        } 
```
See [here](https://github.com/open-feature/java-sdk/blob/main/src/test/java/dev/openfeature/sdk/e2e/steps/ContextSteps.java#L59). Invokation and Before Hooks should be different things

chrfwow (changes requested):
Looks good, but the Before Hook context needs some work

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
kylejuliandev: Fixed

test/OpenFeature.E2ETests/Steps/ContextMergingPrecedenceStepDefinitions.cs
kylejuliandev: Whoops, I must have left this over from some earlier refactor. Have removed the unnecessary tostring call

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
kylejuliandev: Ah! Thanks for the feedback, I borrowed some concepts from the implementation in Java. That should use the Before hook properly now

PR #442, opened 2025-04-21
refactor: InMemoryProvider throwing when types mismatched

### LINKED ISSUE #441: [BUG] In-Memory Provider throws exception
### Observed behavior

The InMemory provider is throwing an exception here: https://github.com/open-feature/dotnet-sdk/blob/456351216ce9113d84b56d0bce1dad39430a26cd/src/OpenFeature/Providers/Memory/InMemoryProvider.cs#L115.

We should move from throwing exceptions to returning an ErrorType with a default value. Something like: 
```csharp
return new ResolutionDetails<T>(flagKey, defaultValue, ErrorType.TypeMismatch, Reason.Error, $"flag {flagKey} is not of type {typeof(T)}");
```

### Expected Behavior

The In Memory provider doesn't throw any exceptions, and all unit tests pass.

### Steps to reproduce

N/A

kylejuliandev:
## This PR

- Update InMemoryProvider to return an ErrorType with a default value instead of throwing exceptions
- Add unit test to cover the new behavior

### Related Issues

Fixes #441

### DIFF
--- a/src/OpenFeature/Providers/Memory/InMemoryProvider.cs
+++ b/src/OpenFeature/Providers/Memory/InMemoryProvider.cs
@@ -3,7 +3,6 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using OpenFeature.Constant;
-using OpenFeature.Error;
 using OpenFeature.Model;
 
 namespace OpenFeature.Providers.Memory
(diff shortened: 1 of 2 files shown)


askpt (approved):
LGTM!

kylejuliandev:
> I'm not sure this is technically a bug, it's more of an optimization. It's valid for providers to throw, it's just less performant.
> 
> I'd suggest we adjust the prefix to "chore" or "refactor".

Happy to update the title - I went with fix because https://github.com/open-feature/dotnet-sdk/issues/441 was labelled as a bug

beeme1mr (approved):

PR #449, opened 2025-04-23
docs: update documentation on SetProviderAsync

### LINKED ISSUE #445: [DOC] Clarify the behavior of  SetProviderAsync
### Change in the documentation

The `SetProviderAsync` behaves differently than nearly every other method in OpenFeature because it can throw if the underlying provider throws during its initiation. This was a conscious design choice documented in the spec ([requirement 1.4.10](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1410)). However, it's not obvious in our examples that an error can be returned while setting a provider.

## Requirements

- Ensure that the XML Doc comments for the `SetProviderAsync` method clearly defines the behavior.
- Update the examples in the readme to use a try/catch when using `SetProviderAsync`.

kylejuliandev:
## This PR

- Clarifies the behaviour of SetProviderAsync. Exceptions can be thrown by a provider during initialization.

### Related Issues

Fixes #445

### DIFF
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -43,7 +43,7 @@ namespace OpenFeature
         /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
         /// await the returned task.
         /// </summary>
-        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
+        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect. May throw an exception if <paramref name="featureProvider"/> cannot be initialized.</remarks>
         /// <param name="featureProvider">Implementation of <see cref="FeatureProvider"/></param>
         public async Task SetProviderAsync(FeatureProvider featureProvider)
         {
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):
Perfect, thanks!

src/OpenFeature/Api.cs
```diff
@@ -43,7 +43,7 @@ private Api() { }
         /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
         /// await the returned task.
         /// </summary>
-        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect.</remarks>
+        /// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect. May throw an exception if <paramref name="featureProvider"/> cannot be initialized.</remarks>
```
askpt, line 46: I wonder if it would make sense to specify the exception type that can be thrown here.

askpt (approved):
LGTM! Just a small non blocking comment.

src/OpenFeature/Api.cs
kylejuliandev: ArgumentNullException is one possible exception (from the domain string), although potentially any exception could come from the Provider itself

I can clarify this if we want with a description like:

```
/// <remarks>The provider cannot be set to null. Attempting to set the provider to null has no effect. Will throw a <see cref="ArgumentNullException"/> when <paramref name="domain" /> is null or empty. May throw an exception if <paramref name="featureProvider"/> cannot be initialized.</remarks>
```

kylejuliandev: Ah my bad, I misunderstood what you said initially 🤣 I suppose adding an exception could be helpful for providers. Although, I imagine lower level exceptions, like for example a HttpRequestException, would be aggregated into this?

PR #459, opened 2025-04-29
feat: Add Extension Method for adding global Hook via DependencyInjection

### LINKED ISSUE #456: [FEATURE] Add Hooks to the global instance of OpenFeature using DI syntax
## Summary
This feature request aims to enhance the OpenFeature .NET SDK by enabling the addition of hooks to the global instance of OpenFeature using Dependency Injection (DI) syntax. This will align the SDK with common DI practices and improve its integration capabilities.

#### Additional Context:
As part of the DI library, the proposed implementation is expected to follow a syntax similar to the example below:

```csharp
builder.Services.AddOpenFeature(openfeatureBuilder =>
{
    openfeatureBuilder.AddHostedFeatureLifecycle()
    .AddProvider(_ => new FlagdProvider())
    .AddHook(_ => new LoggingHook());
});
```

This addition allows developers to seamlessly configure the OpenFeature global instance and integrate hooks using fluent builder methods.

kylejuliandev:
## This PR

- This adds a new OpenFeatureBuilder extension method to add global or not domain-bound hooks to the OpenFeature Api.

### Related Issues

Fixes #456

### Notes

We inject any provided Hooks as Singletons in the DI container. We use keyed singletons and use the class name as the key. Maybe we'd want to use a different name to avoid conflicts?

I've done some manual testing with a sample weatherforecast ASP.NET Core web application

### DIFF
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -262,4 +262,26 @@ public static partial class OpenFeatureBuilderExtensions
     /// <returns>The configured <see cref="OpenFeatureBuilder"/> instance.</returns>
     public static OpenFeatureBuilder AddPolicyName(this OpenFeatureBuilder builder, Action<PolicyNameOptions> configureOptions)
         => AddPolicyName<PolicyNameOptions>(builder, configureOptions);
+
+    /// <summary>
+    /// Adds a feature hook to the service collection using a factory method. Hooks added here are not domain-bound.
+    /// </summary>
+    /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="implementationFactory"></param>
+    /// <returns></returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook> implementationFactory)
+        where THook : Hook
+    {
+        var hookName = typeof(THook).Name;
+
+        builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHookName(hookName));
+
+        builder.Services.AddKeyedSingleton<Hook>(hookName, (serviceProvider, key) =>
+        {
+            return implementationFactory(serviceProvider);
+        });
+
+        return builder;
+    }
 }
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -269,24 +269,36 @@ public static partial class OpenFeatureBuilderExtensions
     /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
     /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
     /// <param name="implementationFactory">Optional factory for controlling how <typeparamref name="THook"/> will be created in the DI container.</param>
-    /// <returns></returns>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance.</returns>
     public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook>? implementationFactory = null)
         where THook : Hook
     {
-        var hookName = typeof(THook).Name;
+        return builder.AddHook(typeof(THook).Name, implementationFactory);
+    }
 
+    /// <summary>
+    /// Adds a feature hook to the service collection using a factory method and specified name. Hooks added here are not domain-bound.
+    /// </summary>
+    /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="hookName">The name of the <see cref="Hook"/> that is being added.</param>
+    /// <param name="implementationFactory">Optional factory for controlling how <typeparamref name="THook"/> will be created in the DI container.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance.</returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, string hookName, Func<IServiceProvider, THook>? implementationFactory = null)
+        where THook : Hook
+    {
         builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHookName(hookName));
 
         if (implementationFactory is not null)
         {
-            builder.Services.AddKeyedSingleton<Hook>(hookName, (serviceProvider, key) =>
+            builder.Services.TryAddKeyedSingleton<Hook>(hookName, (serviceProvider, key) =>
             {
                 return implementationFactory(serviceProvider);
             });
         }
         else
         {
-            builder.Services.AddKeyedSingleton(typeof(Hook), hookName, typeof(THook));
+            builder.Services.TryAddKeyedSingleton<Hook, THook>(hookName);
         }
 
         return builder;
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -268,19 +268,26 @@ public static partial class OpenFeatureBuilderExtensions
     /// </summary>
     /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
     /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
-    /// <param name="implementationFactory"></param>
+    /// <param name="implementationFactory">Optional factory for controlling how <typeparamref name="THook"/> will be created in the DI container.</param>
     /// <returns></returns>
-    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook> implementationFactory)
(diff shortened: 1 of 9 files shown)


src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -262,4 +262,26 @@ public static OpenFeatureBuilder AddPolicyName<TOptions>(this OpenFeatureBuilder
     /// <returns>The configured <see cref="OpenFeatureBuilder"/> instance.</returns>
     public static OpenFeatureBuilder AddPolicyName(this OpenFeatureBuilder builder, Action<PolicyNameOptions> configureOptions)
         => AddPolicyName<PolicyNameOptions>(builder, configureOptions);
+
+    /// <summary>
+    /// Adds a feature hook to the service collection using a factory method. Hooks added here are not domain-bound.
+    /// </summary>
+    /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="implementationFactory"></param>
```
kylejuliandev, line 271: Missing description for implementationFactory

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -262,4 +262,26 @@ public static OpenFeatureBuilder AddPolicyName<TOptions>(this OpenFeatureBuilder
...
+    /// <returns></returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook> implementationFactory)
+        where THook : Hook
+    {
+        var hookName = typeof(THook).Name;
```
kylejuliandev, line 276: `FeatureProvider`'s have a name or metadata tag which we use as the key when injecting them into the DI container. Here we use the Hook class name. Maybe we want to use something else incase hooks clash?

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -262,4 +262,26 @@ public static OpenFeatureBuilder AddPolicyName<TOptions>(this OpenFeatureBuilder
...
         => AddPolicyName<PolicyNameOptions>(builder, configureOptions);
+
+    /// <summary>
+    /// Adds a feature hook to the service collection using a factory method. Hooks added here are not domain-bound.
+    /// </summary>
+    /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="implementationFactory"></param>
+    /// <returns></returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook> implementationFactory)
```
askpt, line 273: I wonder if we should make `implementationFactory` optional. Then we can add a new hook using `Services.AddHook<NoOpHook>()`. In cases where the dependencies are already declared in the container, we should make use of it and not create a new ones.

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
askpt: They are being added as singleton. So if the hook already exists in the `container` should we really add them again? I would vote to do a `continue` and ignore adding more instances.

askpt (commented):
Looks good to me. Please look into the suggestions!

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
kylejuliandev: Great suggestion! I've marked `implementationFactory` as optional

This means when you're setting up OpenFeature you can now do something like

```csharp
builder.Services.AddOpenFeature(featureBuilder => {
    featureBuilder
        .AddHostedFeatureLifecycle() // From Hosting package
        .AddInMemoryProvider()
        .AddHook<LoggingHook>();
});
```

You still have the flexibility or control over how the hook can be created, when you provide the implementation factory

```csharp
builder.Services.AddOpenFeature(featureBuilder => {
    featureBuilder
        .AddHostedFeatureLifecycle() // From Hosting package
        .AddInMemoryProvider()
        .AddHook(serviceProvider => new LoggingHook());
});
```

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
askpt: Actually, in this case, what would happen is that the latest one to be `Added` is the one that gets resolved. We must use the `TryAdd` method to resolve with the first instantiation.

I feel that I should change it to `TryAdd`.

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
askpt: Beautiful!

askpt:
I forgot to add it. Can you please update the `Readme` as part of this PR?

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
kylejuliandev: The same issue can manifest when you add two providers, although in the providers case we use the Metadata name as the key when injecting into the container

I can change it to `.TryAddHook<Hook>()` if we prefer, not sure if the syntax is as nice as just `.AddHook<Hook>()`

askpt: Sorry for the confusion! I meant the implementation and not the method name! I don’t know, I have no real preference and it would be difficult to find another library that suffered from the same issue.

@toddbaert, @beeme1mr, @weyert, @arttonoyan Do you have any preference? In case of repeated Hook instantiation/declaration, what should we do?

A) Use the first declaration (then we change the implementation to `TryAdd`
B) we use the last declaration (as it is)

I think we should go by A route.

arttonoyan: @kylejuliandev Certainly I believe the current implementation is functionally correct, but I recommend switching to `TryAddKeyedSingleton`. TryAdd prevents accidental overrides and makes service registration more predictable. It’s a safer and more composable pattern, especially in shared or framework-level code. It also enables downstream consumers to override defaults more easily - something particularly useful in integration testing scenarios.

For example, in tests we can need to override a hook registered during application startup:

```csharp
builder.Services.TryAddKeyedSingleton<Hook>("MyHook", _ => new MyHook());
// Later in test
builder.Services.AddKeyedSingleton<Hook>("MyHook", _ => new TestHook()); // Overrides only if TryAdd was used initially
```

So I support **Option A** - resolving to the **first** registered hook - and suggest updating the implementation to use `TryAdd`.

Additionally, relying solely on `typeof(THook).Name` as the key limits flexibility and makes overrides more cumbersome. **Supporting explicit names provides better control for complex or test scenarios.** I suggest introducing an overload that accepts a custom `hookName`:

```csharp
public static OpenFeatureBuilder AddHook<THook>(
    this OpenFeatureBuilder builder,
    string hookName,
    Func<IServiceProvider, THook>? implementationFactory = null)
    where THook : Hook
```

The current method can then delegate to this overload:

```csharp
public static OpenFeatureBuilder AddHook<THook>(
    this OpenFeatureBuilder builder,
    Func<IServiceProvider, THook>? implementationFactory = null)
    where THook : Hook =>
        builder.AddHook(typeof(THook).Name, implementationFactory);
```

This approach preserves the existing API while allowing greater flexibility for consumers.

Finally, I recommend adding an integration test to validate this behavior. Here's a useful reference from the repo:  
📎 https://github.com/open-feature/dotnet-sdk/blob/main/test/OpenFeature.IntegrationTests/FeatureFlagIntegrationTest.cs

cc: @askpt

kylejuliandev: Sounds good to me, I'll revisit this tomorrow and make the recommended changes. Thanks for the help!

askpt (approved):

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -262,4 +262,45 @@ public static OpenFeatureBuilder AddPolicyName<TOptions>(this OpenFeatureBuilder
...
         => AddPolicyName<PolicyNameOptions>(builder, configureOptions);
+
+    /// <summary>
+    /// Adds a feature hook to the service collection using a factory method. Hooks added here are not domain-bound.
+    /// </summary>
+    /// <typeparam name="THook">The type of<see cref="Hook"/> to be added.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="implementationFactory">Optional factory for controlling how <typeparamref name="THook"/> will be created in the DI container.</param>
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance.</returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, Func<IServiceProvider, THook>? implementationFactory = null)
```
arttonoyan, line 273: **Suggestion**: Would it make sense to rename `AddHook` to `AddHooks` to better reflect that multiple hooks can be registered?
This would align with established .NET naming conventions — for example, YARP uses `AddTransforms<T>()` in a similar context: [YARP source](https://github.com/dotnet/yarp/blob/96aee06f1336fdef483ff45d126d298b8e202e94/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs#L121)

It's not critical at all, just a small naming improvement that might make the API a bit more intuitive.

Happy to hear your thoughts on this!
cc: @askpt

kinyoklion (approved):

PR #462, opened 2025-05-07
feat: add AddHandler extension method to Dependency Injection package

### LINKED ISSUE #457: [FEATURE] Add Events to the global instance of OpenFeature using DI syntax
### Requirements

_No response_

kylejuliandev:
## This PR

- Introduces a new extension method in the OpenFeature.DependencyInjection package which allows consumers to add, at a global/api level, a handler. 

Example usage:

```csharp
builder.Services.AddOpenFeature(feature =>
{
    feature.AddHostedFeatureLifecycle()
        .AddInMemoryProvider()
        .AddHandler(ProviderEventTypes.ProviderReady, (@event) => { Console.WriteLine("{0}", @event!.ProviderName); })
        .AddHandler(ProviderEventTypes.ProviderReady, sp => (@event) =>
        {
            var logger = sp.GetRequiredService<ILogger<Program>>();
            logger.LogInformation("Provider Ready");
        });
});
```

### Related Issues

Fixes #457

### DIFF
--- a/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
+++ b/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
@@ -43,6 +43,12 @@ internal sealed partial class FeatureLifecycleManager : IFeatureLifecycleManager
         }
 
         _featureApi.AddHooks(hooks);
+
+        foreach (var handlerName in options.HandlerNames)
+        {
+            var handler = _serviceProvider.GetRequiredKeyedService<EventHandlerDelegateWrapper>(handlerName);
+            _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
+        }
     }
 
     /// <inheritdoc />
--- a/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
+++ b/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
@@ -44,13 +44,10 @@ internal sealed partial class FeatureLifecycleManager : IFeatureLifecycleManager
 
         _featureApi.AddHooks(hooks);
 
-        foreach (var handlerName in options.HandlerNames)
+        var handlers = _serviceProvider.GetServices<EventHandlerDelegateWrapper>();
+        foreach (var handler in handlers)
         {
-            var handlers = _serviceProvider.GetKeyedServices<EventHandlerDelegateWrapper>(handlerName);
-            foreach (var handler in handlers)
-            {
-                _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
-            }
+            _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
         }
     }
 
--- a/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
+++ b/src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
@@ -46,8 +46,11 @@ internal sealed partial class FeatureLifecycleManager : IFeatureLifecycleManager
 
         foreach (var handlerName in options.HandlerNames)
         {
-            var handler = _serviceProvider.GetRequiredKeyedService<EventHandlerDelegateWrapper>(handlerName);
-            _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
+            var handlers = _serviceProvider.GetKeyedServices<EventHandlerDelegateWrapper>(handlerName);
+            foreach (var handler in handlers)
+            {
+                _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
+            }
         }
     }
 
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -315,7 +315,7 @@ public static partial class OpenFeatureBuilderExtensions
     /// <returns></returns>
     public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, ProviderEventTypes type, EventHandlerDelegate eventHandlerDelegate)
     {
-        return AddHandler(builder, string.Empty, type, sp => eventHandlerDelegate);
+        return AddHandler(builder, typeof(EventHandlerDelegate).Name, type, sp => eventHandlerDelegate);
     }
 
     /// <summary>
@@ -340,7 +340,7 @@ public static partial class OpenFeatureBuilderExtensions
     /// <returns></returns>
     public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, ProviderEventTypes type, Func<IServiceProvider, EventHandlerDelegate> implementationFactory)
     {
-        return AddHandler(builder, string.Empty, type, implementationFactory);
+        return AddHandler(builder, typeof(EventHandlerDelegate).Name, type, implementationFactory);
     }
 
     /// <summary>
@@ -353,14 +353,14 @@ public static partial class OpenFeatureBuilderExtensions
     /// <returns></returns>
     public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, string handlerName, ProviderEventTypes type, Func<IServiceProvider, EventHandlerDelegate> implementationFactory)
     {
-        if (string.IsNullOrEmpty(handlerName))
-            handlerName = Guid.NewGuid().ToString();
+        if (string.IsNullOrWhiteSpace(handlerName))
+            handlerName = string.Empty;
 
         var key = string.Join(":", handlerName, type.ToString());
 
         builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHandlerName(key));
 
-        builder.Services.TryAddKeyedSingleton(key, (serviceProvider, _) =>
+        builder.Services.AddKeyedSingleton(key, (serviceProvider, _) =>
         {
             var handler = implementationFactory(serviceProvider);
             return new EventHandlerDelegateWrapper(type, handler);
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -1,7 +1,9 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
+using OpenFeature.Constant;
 using OpenFeature.DependencyInjection;
+using OpenFeature.DependencyInjection.Internal;
 using OpenFeature.Model;
 
 namespace OpenFeature;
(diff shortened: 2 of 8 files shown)


src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -303,4 +305,73 @@ public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder,
...
+    }
+
+    static readonly Random _random = new();
+    private static string GenerateRandomName()
+    {
+        const string characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+        const int length = 4;
+
+        return new string([.. Enumerable.Range(0, length).Select(_ => characters[_random.Next(characters.Length)])]);
+    }
```
beeme1mr, line 376: Could you please clarify why this was done? I'm not that familiar with .NET DI. Thanks!

beeme1mr (approved):
Looks good to me. I just had a question about the random name generator but it isn't a blocker.

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
kylejuliandev: If we register or add multiple dependencies with the same key (or same implementation pattern) only the last is actually kept

```csharp
services.AddSingleton<MyLogger>(sp => /* implementation logic 1 */);
services.AddSingleton<MyLogger>(sp => /* implementation logic 2 */);
```

^ Only the last call to `AddSingleton<MyLogger>` is kept in this instance. If a consumer doesn't specify a name (we use name/key synonymously) for the Handler we generate a random one so they can add as many like - without overwriting the last

I went down this path because adding handlers directly via `Api.Instance.AddHandler(...)` is additive, so you can add as many as you like

I admit though this isn't a pattern I think you see often, so I'm happy to take it out if we'd prefer the alternative behaviour (adding multiple with no name only keeps the last)

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -303,4 +305,73 @@ public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder,
...
+    /// Add a <see cref="EventHandlerDelegate"/> to allow you to react to state changes in the provider or underlying flag management system, such as flag definition changes, provider readiness, or error conditions
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="handlerName">The name of the <see cref="EventHandlerDelegate"/>.</param>
+    /// <param name="type">The type <see cref="ProviderEventTypes"/> to handle.</param>
+    /// <param name="implementationFactory">The handler factory for creating a handler which reacts to <see cref="ProviderEventTypes"/>.</param>
+    /// <returns></returns>
+    public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, string handlerName, ProviderEventTypes type, Func<IServiceProvider, EventHandlerDelegate> implementationFactory)
+    {
+        var key = string.Join(":", handlerName, type.ToString());
```
askpt, line 356: I would check if the handler name is empty. If it is empty then generate a random name.

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
askpt: I think this is a bit of overkill. I would prefer you to use a Guid which is simpler to use and it is random enough.

askpt (commented):
Looks good to me. Great work. Added a couple of suggestions 👍

askpt (commented):
It is a very minor request! We should also validate for inputs like whitespace. 

Also, @arttonoyan, would you mind double-checking?

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
arttonoyan: I understand the intent behind generating a `Guid` when `handlerName` is not provided - to avoid overwriting existing handlers - but I have a few reservations:

* **Overriding the previous handler is often intentional** and expected in .NET DI, particularly in test scenarios or adapter-based configurations.
* We already provide an overload that accepts a `handlerName`, so consumers who want multiple distinct handlers can explicitly opt into that.

**Suggestion:**
Instead of generating a random `Guid`, I recommend using `typeof(T).Name` as the default `handlerName` when none is provided. This keeps the key predictable, improves traceability, and avoids registering handlers under opaque, unreachable names.

arttonoyan: **Additional Suggestion:**
Consider introducing a generic `AddHandler<T>` extension method that supports both named and unnamed registrations. This would improve type safety and provide a cleaner experience by defaulting `handlerName` to `typeof(T).Name` when not explicitly specified.

---

**Example: Registering Multiple Implementations**

Consumers can register multiple implementations using either of the following patterns:

**1. Inheritance-based registration**
Define distinct classes inheriting from the same base type:

```csharp
class MyLoggerV1 : MyLogger { }
class MyLoggerV2 : MyLogger { }

services.AddSingleton<MyLoggerV1>(sp => /* implementation logic 1 */);
services.AddSingleton<MyLoggerV2>(sp => /* implementation logic 2 */);
```

**2. Named registration**
Use the same base type and differentiate using keys:

```csharp
services.AddSingleton<MyLogger>("MyLoggerV1", sp => /* implementation logic 1 */);
services.AddSingleton<MyLogger>("MyLoggerV2", sp => /* implementation logic 2 */);
```

These approaches help avoid accidental replacement and clarify intent—whether by using explicit types or consistent naming, without relying on arbitrary or implicit keys.

arttonoyan: @kylejuliandev @askpt What do you think, guys?

askpt: @arttonoyan You have a pretty valid point there. Also, for `Hook` we are using the `TryAdd`, which mimics what you explained in your review comments.

@kylejuliandev What are your thoughts?

kylejuliandev: I'm not sure we can use `typeof(T).Name` for this because the Handler is a delegate, so I wouldn't be able to constrain T

I agree with other comments though around generating unique names for the dependency key and using a consistent behaviour for overwriting any previously injected handlers

I'll make some changes to address the outstanding comments 👍

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -353,25 +353,19 @@ public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, Pro
...
     {
+        if (string.IsNullOrWhiteSpace(handlerName))
+            handlerName = string.Empty;
+
         var key = string.Join(":", handlerName, type.ToString());
 
         builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHandlerName(key));
 
-        builder.Services.TryAddKeyedSingleton(key, (serviceProvider, _) =>
+        builder.Services.AddKeyedSingleton(key, (serviceProvider, _) =>
```
askpt, line 363:
```suggestion
        builder.Services.TryAddKeyedSingleton(key, (serviceProvider, _) =>
```
It was like this. Did you revert it?

askpt (commented):
LGTM. Just a minor question, and should be ready to merge! Thanks!

arttonoyan (changes requested):

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -303,4 +305,67 @@ public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder,
...
+    /// </summary>
+    /// <param name="builder">The <see cref="OpenFeatureBuilder"/> instance.</param>
+    /// <param name="handlerName">The name of the <see cref="EventHandlerDelegate"/>.</param>
+    /// <param name="type">The type <see cref="ProviderEventTypes"/> to handle.</param>
+    /// <param name="implementationFactory">The handler factory for creating a handler which reacts to <see cref="ProviderEventTypes"/>.</param>
+    /// <returns></returns>
+    public static OpenFeatureBuilder AddHandler(this OpenFeatureBuilder builder, string handlerName, ProviderEventTypes type, Func<IServiceProvider, EventHandlerDelegate> implementationFactory)
+    {
+        if (string.IsNullOrWhiteSpace(handlerName))
+            handlerName = string.Empty;
```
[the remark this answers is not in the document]
askpt: Nice catch! I agree with your point. In the first instance we were generating names in the absence of one. Now, throwing makes sense since name is not passed and we are not generating random names as fallback.

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
kylejuliandev: Yeah - so I think the _desired_ behaviour is to have `AddHandler` be additive. Everytime a consumer calls the extension method the delegate they provided is added. `TryAdd` would limit this behaviour as multiple calls (with the name Handler name) would only result in one being added

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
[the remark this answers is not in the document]
arttonoyan: @kylejuliandev Apologies - I forgot to mention earlier: please use [Guard.ThrowIfNullOrWhiteSpace](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature.DependencyInjection/Guard.cs#L15) so we maintain a consistent standard throughout the codebase.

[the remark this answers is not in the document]
kylejuliandev: Good spot! I've updated the code to use `Guard.ThrowIfNullOrWhiteSpace`

askpt (approved):

askpt:
LGTM! If @arttonoyan is happy with the changes @kylejuliandev made, we can merge it 👍 
Thanks for your efforts!

src/OpenFeature.DependencyInjection/Internal/FeatureLifecycleManager.cs
```diff
@@ -43,6 +43,15 @@ public async ValueTask EnsureInitializedAsync(CancellationToken cancellationToke
...
 
         _featureApi.AddHooks(hooks);
+
+        foreach (var handlerName in options.HandlerNames)
+        {
+            var handlers = _serviceProvider.GetKeyedServices<EventHandlerDelegateWrapper>(handlerName);
+            foreach (var handler in handlers)
+            {
+                _featureApi.AddHandler(handler.ProviderEventType, handler.EventHandlerDelegate);
+            }
```
arttonoyan, line 53: Could you clarify the rationale behind using `GetKeyedServices<EventHandlerDelegateWrapper>(handlerName)` here?
From what I see, `EventHandlerDelegateWrapper` is internal, so consumers can't resolve or register it directly. Also, if someone registers the same handler multiple times the same delegate will be added multiple times to `_featureApi` and invoked repeatedly.  
If that's the intended behavior, could you elaborate on the use of `handlerName` as the key - especially since duplicate handler**s** can be registered?

```
if (eventHandlerDelegate is not null)
{
    cfg.AddHandler(ProviderEventTypes.ProviderReady, eventHandlerDelegate);
    cfg.AddHandler(ProviderEventTypes.ProviderReady, eventHandlerDelegate);
}
```

kylejuliandev: `_handlerNames` in [OpenFeatureOptions.cs](https://github.com/open-feature/dotnet-sdk/pull/462/files#diff-e7c45146e3d637452be99b10fdbf11e287dcc3bbefd7f2aee5d4fa35f0e7fead) are stored as a hash map, so we don't get duplicate names. `GetKeyedServices<EventHandlerDelegateWrapper>(handlerName)` will return all handlers added with that key, but they will only be registered once

You raise a good point though, if we want to be able to add multiple handlers (with the same key) and it not be overwritten - do we need a `handlerName` at all? We could register them as a singleton regardless of handler name and resolve them all in [FeatureLifecycleManager.cs](https://github.com/open-feature/dotnet-sdk/pull/462/files#diff-c6ad1257f6605d20e7bd5be669d63cc3134a289142ca93a9977a16b22ffa2f88)

arttonoyan: @kylejuliandev Thanks, that’s a great point. To validate the behavior, could you please add an integration test with this configuration to ensure everything works as expected?

Specifically, in the test setup, try registering the same handler twice:

```csharp
if (eventHandlerDelegate is not null)
{
    cfg.AddHandler(ProviderEventTypes.ProviderReady, eventHandlerDelegate);
    cfg.AddHandler(ProviderEventTypes.ProviderReady, eventHandlerDelegate);
}
```

Then, in the `VerifyHandlerIsRegisteredAsync` test, increment a counter within the handler registration like this:

```csharp
int count = 0;
using var server = await CreateServerAsync(ServiceLifetime.Transient, logger, configureServices, (_) =>
{
    count++;
    handlerSuccess = true;
}).ConfigureAwait(true);
```

My intention here is to confirm how the current setup behaves when multiple handlers are registered under the same key. I understand we’re using a `HashSet` to register handler names, so duplicates aren’t stored at the registration level. However, in `FeatureLifecycleManager`, we're retrieving them using `GetKeyedServices`, which returns all services registered under that name - as per the design of Microsoft.Extensions.DependencyInjection.

test/OpenFeature.IntegrationTests/FeatureFlagIntegrationTest.cs
```diff
@@ -115,9 +122,40 @@ public async Task VerifyHandlerIsRegisteredAsync()
...
+            .ConfigureAwait(true);
+
+        var client = server.CreateClient();
+        var requestUri = $"/features/{TestUserId}/flags/{FeatureA}";
+
+        // Act
+        var response = await client.GetAsync(requestUri).ConfigureAwait(true);
+
+        // Assert
+        Assert.Equal(2, counter);
```
arttonoyan, line 151: @askpt @beeme1mr Can we confirm - is this the intended behavior? It seems we’ll get the same result even when using named handlers.

If the counter ends up as `2`, then it means all registered handlers under the same name are being executed. In that case, what's the purpose of using named handlers? It appears that any number of handlers can be added under the same name and they will all be invoked.

```csharp
var @lock = new Lock();
var counter = 0;
Action<OpenFeatureBuilder> openFeatureBuilder = cfg =>
{
    cfg.AddHandler("test", ProviderEventTypes.ProviderReady, (_) => { lock (@lock) { counter++; } });
    cfg.AddHandler("test", ProviderEventTypes.ProviderReady, (_) => { lock (@lock) { counter++; } });
};
```

If this is expected, perhaps we should clarify the semantics behind handler naming. Otherwise, it might give the impression that the name implies uniqueness, which isn’t enforced at invocation time.

kylejuliandev: I'd be happy removing the "named" constraint entirely. I think I included initially because of how the other extension methods were formed. But for consistency with the underlying `Api.Instance.AddHandler` code, maybe it would be best to remove names entirely in this PR

beeme1mr: Yeah, I would remove it for now. We don't have domain support in the global event handle yet. It may make sense to add in the future but I wouldn't include it in this pr. Good catch!

arttonoyan (approved):

kylejuliandev:
Thank you @askpt @arttonoyan @beeme1mr! This ended up being a code review and API review all in one; I appreciate the collaboration!

PR #469, opened 2025-05-14
test: Fix flaky integration flaky

### LINKED ISSUE #468: Investigate flaky test
> Not certain why, it looks like I potentially introduced a flaky test in #462. I will take a look later today and see if I can reproduce it

_Originally posted by @kylejuliandev in https://github.com/open-feature/dotnet-sdk/issues/415#issuecomment-2880571891_

We have a flaky test: `VerifyMultipleHandlersAreRegisteredAsync`. We need to investigate why multiple handlers are registered (4, 3 vs 2.).

kylejuliandev:
### This PR

Reworked the `VerifyMultipleHandlersAreRegisteredAsync` test assertions in the Integration tests to make them more reliable.

### Releated issues

Fixes #468 

### Notes

Before we were asserting that the handler was invoked twice. However, we don't necessarily need to test how many invocations occured but rather that the invocations happened. 

I've tweaked the counter to increment when the handler is injected (which should always be twice) and added a couple of boolean variables that we can use to track whether each handler did invoke at least once. I think this makes for a more accurate and reliable test case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         run: dotnet build -c Release --no-restore
 
       - name: Test
-        run: dotnet test -c Release --no-build --logger GitHubActions
+        run: dotnet test -c Release --no-build --logger "console;verbosity=detailed"
 
   packaging:
     needs: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         run: dotnet build -c Release --no-restore
 
       - name: Test
-        run: dotnet test -c Release --no-build --logger "console;verbosity=detailed"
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         run: dotnet build -c Release --no-restore
 
       - name: Test
-        run: dotnet test -c Release --no-build --logger GitHubActions
+        run: dotnet test -c Release --no-build --logger "console;verbosity=detailed"
```
askpt, line 46:
```suggestion
        run: dotnet test --blame -c Release --no-build --logger "console;verbosity=detailed"
```
@kylejuliandev months ago, I came across this "tweet": https://bsky.app/profile/davidfowl.com/post/3lflw4apqgc2w
Might worth a try

askpt (approved):

PR #488, opened 2025-06-06
feat: Add multi-provider support

### LINKED ISSUE #487: Add multi-provider support
This issue proposes to add support for multiple feature flag providers within the OpenFeature .NET SDK.

### Motivation
Supporting multi-provider functionality would allow users to integrate and delegate feature flag evaluations across several providers in the same application. This improves flexibility and enables complex use cases where different providers might be required for different environments or flag types.

### Reference
For more details on the multi-provider concept in OpenFeature, please see the official specification:  
https://openfeature.dev/specification/appendix-a#multi-provider

askpt:
## This PR

This pull request introduces the experimental "Multi-Provider" feature to the OpenFeature library, enabling simultaneous use of multiple feature flag providers with configurable evaluation strategies. The changes include updates to documentation, new classes and methods to support multi-provider functionality, and a sample implementation in `AspNetCore`. Below is a summary of the most important changes grouped by theme.

### Documentation Updates:
* Added "Multi-Provider" to the feature list in `README.md`, marking it as experimental.
* Included a detailed section in `README.md` explaining the usage, evaluation strategies, modes, and limitations of the Multi-Provider feature.

### Sample Implementation:
* Updated `samples/AspNetCore/Program.cs` to demonstrate Multi-Provider usage, including flag evaluation and error handling. [[1]](diffhunk://#diff-41550b31d77b5898b38a3280f8ffbc5d2531fc4d4884079ebf3c5e953a85075dR6-R8) [[2]](diffhunk://#diff-41550b31d77b5898b38a3280f8ffbc5d2531fc4d4884079ebf3c5e953a85075dR62-R127)

### Core Multi-Provider Functionality:
* Introduced the `MultiProvider` class in `src/OpenFeature/Providers/MultiProvider/MultiProvider.cs`, implementing support for multiple providers, evaluation strategies, initialization, and shutdown logic.
* Added the `ProviderEntry` class to represent individual providers in the Multi-Provider configuration.

### Supporting Models and Extensions:
* Created `ProviderStatus` and `RegisteredProvider` models to track provider states and metadata. [[1]](diffhunk://#diff-1554ff27d24fe48dab38c38e756f586c786750861e7c15679ae6ee955eace066R1-R7) [[2]](diffhunk://#diff-8805dd0881e151c53f45f8f0e6dc51ac29c7e24db4ef5f9adadb5b2997d9d664R1-R21)
* Added `ProviderExtensions` to handle evaluation logic for individual providers.

### Related Issues

Fixes #487 

### Notes

I didn't do proper testing for `Hooks` and `Events`.  This PR is already quite big, and I would appreciate some feedback on the overall design.

### Follow-up Tasks

- [ ] Add Hooks and Events support for Multi-Provider.
- [ ] Add Dependency Injection support.

### How to test

Open the `samples` app and enjoy!


askpt:
> Do you have any idea when this will be implemented in the NuGet package?

@sergheevxo I am actively working on #338, and this feature is next on my list.

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
```diff
@@ -0,0 +1,79 @@
...
+        ProviderResolutionResult<T>? fallbackResolution = null;
+
+        // Find fallback provider if specified
+        if (this._fallbackProvider != null)
+        {
+            fallbackResolution = successfulResolutions.FirstOrDefault(r => ReferenceEquals(r.Provider, this._fallbackProvider));
+        }
+
+        // Check if all successful results agree on the value
+        var allAgree = successfulResolutions.All(r => EqualityComparer<T>.Default.Equals(r.ResolutionDetails.Value, firstResult.ResolutionDetails.Value));
```
kylejuliandev, line 47: Does `Value` have a equality comparer? I assume this would always be false?

`bool` and `int` are relatively easy to compare, `double` might be harder because different providers may surface different levels of precision

src/OpenFeature/Providers/MultiProvider/Strategies/FirstMatchStrategy.cs
```diff
@@ -0,0 +1,35 @@
+using OpenFeature.Constant;
+using OpenFeature.Model;
+using OpenFeature.Providers.MultiProvider.Strategies.Models;
+
+namespace OpenFeature.Providers.MultiProvider.Strategies;
+
+/// <summary>
+/// Return the first result that did not indicate "flag not found".
```
kylejuliandev, line 8: Is it worth including a comment that FirstMatch is sequential? Like in the ComparisonStrategy

src/OpenFeature/Providers/MultiProvider/Strategies/FirstMatchStrategy.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /// <inheritdoc/>
+    public override FinalResult<T> DetermineFinalResult<T>(StrategyEvaluationContext strategyContext, string key, T defaultValue, EvaluationContext? evaluationContext, List<ProviderResolutionResult<T>> resolutions)
+    {
+        var lastResult = resolutions.LastOrDefault();
+        if (lastResult != null)
+        {
+            return ToFinalResult(lastResult);
+        }
```
kylejuliandev, line 27: Should we find the first result, not the last?

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <inheritdoc/>
+    public override FinalResult<T> DetermineFinalResult<T>(StrategyEvaluationContext strategyContext, string key, T defaultValue, EvaluationContext? evaluationContext, List<ProviderResolutionResult<T>> resolutions)
+    {
+        var successfulResolutions = resolutions.Where(r => !HasError(r)).ToList();
+
+        if (successfulResolutions.Count == 0)
+        {
+            var errorDetails = new ResolutionDetails<T>(key, defaultValue, ErrorType.ProviderNotReady, Reason.Error, errorMessage: "No providers available or all providers failed");
+            var errors = resolutions.Select(r => new ProviderError(r.ProviderName, new InvalidOperationException($"Provider {r.ProviderName} failed"))).ToList();
+            return new FinalResult<T>(errorDetails, null!, "MultiProvider", errors);
```
kylejuliandev: We refer to the name `"MultiProvider"` in a few places, is it worth creating an internal (or public) constant?

src/OpenFeature/Providers/MultiProvider/Strategies/FirstSuccessfulStrategy.cs
```diff
@@ -0,0 +1,46 @@
+using OpenFeature.Constant;
+using OpenFeature.Model;
+using OpenFeature.Providers.MultiProvider.Strategies.Models;
+
+namespace OpenFeature.Providers.MultiProvider.Strategies;
+
+/// <summary>
+/// Return the first result that did not result in an error.
```
kylejuliandev, line 8: Is it worth calling out this startegy executes sequentially? Might be overkill as the README documents this

src/OpenFeature/Providers/MultiProvider/Strategies/Models/StrategyEvaluationContext.cs
```diff
@@ -0,0 +1,28 @@
...
+/// Evaluation context specific to strategy evaluation containing flag-related information.
+/// </summary>
+public class StrategyEvaluationContext
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="StrategyEvaluationContext"/> class.
+    /// </summary>
+    /// <param name="flagKey">The feature flag key being evaluated.</param>
+    /// <param name="flagType">The type of the flag value being evaluated.</param>
+    public StrategyEvaluationContext(string flagKey, Type flagType)
```
kylejuliandev: You've opted for `Type flagType` here, would a type constraint also work? Within any one evaluation context Type will always be the same as `T`?

src/OpenFeature/Providers/MultiProvider/Strategies/Models/StrategyPerProviderContext.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        this.Provider = provider;
+        this.ProviderName = providerName;
+        this.ProviderStatus = providerStatus;
+    }
+
+    /// <summary>
+    /// The feature provider instance.
+    /// </summary>
+    public FeatureProvider Provider { get; private set; }
```
kylejuliandev: Small nit, I think you can omit the `private set;`

Although this does mean the property becomes readonly (which it effectively is in this case)

askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/bbdf85d191f8d6d4749588f6ba76f1c55cc76181

src/OpenFeature/Providers/MultiProvider/Strategies/FirstMatchStrategy.cs
askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/19fd5892707d5b359c4d25963cc9f18f1aaafc33

src/OpenFeature/Providers/MultiProvider/Strategies/FirstSuccessfulStrategy.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/19fd5892707d5b359c4d25963cc9f18f1aaafc33

src/OpenFeature/Providers/MultiProvider/Strategies/Models/StrategyEvaluationContext.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/1e293cddf312ded39f262c7e849ce457a4e42784

src/OpenFeature/Providers/MultiProvider/Strategies/FirstMatchStrategy.cs
askpt: Interesting no. All the resolutions are stored until the condition is completed.
Let me try put a nice diagram here!

List:
P1 (Not-Found) -> P2 (Success)

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/828d56413f0e1844d059a4c1c353f951af88af65

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
askpt: This is a valid point. `Value` doesn't have an equality comparer so it will fail. I will need to add it to get this PR working perfectly. Perhaps I will sort this out in another PR since this one is quite busy with multi-provider implementation.

askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/523

arttonoyan (approved):

arttonoyan (changes requested):
Apologies for the premature approval - I need to take a closer look at the changes before this can move forward. I’ll complete my review in the next couple of days

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+
+    /// <inheritdoc/>
+    public override async Task InitializeAsync(EvaluationContext context, CancellationToken cancellationToken = default)
+    {
+        var initializationTasks = this._registeredProviders.Select(async rp =>
+        {
+            try
+            {
+                await rp.Provider.InitializeAsync(context, cancellationToken).ConfigureAwait(false);
+                rp.SetStatus(Constant.ProviderStatus.Ready);
```
weyert: Is this thread safe?

src/OpenFeature/Providers/MultiProvider/ProviderExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+        catch (Exception ex)
+        {
+            // Create an error result
+            var errorResult = new ResolutionDetails<T>(
+                key,
+                defaultValue,
+                ErrorType.General,
+                Reason.Error,
+                errorMessage: ex.Message);
```
weyert, line 41: Do we want to use `ex.ToString()`?

src/OpenFeature/Providers/MultiProvider/Models/ProviderStatus.cs
```diff
@@ -0,0 +1,7 @@
+namespace OpenFeature.Providers.MultiProvider.Models;
+
+internal class ProviderStatus
+{
+    public string ProviderName { get; set; } = string.Empty;
+    public Exception? Exception { get; set; }
```
arttonoyan: I would suggest to use a more descriptive name, such as `Error` or `ExceptionDetail`, to clearly indicate the purpose of the property and avoid shadowing. Reusing the type name as a property name can be confusing.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
arttonoyan: `InitializeAsync` and `ShutdownAsync` iterate through `_registeredProviders` and update each provider’s status concurrently. Each `RegisteredProvider` instance’s `Status` property is mutated inside `Task.WhenAll` without synchronization

Updating shared mutable state in parallel can lead to data races if the `Status` property is not thread‑safe. Either update statuses sequentially, protect them with synchronization (e.g., locks) or redesign `RegisteredProvider` to be immutable and return a new instance with the updated status.

---
To ensure we don’t run into issues here, I would suggest consider adding an **integration test** that validates the concurrent status updates behave correctly. This will help verify thread-safety and prevent potential data races.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+                await rp.Provider.ShutdownAsync(cancellationToken).ConfigureAwait(false);
+                rp.SetStatus(Constant.ProviderStatus.NotReady);
+                return new ProviderStatus { ProviderName = rp.Name };
+            }
+            catch (Exception ex)
+            {
+                rp.SetStatus(Constant.ProviderStatus.Fatal);
+                return new ProviderStatus { ProviderName = rp.Name, Exception = ex };
+            }
+        });
```
arttonoyan: Refer to the synchronization note above.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
arttonoyan: @askpt this was interesting to me so I played around with it a bit.
I’m sharing what I wrote so we don’t lose the code - feel free to improve it and reuse if it’s helpful!

Also, just an idea: it might be useful to add an internal `GetRegisteredProviders()` method in `MultiProvider` to make testing easier. Maybe even a public one if it makes sense longer term.

Here’s the test code I came up with:
```csharp
[Fact]
public async Task MultiProvider_ConcurrentInitializationAndShutdown_ShouldMaintainConsistentProviderStatus()
{
    // Arrange
    const int providerCount = 20;
    var random = new Random();
    var providerEntries = new List<ProviderEntry>();

    for (int i = 0; i < providerCount; i++)
    {
        var provider = Substitute.For<FeatureProvider>();

        provider.InitializeAsync(Arg.Any<EvaluationContext>(), Arg.Any<CancellationToken>())
                .Returns(Task.CompletedTask);

        provider.ShutdownAsync(Arg.Any<CancellationToken>())
                .Returns(Task.CompletedTask);

        provider.GetMetadata()
                .Returns(new Metadata(name: $"provider-{i}"));

        providerEntries.Add(new ProviderEntry(provider));
    }

    var multiProvider = new MultiProvider(providerEntries);

    // Act: simulate concurrent initialization and shutdown with one task each
    var initTasks = Enumerable.Range(0, 1).Select(_ =>
        Task.Run(() => multiProvider.InitializeAsync(Arg.Any<EvaluationContext>(), CancellationToken.None)));

    var shutdownTasks = Enumerable.Range(0, 1).Select(_ =>
        Task.Run(() => multiProvider.ShutdownAsync(CancellationToken.None)));

    await Task.WhenAll(initTasks.Concat(shutdownTasks));

    // Assert: ensure that each provider ends in a valid lifecycle state
    var statuses = GetRegisteredStatuses().ToList();

    Assert.All(statuses, status =>
    {
        Assert.True(
            status is Constant.ProviderStatus.Ready or Constant.ProviderStatus.NotReady,
            $"Unexpected provider status: {status}");
    });

    // Local helper: uses reflection to access the private '_registeredProviders' field 
    // and retrieve the current status of each registered provider.
    // Consider replacing this with an internal or public method if testing becomes more frequent.
    IEnumerable<Constant.ProviderStatus> GetRegisteredStatuses()
    {
        var field = typeof(MultiProvider).GetField("_registeredProviders", BindingFlags.NonPublic | BindingFlags.Instance);
        if (field?.GetValue(multiProvider) is not IEnumerable<object> list)
            throw new InvalidOperationException("Could not retrieve registered providers via reflection.");

        foreach (var p in list)
        {
            var statusProperty = p.GetType().GetProperty("Status", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
            if (statusProperty == null)
                throw new InvalidOperationException($"'Status' property not found on type {p.GetType().Name}.");

            if (statusProperty.GetValue(p) is not Constant.ProviderStatus status)
                throw new InvalidOperationException("Unable to read status property value.");

            yield return status;
        }
    }
}  
```

src/OpenFeature/Providers/MultiProvider/Models/ProviderStatus.cs
askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/dac15c6f34930eae3374ddd619721965765c260b

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/4df6c76dbd2de24a9d5b6507f92a56aff03dc7f2

Please let me know if I overdid the implementation of a thread safety for the initialisation and shutdown. I added a lock to the `ChildProviderStatus` and a semaphore for the lifecycle methods.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Marking as resolved since it will be discussed in the `Initialisation` method.

src/OpenFeature/Providers/MultiProvider/Models/RegisteredProvider.cs
```diff
@@ -2,6 +2,14 @@ namespace OpenFeature.Providers.MultiProvider.Models;
 
 internal class RegisteredProvider
 {
+#if NET9_0_OR_GREATER
+    private readonly Lock _statusLock = new();
+#else
+    private readonly object _statusLock = new object();
+#endif
+
+    private volatile Constant.ProviderStatus _status = Constant.ProviderStatus.NotReady;
```
kylejuliandev: Is `volatile` necessary if we're using a lock object to safeguard access to a single thread?

askpt: Thanks for pointing that out. It is not necessary here and in the `MultiProvider` class. Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/6685f018fb1d4a5f1e9036e03b64f28b3cc39aa7

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,315 @@
...
+    private readonly SemaphoreSlim _initializationSemaphore = new(1, 1);
+    private readonly SemaphoreSlim _shutdownSemaphore = new(1, 1);
+    private ProviderStatus _providerStatus = ProviderStatus.NotReady;
+    private bool _disposed;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MultiProvider"/> class with the specified provider entries and evaluation strategy.
+    /// </summary>
+    /// <param name="providerEntries">A collection of provider entries containing the feature providers and their optional names.</param>
+    /// <param name="evaluationStrategy">The base evaluation strategy to use for determining how to evaluate features across multiple providers.</param>
```
chrfwow: We might want to add here to what strategy the provider will default to if none is given.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,315 @@
...
+public sealed class MultiProvider : FeatureProvider, IDisposable
+{
+    private readonly BaseEvaluationStrategy _evaluationStrategy;
+    private readonly IReadOnlyList<RegisteredProvider> _registeredProviders;
+    private readonly Metadata _metadata;
+
+    private readonly SemaphoreSlim _initializationSemaphore = new(1, 1);
+    private readonly SemaphoreSlim _shutdownSemaphore = new(1, 1);
+    private ProviderStatus _providerStatus = ProviderStatus.NotReady;
+    private bool _disposed;
```
chrfwow: This should probably be `volatile`

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    /// <summary>
+    /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+    /// </summary>
+    public void Dispose()
+    {
+        if (!this._disposed)
+        {
+            this._initializationSemaphore.Dispose();
```
chrfwow, line 281: Is it ok to dispose a semaphore more than once? If no, we need a sync mechanism. I could not find information about that in the [documentation](https://learn.microsoft.com/en-gb/dotnet/api/system.threading.semaphoreslim.dispose?view=net-8.0#system-threading-semaphoreslim-dispose)

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            throw new ObjectDisposedException(nameof(MultiProvider));
+        }
+
+        await this._shutdownSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+        try
+        {
+            if (this._providerStatus != ProviderStatus.Ready || this._disposed)
+            {
+                return;
```
chrfwow: If one of the providers fails during init, the multi provider state will be set to `Fatal`, and we will return here. In this case, we will never shut down the other providers.

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
```diff
@@ -0,0 +1,79 @@
...
+            return new FinalResult<T>(errorDetails, null!, MultiProviderConstants.ProviderName, errors);
+        }
+
+        var firstResult = successfulResolutions.First();
+        ProviderResolutionResult<T>? fallbackResolution = null;
+
+        // Find fallback provider if specified
+        if (this._fallbackProvider != null)
+        {
+            fallbackResolution = successfulResolutions.FirstOrDefault(r => ReferenceEquals(r.Provider, this._fallbackProvider));
```
chrfwow: It seems as if the `fallbackResolution` is only used at the end of this function when the providers disagree. I think it should be moved down there, so we can save this operation if we do have agreement on all providers.

chrfwow (commented):
Nice work

src/OpenFeature/Providers/MultiProvider/ProviderExtensions.cs
askpt: I think in this case just the `Message` would be enough since it is the error message that we are looking for and not the entire stack of the exception.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Not really. Very good catch. The `Dispose` in the `SempahoreSlim` is not thread-safe.

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/036d23d04b5288c3c2aaaf2096bd9813599afdfc

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Valid point! Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/ed784d4da1e6b7ac11aa83eac958adf9d6abbf66

src/OpenFeature/Providers/MultiProvider/Strategies/ComparisonStrategy.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/8d1bb37d3d5dd9ef176f71e460b630fed3f0fec2

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/488/commits/eb98a1b62f912190d090b24094b9f7796b9bf47d

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
askpt: ~Because of the changes on https://github.com/open-feature/dotnet-sdk/pull/488/commits/eb98a1b62f912190d090b24094b9f7796b9bf47d, this no longer needs to be `volatile`.~

Did a bit of a reading and you are right. To make sure the threading is "safe" you need to or mark this a volatile or use `Interlocked.Read` as you mentioned to me in Slack. Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/003e3433f6a850ef9e61e4ddebb8356399c6415d

kylejuliandev (approved):

chrfwow (approved):
lgtm 🚀

askpt:
@arttonoyan Could you please have another look into this PR?

src/OpenFeature/Providers/MultiProvider/MultiProvider.cs
```diff
@@ -0,0 +1,320 @@
...
+    {
+        if (Interlocked.Exchange(ref this._disposed, 1) == 1)
+        {
+            // Already disposed
+            return;
+        }
+
+        this._initializationSemaphore.Dispose();
+        this._shutdownSemaphore.Dispose();
+    }
```
arttonoyan: **Suggestion: Consider using `IAsyncDisposable` to make shutdown and cleanup safer**

Right now, you're using `Dispose()` to clean up the semaphores, which works — but there's a small risk:
If `Dispose()` is called while `InitializeAsync()` or `ShutdownAsync()` is still running (or might run soon), the semaphores could be disposed too early, which can cause issues.

Also, even though you're doing the right thing by not using `DisposeAsync()` directly, there's another edge case:

> `Dispose()` doesn’t guarantee that `ShutdownAsync()` was called — and some users might forget to call it before disposing the provider.

To make this safer and easier for users, you could implement `IAsyncDisposable` and automatically call `ShutdownAsync()` during cleanup.

### Example:

```csharp
public async ValueTask DisposeAsync()
{
    if (Interlocked.Exchange(ref _disposed, 1) == 1)
        return;

    try
    {
        await ShutdownAsync().ConfigureAwait(false);
    }
    finally
    {
        _initializationSemaphore.Dispose();
        _shutdownSemaphore.Dispose();
    }
}
```

And in all public `async` methods:

```csharp
if (_disposed == 1)
    throw new ObjectDisposedException(nameof(MultiProvider));
```

askpt: Great suggestion. I fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/cce748ef5fddfd4ab5d02739fc914fd8eb81c421

askpt: Great suggestion. Fixed with https://github.com/open-feature/dotnet-sdk/pull/488/commits/bfeab2859d123e1e4d080015fd0bfe5bcc7926d2

kylejuliandev (approved):

weyert (approved):

arttonoyan (approved):

test/OpenFeature.Tests/Providers/MultiProvider/MultiProviderTests.cs
```diff
@@ -0,0 +1,851 @@
...
+        // Act
+        await multiProvider.ShutdownAsync();
+
+        // Assert
+        await this._mockProvider1.Received(1).ShutdownAsync(Arg.Any<CancellationToken>());
+        await this._mockProvider2.Received(1).ShutdownAsync(Arg.Any<CancellationToken>());
+    }
+
+    [Fact]
+    public async Task ShutdownAsync_WithFatalProvider_ShutsDownAllProviders()
```
toddbaert, line 222: I think shutting down all providers on a single FATAL makes perfect sense - is this actually specified or should it be added? I can't see anything in the spec about it.

toddbaert (approved):

test/OpenFeature.Tests/Providers/MultiProvider/MultiProviderTests.cs
askpt: I don't think it is specified anywhere, I just followed the logic of, if the MP is `Fatal`, the child providers should be shutdown to save resources.

Also discussed here: https://github.com/open-feature/dotnet-sdk/pull/488#discussion_r2245584816
The only mention in the specification is here: https://openfeature.dev/specification/appendix-a#multi-provider-status

toddbaert: I think we should probably add this to the spec - but it's not a blocker for this PR.

PR #490, opened 2025-06-13
feat: Support retrieving numeric metadata as either integers or decimals 

### LINKED ISSUE #443: [FEATURE] Support retreiving numeric metadata as either either integers or decimals
### Requirements

The [metadata class](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Model/ImmutableMetadata.cs) is a simple abstraciton that holds flag metadata. It can store bool/string/numeric values. These are often deserialized from JSON or some other encoding. For that reason, **we should support the ability to interpret numeric values as either decimals or whole integers** (currently we either store them as doubles or ints and retrieving them the wrong way results in null values).

Link to spec for metadata: https://openfeature.dev/specification/types#flag-metadata

We do something similar to [evaluation context values](https://github.com/open-feature/dotnet-sdk/blob/456351216ce9113d84b56d0bce1dad39430a26cd/src/OpenFeature/Model/Value.cs#L146).

lager95:
## This PR

- adds this new feature

### Related Issues

Fixes #443


src/OpenFeature/Model/ImmutableMetadata.cs
```diff
@@ -45,7 +45,13 @@ public ImmutableMetadata(Dictionary<string, object> metadata)
     /// <returns>The integer value associated with the key, or null if the key is not found.</returns>
```
kinyoklion: The behavior will need to be updated in the document comments for both the methods.

src/OpenFeature/Model/ImmutableMetadata.cs
```diff
@@ -45,7 +45,13 @@ public ImmutableMetadata(Dictionary<string, object> metadata)
...
     public int? GetInt(string key)
     {
-        return this.GetValue<int>(key);
+        var hasValue = this._metadata.TryGetValue(key, out var value);
+        if (!hasValue)
+        {
+            return null;
+        }
+
+        return value is double || value is int ? Convert.ToInt32(value) : null;
```
arttonoyan, line 54: Just a quick thought - do we want to allow implicit conversion here (e.g., from `double`) using `Convert.ToInt32`? Since it can silently round or truncate, I’m curious if that’s acceptable for our use case.

Would it make sense to add an explicit check like this to avoid unexpected rounding?

```csharp
if (value is double d && d % 1 == 0)
    return Convert.ToInt32(d);
```

chrfwow: I tend to agree. Otherwise we might see values that do not correspond to any variant. However, I am not sure if the implementation you suggested couldn't lead to decimal precision errors

src/OpenFeature/Model/ImmutableMetadata.cs
```diff
@@ -55,7 +61,13 @@ public ImmutableMetadata(Dictionary<string, object> metadata)
...
     public double? GetDouble(string key)
     {
-        return this.GetValue<double>(key);
+        var hasValue = this._metadata.TryGetValue(key, out var value);
+        if (!hasValue)
+        {
+            return null;
+        }
+
+        return value is double || value is int ? Convert.ToDouble(value) : null;
```
chrfwow, line 70: Could it be possible that we get different types from the underlying implementation (e.g. `long`, `float`, ...)?

src/OpenFeature/Model/ImmutableMetadata.cs
arttonoyan: @chrfwow You're right - the example was more to illustrate the idea rather than a final implementation. I believe it should work, but we definitely need a unit test to validate the behavior and catch any potential edge cases. Alternatively, we might explore a more robust approach if there's a better algorithm for handling this conversion safely.

aepfli: This is actually somewhat interesting, because, eg, in Java we try to cast, and you can't cast a float to an int. Maybe we should put a little bit more effort into specing this behavior - @open-feature/technical-steering-committee wdyt?

//Edit: ran some tests in java, and we can not use it in a different type, the cast will simply fail. 
<details>

<summary>unit test code java</summary>

### code

```java
    void TestingIntAndLongAndDoubleAndFloatCasts() {
        ImmutableMetadata i1 =
                ImmutableMetadata.builder()
                        .addInteger("int", 1)
                        .addDouble("double", 1.1d)
                        .addDouble("doubleZero", 1.0d)
                        .addFloat("float", 1.2f)
                        .addFloat("floatZero", 1.0f)
                        .addLong("long", 1L)
                        .build();

        SoftAssertions softly = new SoftAssertions();

        softly.assertThat(i1.getInteger("int")).as("integer to integer").isEqualTo(1);
        softly.assertThat(i1.getInteger("double")).as("double to integer").isEqualTo(1);
        softly.assertThat(i1.getInteger("doubleZero")).as("doubleZero to integer").isEqualTo(1);
        softly.assertThat(i1.getInteger("float")).as("float to integer").isEqualTo(1);
        softly.assertThat(i1.getInteger("long")).as("long to integer").isEqualTo(1);


        softly.assertThat(i1.getDouble("int")).as("integer to double").isEqualTo(1);
        softly.assertThat(i1.getDouble("double")).as("double to double").isEqualTo(1.1);
        softly.assertThat(i1.getDouble("doubleZero")).as("doubleZero to double").isEqualTo(1);
        softly.assertThat(i1.getDouble("float")).as("float to double").isEqualTo(1.2);
        softly.assertThat(i1.getDouble("floatZero")).as("float to double").isEqualTo(1);
        softly.assertThat(i1.getDouble("long")).as("long to double").isEqualTo(1);
        softly.assertAll();
    }
```
### result

```
Multiple Failures (8 failures)
-- failure 1 --
[double to integer] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:37)
-- failure 2 --
[doubleZero to integer] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:38)
-- failure 3 --
[float to integer] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:39)
-- failure 4 --
[long to integer] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:40)
-- failure 5 --
[integer to double] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:43)
-- failure 6 --
[float to double] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:46)
-- failure 7 --
[float to double] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:47)
-- failure 8 --
[long to double] 
Expecting actual not to be null
at ImmutableMetadataTest.TestingIntAndLongAndDoubleAndFloatCasts(ImmutableMetadataTest.java:48)
```
</details>

src/OpenFeature/Model/ImmutableMetadata.cs
chrfwow: nvm, this should be fine

chrfwow:
It seems like you added a duplicate `either` to the PR title by accident

PR #508, opened 2025-07-02
fix: ArgumentNullException when creating a client with optional name

### LINKED ISSUE #491: [BUG] ArgumentNullException thrown when adding a Handler to an unnamed `IFeatureClient`.
### Observed behavior

When attempting to `AddHandler` on an `IFeatureClient` that does not have the name metadata, a generic `ArgumentNullException` exception is thrown.
> Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at OpenFeature.EventExecutor.AddClientHandler(String client, ProviderEventTypes eventType, EventHandlerDelegate handler)
   at OpenFeature.Api.AddClientHandler(String client, ProviderEventTypes eventType, EventHandlerDelegate handler)
   at OpenFeature.FeatureClient.AddHandler(ProviderEventTypes eventType, EventHandlerDelegate handler)

From ignoring the null warning here:
https://github.com/open-feature/dotnet-sdk/blob/e18ad50e3298cb0dd19143678c3ef0fdcb4484d9/src/OpenFeature/OpenFeatureClient.cs#L102

And then expected to be set here:
https://github.com/open-feature/dotnet-sdk/blob/e18ad50e3298cb0dd19143678c3ef0fdcb4484d9/src/OpenFeature/EventExecutor.cs#L59

### Expected Behavior

Either provide a better error around requiring the name metadata or allow Handlers to be added to an unnamed FeatureClient (ideal when using dependency injection).

As a workaround, I am able to inject the `Api` instance and call `GetClient(name)` to get a named client.

### Steps to reproduce

* Inject the registered `IFeatureClient` (unnamed)
* Call `AddHandler`

kylejuliandev:
## This PR

- Adds a default placeholder name that gets generated when a Client is generated. This is inspired from the Java SDK (see [EventSource.java](https://github.com/open-feature/java-sdk/blob/0515ad54c4f71863373eb1b7f429393923b27d90/src/main/java/dev/openfeature/sdk/EventSupport.java#L40)). The exception is thrown when an optional client name is specified when retrieving a client from `Api.Instance`. When no name is specified we use a generated one.

### Related Issues

Fixes #491

### DIFF
--- a/src/OpenFeature/EventExecutor.cs
+++ b/src/OpenFeature/EventExecutor.cs
@@ -315,7 +315,7 @@ internal sealed partial class EventExecutor : IAsyncDisposable
 
     private static string GetClientName(string client)
     {
-        if (string.IsNullOrEmpty(client))
+        if (string.IsNullOrWhiteSpace(client))
         {
             return _defaultClientName.ToString();
         }
--- a/src/OpenFeature/EventExecutor.cs
+++ b/src/OpenFeature/EventExecutor.cs
@@ -14,6 +14,9 @@ internal sealed partial class EventExecutor : IAsyncDisposable
     private readonly Dictionary<string, FeatureProvider> _namedProviderReferences = [];
     private readonly List<FeatureProvider> _activeSubscriptions = [];
 
+    /// placeholder for anonymous clients
+    private static Guid _defaultClientName = Guid.NewGuid();
+
     private readonly Dictionary<ProviderEventTypes, List<EventHandlerDelegate>> _apiHandlers = [];
     private readonly Dictionary<string, Dictionary<ProviderEventTypes, List<EventHandlerDelegate>>> _clientHandlers = [];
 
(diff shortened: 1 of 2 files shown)


src/OpenFeature/EventExecutor.cs
```diff
@@ -303,6 +313,14 @@ private void ProcessDefaultProviderHandlers(Event e)
         }
     }
 
+    private static string GetClientName(string client)
+    {
+        if (string.IsNullOrEmpty(client))
```
askpt: Should we check for whitespace?


```suggestion
        if (string.IsNullOrWhiteSpace(client))
```

test/OpenFeature.Tests/OpenFeatureClientTests.cs
```diff
@@ -597,6 +597,20 @@ public async Task PassingABlankStringAsTrackingEventName_ThrowsArgumentException
         Assert.Throws<ArgumentException>(() => client.Track(" \n "));
     }
 
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
```
askpt, line 602: In case we should check for whitespace (https://github.com/open-feature/dotnet-sdk/pull/508/files#r2180899969)


```suggestion
    [InlineData("")]
    [InlineData(" ")]
```

toddbaert (approved):

PR #512, opened 2025-07-05
feat: Add Metric Hook Custom Attributes

### LINKED ISSUE #509: [OTel Hook] Add custom-dimensions support to Metrics Hook
Support the addition of adding custom dimensions (attributes) to the `evaluation_success_total` metric. This is done by allowing users to define a function to extract dimensions from flag metadata.

[Here](https://github.com/open-feature/java-sdk-contrib/blob/main/hooks/open-telemetry/README.md#custom-dimensions-attributes-1) is the equivalent Java implementation.

Task:

- [ ] implement functionality
- [ ] add relevant tests
- [ ] add associated documentation

kylejuliandev:
## This PR

Adds new `MetricsHookOptions` and `MetricsHookOptionsBuilder` to  optionally configure custom attributes that can be tagged on the `feature_flag.evaluation_success_total` metric. Example usage:

```csharp
var options = MetricsHookOptions.CreateBuilder()
    .WithCustomDimension("custom_dimension_key", "custom_dimension_value")
    .WithFlagEvaluationMetadata("boolean", s => s.GetBool("boolean"))
    .Build();

OpenFeature.Api.Instance.AddHooks(new MetricsHook(options));
```

Screenshot below shows the AspNetCore sample application tagging the `feature_flag.evaluation_success_total` counter with the specified dimensions.

image

### Related Issues

Fixes #509

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,6 +663,24 @@ namespace OpenFeatureTestApp
 
 After running this example, you should be able to see some metrics being generated into the console.
 
+You can specify custom dimensions on the `feature_flag.evaluation_success_total` metric by providing `MetricsHookOptions` when adding the hook:
+
+```csharp
+var options = MetricsHookOptions.CreateBuilder()
+    .AddCustomDimension("custom_dimension_key", "custom_dimension_value")
+    .Build();
+
+OpenFeature.Api.Instance.AddHooks(new MetricsHook(options));
+```
+
+You can also write your own extraction logic against the Flag metadata by providing a callback to `WithFlagEvaluationMetadata`.
+
+```csharp
+var options = MetricsHookOptions.CreateBuilder()
+    .WithFlagEvaluationMetadata("boolean", s => s.GetBool("boolean"))
+    .Build();
+```
+
 <!-- x-hide-in-docs-start -->
 
 ## ⭐️ Support the project
--- a/README.md
+++ b/README.md
@@ -667,7 +667,7 @@ You can specify custom dimensions on the `feature_flag.evaluation_success_total`
 
 ```csharp
 var options = MetricsHookOptions.CreateBuilder()
-    .AddCustomDimension("custom_dimension_key", "custom_dimension_value")
+    .WithCustomDimension("custom_dimension_key", "custom_dimension_value")
     .Build();
 
 OpenFeature.Api.Instance.AddHooks(new MetricsHook(options));
--- a/src/OpenFeature/Hooks/MetricsHook.cs
+++ b/src/OpenFeature/Hooks/MetricsHook.cs
@@ -24,15 +24,19 @@ public class MetricsHook : Hook
     private readonly Counter<long> _evaluationSuccessCounter;
     private readonly Counter<long> _evaluationErrorCounter;
 
+    private readonly MetricsHookOptions _options;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="MetricsHook"/> class.
     /// </summary>
-    public MetricsHook()
+    /// <param name="options">Optional configuration for the metrics hook.</param>
+    public MetricsHook(MetricsHookOptions? options = null)
     {
         this._evaluationActiveUpDownCounter = Meter.CreateUpDownCounter<long>(MetricsConstants.ActiveCountName, description: MetricsConstants.ActiveDescription);
         this._evaluationRequestCounter = Meter.CreateCounter<long>(MetricsConstants.RequestsTotalName, "{request}", MetricsConstants.RequestsDescription);
         this._evaluationSuccessCounter = Meter.CreateCounter<long>(MetricsConstants.SuccessTotalName, "{impression}", MetricsConstants.SuccessDescription);
         this._evaluationErrorCounter = Meter.CreateCounter<long>(MetricsConstants.ErrorTotalName, description: MetricsConstants.ErrorDescription);
+        this._options = options ?? MetricsHookOptions.Default;
     }
 
     /// <inheritdoc/>
@@ -60,6 +64,11 @@ public class MetricsHook : Hook
             { TelemetryConstants.Provider, context.ProviderMetadata.Name },
             { TelemetryConstants.Reason, details.Reason ?? Reason.Unknown.ToString() }
         };
+
+        foreach (var metadata in this._options.CustomDimensions)
+        {
+            tagList.Add(metadata.Key, metadata.Value);
+        }
 
         this._evaluationSuccessCounter.Add(1, tagList);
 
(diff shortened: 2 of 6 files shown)


test/OpenFeature.Tests/Hooks/MetricsHookTests.cs
```diff
@@ -59,6 +60,75 @@ await metricsHook.AfterAsync(ctx,
...
+    public async Task With_CustomDimensions_After_Test()
+    {
+        // Arrange
+        var metricHookOptions = MetricsHookOptions.CreateBuilder()
+            .WithCustomDimension("custom_dimension_key", "custom_dimension_value")
+            .Build();
+
+        var metricsHook = new MetricsHook(metricHookOptions);
+
+        using var collector = new MetricCollector<long>(metricsHook._evaluationSuccessCounter);
```
askpt, line 74: This seems to be using a different pattern. I think it makes more sense, when you compare with the previous implementation. If we merge this as it is, I would advice to replicate this pattern to the other unit tests.

kylejuliandev: I can do it as part of this PR or on a subsequent PR? I think using the `MetricCollector` approach is clearer and easier test with

askpt: @kylejuliandev Don't change this PR. I'll create some issues to track the work later on 👍

kylejuliandev: I will happily pick it up once this is merged 😄

README.md
```diff
@@ -663,6 +663,24 @@ namespace OpenFeatureTestApp
 
 After running this example, you should be able to see some metrics being generated into the console.
 
+You can specify custom dimensions on the `feature_flag.evaluation_success_total` metric by providing `MetricsHookOptions` when adding the hook:
```
beeme1mr: Why doesn't the custom dimension apply to all metrics?

beeme1mr (commented):
The feature looks great but I had a question about the docs. I also wonder if we should consider supporting something similar for traces.

README.md
askpt: @beeme1mr that was implemented accordingly to the issue (it was a very old one that I only blind-transferred). Do you think it makes sense to expand to the rest of the metrics?

askpt:
> The feature looks great but I had a question about the docs. I also wonder if we should consider supporting something similar for traces.

@beeme1mr That tracker idea might be something useful. Should we create an issue to track it if makes sense?

README.md
beeme1mr: Yeah, I think it would be a bit confusing to limit it to only one metric.

kylejuliandev: I can tag the remaining metrics too. I might as well do #514 in this change too

askpt: @kylejuliandev feel free since this will be a bigger change than anticipated. Don't forget to link the test changes to this PR! Thank you for your contributions!

src/OpenFeature/Hooks/MetricsHook.cs
```diff
@@ -93,8 +104,32 @@ public override ValueTask FinallyAsync<T>(HookContext<T> context,
...
 
+        this.AddCustomDimensions(ref tagList);
+        this.AddFlagMetadataDimensions(evaluationDetails.FlagMetadata, ref tagList);
+
         this._evaluationActiveUpDownCounter.Add(-1, tagList);
 
         return base.FinallyAsync(context, evaluationDetails, hints, cancellationToken);
     }
+
+    private void AddCustomDimensions(ref TagList tagList)
```
arttonoyan, line 115: The `ref` keyword on `tagList` doesn’t appear necessary here. Since `TagList` is a mutable struct and you're only modifying its internal state (not reassigning it), passing by value is sufficient and avoids the extra complexity `ref` introduces.

You can simplify it to:

```csharp
private void AddCustomDimensions(TagList tagList)
{
    foreach (var customDimension in _options.CustomDimensions)
    {
        tagList.Add(customDimension.Key, customDimension.Value);
    }
}
```

src/OpenFeature/Hooks/MetricsHook.cs
```diff
@@ -93,8 +104,32 @@ public override ValueTask FinallyAsync<T>(HookContext<T> context,
...
+    {
+        foreach (var customDimension in this._options.CustomDimensions)
+        {
+            tagList.Add(customDimension.Key, customDimension.Value);
+        }
+    }
+
+    private void AddFlagMetadataDimensions(ImmutableMetadata? flagMetadata, ref TagList tagList)
```
arttonoyan, line 123: Same here: https://github.com/open-feature/dotnet-sdk/pull/512#discussion_r2193332916

src/OpenFeature/Hooks/MetricsHookOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+        private readonly List<KeyValuePair<string, Func<ImmutableMetadata, object?>>> _flagMetadataExpressions = new List<KeyValuePair<string, Func<ImmutableMetadata, object?>>>();
+
+        /// <summary>
+        /// Adds a custom dimension.
+        /// </summary>
+        /// <param name="key">The key for the custom dimension.</param>
+        /// <param name="value">The value for the custom dimension.</param>
+        public MetricsHookOptionsBuilder WithCustomDimension(string key, object? value)
+        {
+            this._customDimensions.Add(new KeyValuePair<string, object?>(key, value));
```
arttonoyan, line 64: **Minor Suggestion:** You could consider using `KeyValuePair.Create(key, value)` here instead of `new KeyValuePair<string, object?>(...)`. It's a bit more concise and improves readability slightly, though functionally equivalent.

```csharp
_customDimensions.Add(KeyValuePair.Create(key, value));
```

Not a blocking comment - feel free to ignore if you prefer the current style. But if you like this approach, feel free to use it in other parts of the PR as well.

src/OpenFeature/Hooks/MetricsHook.cs
kylejuliandev: `ref` is necessary because any modifications I make to TagList within the AddCustomDimensions method are scoped to that context (as I'm mutating a copy of the value, not a reference)

I could new up a new TagList and return that to the Hook (Before/After/Finally) methods and concatenate the two together - although that creates additional allocations on the stack.

src/OpenFeature/Hooks/MetricsHookOptions.cs
kylejuliandev: `KeyValuePair.Create<>` is not available on net462 or netstandard unfortunately, otherwise I would I prefer the simplicity and readability of this extension method

src/OpenFeature/Hooks/MetricsHook.cs
arttonoyan: Thanks for the clarification - that makes sense.

You're right: since `TagList` is a `readonly struct` and mutating it by value would result in changes being scoped to a copy, the `ref` is indeed necessary **if** the caller expects the internal state to be updated in-place.

Appreciate the explanation. Given the context (e.g., reusing the same `TagList` instance between hooks), this is a valid and efficient choice to avoid unnecessary allocations.

beeme1mr (approved):

PR #513, opened 2025-07-06
feat: Add Hook Dependency Injection extension method with Hook instance

kylejuliandev:
## This PR

In #512 I needed to tweak the sample AspNetCore application to pass Hook options to the MetricsHook. I noticed that in order to pass the options I needed to use the delegate based approach, while discarding the `serviceProvider`, like:

```csharp
featureBuilder.AddHostedFeatureLifecycle()
        .AddHook(_ => new MetricsHook(metricsHookOptions))
```

It would be simpler and easier for devs to interact with a third overload that allows them to pass an instance of the hook that they want to interact with, like so:

```csharp
featureBuilder.AddHostedFeatureLifecycle()
    .AddHook(new MetricsHook())
```

### DIFF
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -302,11 +302,7 @@ public static partial class OpenFeatureBuilderExtensions
     public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, string hookName, THook hook)
         where THook : Hook
     {
-        builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHookName(hookName));
-
-        builder.Services.TryAddKeyedSingleton<Hook>(hookName, hook);
-
-        return builder;
+        return builder.AddHook(hookName, _ => hook);
     }
 
     /// <summary>
--- a/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
@@ -279,6 +279,37 @@ public static partial class OpenFeatureBuilderExtensions
     }
 
     /// <summary>
+    /// Adds a feature hook to the service collection. Hooks added here are not domain-bound.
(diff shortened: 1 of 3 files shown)


askpt (approved):

askpt (approved):

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
```diff
@@ -278,6 +278,37 @@ public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder,
...
+    /// <returns>The <see cref="OpenFeatureBuilder"/> instance.</returns>
+    public static OpenFeatureBuilder AddHook<THook>(this OpenFeatureBuilder builder, string hookName, THook hook)
+        where THook : Hook
+    {
+        builder.Services.PostConfigure<OpenFeatureOptions>(options => options.AddHookName(hookName));
+
+        builder.Services.TryAddKeyedSingleton<Hook>(hookName, hook);
+
+        return builder;
+    }
```
arttonoyan: We already have an `AddHook` overload that handles hook registration with a factory. To avoid duplicating logic, I suggest reusing that method or extracting the shared registration logic into a private helper.

> Ideally, we should centralize the hook registration logic in one place - whether by invoking the existing method or extracting the common logic into a private method that both overloads can use.

Duplicating the same logic across methods increases maintenance cost and the risk of inconsistent behavior. If we ever need to change how hooks are registered (e.g., to add logging or metrics), a single implementation path makes that much easier and safer.

kylejuliandev: Good spot! I've chained the new extension method with a discard on the `implementationFactory` overload.

beeme1mr (approved):

src/OpenFeature.DependencyInjection/OpenFeatureBuilderExtensions.cs
kylejuliandev: @arttonoyan are the latest changes ok? Happy to adjust as needed!

arttonoyan (approved):

PR #515, opened 2025-07-07
refactor: Simplify Provider Repository

askpt:
## This PR

This pull request modifies the `ProviderRepository` class in the `src/OpenFeature/ProviderRepository.cs` file to improve cancellation support, streamline initialization and shutdown processes, and enhance compatibility across different .NET versions. The most notable changes include adding `CancellationToken` parameters to several methods, updating method access modifiers, and simplifying object initializations.

### Cancellation Support Enhancements:
* Added `CancellationToken` parameters to methods such as `SetProviderAsync`, `InitProviderAsync`, `ShutdownIfUnusedAsync`, and `SafeShutdownProviderAsync` to allow for operation cancellation. [[1]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL56-R60) [[2]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL82-R98) [[3]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL107-R108) [[4]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL169-R192) [[5]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL195-R204) [[6]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL207-R216)

### Method Access Modifier Updates:
* Changed several methods from `public` to `internal`, including `SetProviderAsync`, `GetProvider`, and `ShutdownAsync`, to restrict their accessibility and better align with the class's intended usage. [[1]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL56-R60) [[2]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL137-R157) [[3]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL237-R255) [[4]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL257-R266)

### Compatibility Improvements:
* Added conditional logic to handle `string.IsNullOrWhiteSpace` differently in .NET Framework and .NET Standard due to nullable compatibility issues. [[1]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL137-R157) [[2]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL237-R255)

### Code Simplifications:
* Simplified object initializations by using shorthand syntax, such as replacing explicit constructor calls with `{}` for `ConcurrentDictionary` and `ReaderWriterLockSlim`. [[1]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL7-R18) [[2]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bL32-R30)

### DIFF
--- a/src/OpenFeature/ProviderRepository.cs
+++ b/src/OpenFeature/ProviderRepository.cs
@@ -135,7 +135,7 @@ internal sealed partial class ProviderRepository : IAsyncDisposable
     /// initialization
     /// </param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to cancel any async side effects.</param>
-    internal async Task SetProviderAsync(string? domain,
+    internal async Task SetProviderAsync(string domain,
         FeatureProvider? featureProvider,
         EvaluationContext context,
         Func<FeatureProvider, Task>? afterInitSuccess = null,
@@ -143,18 +143,10 @@ internal sealed partial class ProviderRepository : IAsyncDisposable
         CancellationToken cancellationToken = default)
     {
         // Cannot set a provider for a null domain.
-#if NETFRAMEWORK || NETSTANDARD
-        // This is a workaround for the issue in .NET Framework where string.IsNullOrEmpty is not nullable compatible.
-        if (domain == null)
-        {
-            return;
-        }
-#else
         if (string.IsNullOrWhiteSpace(domain))
         {
             return;
         }
-#endif
 
         this._providersLock.EnterWriteLock();
 
--- a/src/OpenFeature/ProviderRepository.cs
+++ b/src/OpenFeature/ProviderRepository.cs
@@ -4,7 +4,6 @@ using Microsoft.Extensions.Logging.Abstractions;
 using OpenFeature.Constant;
 using OpenFeature.Model;
 
-
 namespace OpenFeature;
 
 /// <summary>
@@ -16,8 +15,7 @@ internal sealed partial class ProviderRepository : IAsyncDisposable
 
     private FeatureProvider _defaultProvider = new NoOpFeatureProvider();
 
-    private readonly ConcurrentDictionary<string, FeatureProvider> _featureProviders =
-        new ConcurrentDictionary<string, FeatureProvider>();
+    private readonly ConcurrentDictionary<string, FeatureProvider> _featureProviders = new();
 
     /// The reader/writer locks is not disposed because the singleton instance should never be disposed.
     ///
@@ -29,7 +27,7 @@ internal sealed partial class ProviderRepository : IAsyncDisposable
     /// The second is that a concurrent collection doesn't provide any ordering, so we could check a provider
     /// as it was being added or removed such as two concurrent calls to SetProvider replacing multiple instances
     /// of that provider under different names.
-    private readonly ReaderWriterLockSlim _providersLock = new ReaderWriterLockSlim();
+    private readonly ReaderWriterLockSlim _providersLock = new();
 
     public async ValueTask DisposeAsync()
     {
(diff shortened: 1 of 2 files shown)


src/OpenFeature/ProviderRepository.cs
```diff
@@ -134,18 +135,26 @@ private static async Task InitProviderAsync(
...
-    public async Task SetProviderAsync(string? domain,
+    internal async Task SetProviderAsync(string? domain,
         FeatureProvider? featureProvider,
         EvaluationContext context,
         Func<FeatureProvider, Task>? afterInitSuccess = null,
         Func<FeatureProvider, Exception, Task>? afterInitError = null,
         CancellationToken cancellationToken = default)
     {
         // Cannot set a provider for a null domain.
+#if NETFRAMEWORK || NETSTANDARD
```
kinyoklion: We should likely make these functionally the same as they are different this way.  And also update comment if we also want to disallow empty/white space.

src/OpenFeature/ProviderRepository.cs
```diff
@@ -234,16 +243,16 @@ public FeatureProvider GetProvider()
         }
     }
 
-    public FeatureProvider GetProvider(string? domain)
+    internal FeatureProvider GetProvider(string? domain)
     {
-#if NET6_0_OR_GREATER
-        if (string.IsNullOrEmpty(domain))
+#if NETFRAMEWORK || NETSTANDARD
```
kinyoklion, line 240: Once we know the logic we want, per other comment, we should de-duplicate it.

kinyoklion (changes requested):

src/OpenFeature/ProviderRepository.cs
askpt: Actually, did a better simplification here. The only place where `SetProviderAsync` with a domain is called is in the `Api`, which already has a `null` verification. This cleans up the code. Check: https://github.com/open-feature/dotnet-sdk/pull/515/commits/b6d45eaf230deb72a6384b5f982c94602020889c

src/OpenFeature/ProviderRepository.cs
askpt: As per the comment in the public api: https://github.com/open-feature/dotnet-sdk/blob/929fa7497197214d385eeaa40aba008932d00896/src/OpenFeature/Api.cs#L88
I would assume that when the `domain` is null, empty, or whitespace, it should return the default. 

In this case, the `precondition` needs to be present because `IsNullOrEmpty` is not nullable compatible. Thus, it has the same behaviour in the end.

kinyoklion: This was in reference to the ifdefs that were in two places. Just if we needed this, then it should be on one place. Not ifdefs multiple places for the same intent.

I still think comment updates are needed as empty versus whitespace only are not the same.

kylejuliandev (approved):

src/OpenFeature/ProviderRepository.cs
askpt: That is fair! Thanks for pointing out the comments. Fixed with https://github.com/open-feature/dotnet-sdk/pull/515/commits/c071451d54dce670d89815cab8459923085a1d79

arttonoyan (approved):

askpt:
@kinyoklion, can you please have a look at this PR again?

toddbaert:
@kinyoklion I'll merge this in the next 2 days unless I hear objections; it seems @askpt has resolved your suggestions. :pray:

kinyoklion:
Sorry for the delay. Taking a look now.

kinyoklion (approved):

PR #523, opened 2025-07-13
chore: Add comparison to Value

askpt:
Signed-off-by: André Silva <2493377+askpt@users.noreply.github.com>

## This PR

This pull request introduces enhancements to the `Value` class in the `OpenFeature.Model` namespace, ensuring better equality handling, and updates dependencies to include `Microsoft.Bcl.HashCode`. The most significant changes include implementing equality comparison for `Value`, adding hash code generation, and updating project files to include the new dependency.

### Enhancements to `Value` class:

* [`src/OpenFeature/Model/Value.cs`](diffhunk://#diff-336aacd3c42458899187108a2064648ae21e439b2b11e6ca7f25b7b7fef00609L9-R9): The `Value` class now implements `IEquatable<Value>` and includes methods for equality comparison (`Equals`, `==`, `!=`), hash code generation (`GetHashCode`), and internal helpers for comparing complex types like structures and lists. This ensures more robust and consistent equality checks. [[1]](diffhunk://#diff-336aacd3c42458899187108a2064648ae21e439b2b11e6ca7f25b7b7fef00609L9-R9) [[2]](diffhunk://#diff-336aacd3c42458899187108a2064648ae21e439b2b11e6ca7f25b7b7fef00609R187-R378)

### Dependency updates:

* [`Directory.Packages.props`](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156L13-R25): Added `Microsoft.Bcl.HashCode` as a dependency for hash code generation. Other package references were reformatted for consistency.
* [`src/OpenFeature/OpenFeature.csproj`](diffhunk://#diff-711ea17cbdebe419375c7684c8c39a1423d2bebcf8976ddd7bdd78deaab65b21R11): Included `Microsoft.Bcl.HashCode` for specific target frameworks (`net462` and `netstandard2.0`).

### Notes

This implementation is necessary for the comparison in the MultiProvider. See: https://github.com/open-feature/dotnet-sdk/pull/488#discussion_r2201848459

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,13 +10,19 @@
 
   <ItemGroup Label="src">
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions"
+      Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions"
+      Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions"
+      Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection"
+      Version="$(MicrosoftExtensionsVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsVersion)" />
     <PackageVersion Include="System.Collections.Immutable" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="System.Diagnostics.DiagnosticSource"
+      Version="$(MicrosoftExtensionsVersion)" />
     <PackageVersion Include="System.Threading.Channels" Version="$(MicrosoftExtensionsVersion)" />
     <PackageVersion Include="System.ValueTuple" Version="4.6.1" />
   </ItemGroup>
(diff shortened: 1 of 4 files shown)


Directory.Packages.props
```diff
@@ -10,13 +10,19 @@
 
   <ItemGroup Label="src">
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageVersion Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
```
kylejuliandev, line 13: What is Bcl.HashCode used for/in?

askpt, line 13: Great question. It is used to generate the `HashCode` for the `IEquatable` override. Check the documentation here: https://learn.microsoft.com/en-us/dotnet/api/system.hashcode?view=net-9.0

kylejuliandev, line 13: How come you're using v1.1.1 and v6.0.0?

src/OpenFeature/Model/Value.cs
```diff
@@ -184,4 +184,196 @@ public Value(Object value)
...
+
+    /// <summary>
+    /// Returns the hash code for this <see cref="Value"/>.
+    /// </summary>
+    /// <returns>A hash code for the current <see cref="Value"/>.</returns>
+    public override int GetHashCode()
+    {
+        if (this.IsNull) return 0;
+
+        return this.GetValueType() switch
```
kylejuliandev, line 235: Should a non-value type still get a unique hash code?

Directory.Packages.props
askpt: Sorry! Didn't get the question

src/OpenFeature/Model/Value.cs
```diff
@@ -184,4 +184,196 @@ public Value(Object value)
...
+    private bool ListEquals(Value other)
+    {
+        var thisList = this.AsList!;
+        var otherList = other.AsList!;
+
+        if (thisList.Count != otherList.Count) return false;
+
+        for (int i = 0; i < thisList.Count; i++)
+        {
+            if (!thisList[i].Equals(otherList[i]))
```
kylejuliandev, line 321: Would two lists be considered equal if the order was not the same?

src/OpenFeature/Model/Value.cs
askpt: In this case, it shouldn't. Since we are out of the _accepted_ value types, we are in error territory (something is quite wrong). I considered throwing an error here, but zero is the best option.

src/OpenFeature/Model/Value.cs
askpt: Not really. The list considers the order of the elements. It is not equal if the list has the elements in the wrong order. Order is not essential for the structure (the same as the dictionary), so you need to search by element instead.

kylejuliandev (approved):

Directory.Packages.props
kylejuliandev: Resolved as discussed in Slack. Targeting the earliest compatible version is the most ideal situation for sdk consumers

beeme1mr (approved):

PR #526, opened 2025-07-19
feat: Add TraceEnricherHookOptions Custom Attributes

### LINKED ISSUE #516: Track and add custom attributes for Traces similar to Metric Hook
### Requirements

We should implement a feature to add and track custom attributes for Traces, similar to what was done for Metric Hook in PR #512. This would help enhance trace observability and flexibility in the SDK.

- Review the implementation of custom attributes in the Metric Hook (see PR #512)
- Propose a similar approach for Traces
- Ensure parity in attribute handling between Traces and Metrics
- Document usage and provide examples

This issue will track progress and discussions for this enhancement.

kylejuliandev:
## This PR

Adds `TraceEnricherHookOptions`, largely a copy of `MetricHookOptions` from #512, to enrich the `feature_flag.evaluation` span event.

Example usage:

```csharp
var traceHookOptions = TraceEnricherHookOptions.CreateBuilder()
    .WithCustomDimension("key1", "value1")
    .WithFlagEvaluationMetadata("provider", s => s.GetString("provider"))
    .Build();

featureBuilder.AddHostedFeatureLifecycle()
    .AddHook(new TraceEnricherHook(traceHookOptions));
```

These adds tags to the event associated with the span. Below is an example with the above custom dimension and metadata callback:

image

### Related Issues

Fixes #516

### Follow-up Tasks

I kept the builder method names `WithCustomDimension` and `WithFlagEvaluationMetadata`, maybe this should be different for Traces? In the Java implementation the both hook options look very similar: [TracesHookOptions.java](https://github.com/open-feature/java-sdk-contrib/blob/main/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/TracesHookOptions.java) and [MetricHookOptions.java](https://github.com/open-feature/java-sdk-contrib/blob/main/hooks/open-telemetry/src/main/java/dev/openfeature/contrib/hooks/otel/MetricHookOptions.java)

### DIFF
--- a/src/OpenFeature/Hooks/TraceEnricherHookOptions.cs
+++ b/src/OpenFeature/Hooks/TraceEnricherHookOptions.cs
@@ -13,9 +13,9 @@ public sealed class TraceEnricherHookOptions
     public static TraceEnricherHookOptions Default { get; } = new TraceEnricherHookOptions();
 
     /// <summary>
-    /// Custom dimensions or tags to be associated with current <see cref="System.Diagnostics.Activity"/> in <see cref="TraceEnricherHook"/>.
+    /// Custom tags to be associated with current <see cref="System.Diagnostics.Activity"/> in <see cref="TraceEnricherHook"/>.
     /// </summary>
-    public IReadOnlyCollection<KeyValuePair<string, object?>> CustomDimensions { get; }
+    public IReadOnlyCollection<KeyValuePair<string, object?>> Tags { get; }
 
     /// <summary>
     /// Flag metadata callbacks to be associated with current <see cref="System.Diagnostics.Activity"/>.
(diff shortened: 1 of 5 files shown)


src/OpenFeature/Hooks/TraceEnricherHookOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// The default options for the <see cref="TraceEnricherHookOptions"/>.
+    /// </summary>
+    public static TraceEnricherHookOptions Default { get; } = new TraceEnricherHookOptions();
+
+    /// <summary>
+    /// Custom dimensions or tags to be associated with current <see cref="System.Diagnostics.Activity"/> in <see cref="TraceEnricherHook"/>.
+    /// </summary>
+    public IReadOnlyCollection<KeyValuePair<string, object?>> CustomDimensions { get; }
```
WeihanLi: What about using `Tags` since it's actually refer to the tags?

src/OpenFeature/Hooks/TraceEnricherHookOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        private readonly List<KeyValuePair<string, object?>> _customDimensions = new List<KeyValuePair<string, object?>>();
+        private readonly List<KeyValuePair<string, Func<ImmutableMetadata, object?>>> _flagMetadataExpressions = new List<KeyValuePair<string, Func<ImmutableMetadata, object?>>>();
+
+        /// <summary>
+        /// Adds a custom dimension.
+        /// </summary>
+        /// <param name="key">The key for the custom dimension.</param>
+        /// <param name="value">The value for the custom dimension.</param>
+        public TraceEnricherHookOptionsBuilder WithCustomDimension(string key, object? value)
```
WeihanLi: What about naming it to `WithTag` since it's going to add tag for the activity

```suggestion
        public TraceEnricherHookOptionsBuilder WithTag(string key, object? value)
```

kylejuliandev: Good idea! It is probably best to use consistent terms when working with Activities versus Meters

src/OpenFeature/Hooks/TraceEnricherHookOptions.cs
kylejuliandev: Will update this later today!

askpt (approved):

beeme1mr (approved):

Issue #424, opened 2025-04-10
[BUG] Change dependency assets


askpt:
### Observed behavior

While trying to use this package, I am getting this error:
 ```
error NU1605: 
      Warning As Error: Detected package downgrade: Microsoft.Extensions.Logging.Abstractions from 9.0.3 to 9.0.0. Reference the package directly from the project to select a different version. 
       MyWebsite -> OpenFeature 2.3.2 -> Microsoft.Extensions.Logging.Abstractions (>= 9.0.3)
```

### Expected Behavior

The libraries are not exposed.

### How to replicate
1 - Create a new web project
2 - Reference `Microsoft.Extensions.Logging.Abstractions` version 9.0.2 or lower
3 - Add OpenFeature version 2.3.2
4 - Compile

Check attachment: [WebApplication1.zip](https://github.com/user-attachments/files/19688985/WebApplication1.zip)

PR #529, opened 2025-07-22
feat: Support JSON Serialize for Value

### LINKED ISSUE #517: Best practices to use JSON value
We're integrating the LaunchDarkly based on OpenFeature SDK, it did not goes well when we tried to integrate JSON values from LaunchDarkly, are there some best practices about using JSON values?

With my limited research, we should use the `GetObject` API for getting JSON values, while the value we got is a `Structure` object, how should we convert it strong typed object performantly?

```cs
var result = await _featureClient.GetObjectDetailsAsync(
            flagKey, default, evaluationContext, cancellationToken: cancellationToken);
var obj = result.Value.AsObject;
```

There's a `ToJsonString()` method for the `LDValue`, and implemented a customized JsonConverter for the `LdValue` in the LaunchDarkly SDK, maybe something similar would be needed for the `Value` in the open feature SDK also?

WeihanLi:
## This PR

- add `ValueJsonConverter` to support STJ JSON Serialize

### Related Issues

Fixes #517 

### Notes

- Add `ValueJsonConverter` to support STJ JSON Serialize, keep it `internal` for now, think we could make it `public` when necessary

### Follow-up Tasks

- Do we need to add a `ToJsonString()` method likes LD SDK?
- Do we need to add a method to convert `Value` to a strong-typed `T`?

### How to test

refer to the unit test cases

### DIFF
--- a/src/OpenFeature/Model/ValueJsonConverter.cs
+++ b/src/OpenFeature/Model/ValueJsonConverter.cs
@@ -28,16 +28,7 @@ internal sealed class ValueJsonConverter : JsonConverter<Value>
 
         if (value.IsNumber)
         {
-            var doubleVal = value.AsDouble;
-            Debug.Assert(doubleVal is not null);
-            if (((doubleVal % 1) == 0) && doubleVal!.Value is >= int.MinValue and <= int.MaxValue)
-            {
-                writer.WriteNumberValue(value.AsInteger!.Value);
-            }
-            else
-            {
-                writer.WriteNumberValue(doubleVal!.Value);
-            }
+            writer.WriteNumberValue(value.AsDouble!.Value);
             return;
         }
 


Directory.Packages.props
```diff
@@ -23,6 +23,8 @@
     <PackageVersion Include="System.Collections.Immutable" Version="$(MicrosoftExtensionsVersion)" />
     <PackageVersion Include="System.Diagnostics.DiagnosticSource"
       Version="$(MicrosoftExtensionsVersion)" />
+      <PackageVersion Include="System.Text.Json"
+      Version="8.0.5" />
```
kylejuliandev, line 27: I think we may need to target the earliest supported version that works for our use case (and for the tfms we support), otherwise we may cause dependency issues for consumers

See #424

If v8.0.5 is the earliest version then feel free to ignore this comment 😄

src/OpenFeature/Model/ValueJsonConverter.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        if (value.IsString)
+        {
+            writer.WriteStringValue(value.AsString);
+            return;
+        }
+
+        if (value.IsDateTime)
+        {
+            writer.WriteStringValue(value.AsDateTime!.Value);
```
kylejuliandev, line 43: What format or format provider does STJ default to when handling DateTimes? Is it the round trip or `ISO 8601` one?

src/OpenFeature/Model/ValueJsonConverter.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            writer.WriteBooleanValue(value.AsBoolean.GetValueOrDefault());
+            return;
+        }
+
+        if (value.IsNumber)
+        {
+            var intVal = value.AsInteger;
+            var doubleVal = value.AsDouble;
+            if (intVal.HasValue && Convert.ToDouble(intVal.Value) == doubleVal.GetValueOrDefault())
```
kylejuliandev: If we create `Value` as a double, won't AsInteger always be not null

https://github.com/open-feature/dotnet-sdk/blob/a0ae014d3194fcf6e5e5e4a17a2f92b1df3dc7c7/src/OpenFeature/Model/Value.cs#L144

The underlying Value code converts the number to an integer, which will either succeed (if the double is less than 4 bytes) or fail due to an overflow exception

kylejuliandev: I don't think doing `double == double` equality is a reliable way to determine if the value is actually a floating point number

test/OpenFeature.Tests/StructureTests.cs
```diff
@@ -113,4 +115,45 @@ public void GetEnumerator_Should_Return_Enumerator()
...
+        yield return [new Value(false), "false"];
+        yield return
+        [
+            new Value(Structure.Builder()
+                .Set("name", "Alice")
+                .Set("age", 16)
+                .Set("isMale", false)
+                .Set("bio", new Value())
+                .Set("bornAt", new DateTime(2000, 1, 1))
+                .Set("tags", new Value([new Value("girl"), new Value("beauty")]))
```
kylejuliandev, line 154: One more potential test case might be a nested structure, e.g.

```csharp
.Set("struct", new Value(Structure.Builder().Set("inner", 10)))
```

src/OpenFeature/Model/ValueJsonConverter.cs
WeihanLi: > The System.Text.Json library parses and writes `DateTime` and `DateTimeOffset` values according to the ISO 8601-1:2019 extended profile

From the document, it should be `ISO 8601` one

see details on: https://learn.microsoft.com/en-us/dotnet/standard/datetime/system-text-json-support

test/OpenFeature.Tests/StructureTests.cs
WeihanLi: Nice catch 👍 , will add later

src/OpenFeature/Model/ValueJsonConverter.cs
WeihanLi: Yeah, valid concern. I also think that's not reliable, will update this later

WeihanLi: Thanks

Directory.Packages.props
WeihanLi: Since  `net6.0`/`net7.0` is out of support, try to the 8.x packages and the `8.0.5` is the lowest version I found has no known vulnerabilities

test/OpenFeature.Tests/StructureTests.cs
WeihanLi: added

src/OpenFeature/Model/ValueJsonConverter.cs
WeihanLi: updated

WeihanLi: updated

src/OpenFeature/Model/ValueJsonConverter.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            writer.WriteBooleanValue(value.AsBoolean.GetValueOrDefault());
+            return;
+        }
+
+        if (value.IsNumber)
+        {
+            var doubleVal = value.AsDouble;
+            Debug.Assert(doubleVal is not null);
+            if (((doubleVal % 1) == 0) && doubleVal!.Value is > int.MinValue and < int.MaxValue)
```
WeihanLi:
```suggestion
            if (((doubleVal % 1) == 0) && doubleVal!.Value is >= int.MinValue and <= int.MaxValue)
```

src/OpenFeature/Model/ValueJsonConverter.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            writer.WriteBooleanValue(value.AsBoolean.GetValueOrDefault());
+            return;
+        }
+
+        if (value.IsNumber)
+        {
+            var doubleVal = value.AsDouble;
+            Debug.Assert(doubleVal is not null);
+            if (((doubleVal % 1) == 0) && doubleVal!.Value is >= int.MinValue and <= int.MaxValue)
```
kylejuliandev: I think `1.0` will be written as an integer and not a double, not sure how important it is to maintain the accuracy of this

WeihanLi: Looks as expected when writing as a double number value for an integer. Let me simplify this

> From research, the default `double` format is `G17`
> 
> - https://github.com/dotnet/runtime/blob/3c94a470bc7fbdabafdf5d224b48dc5500e1b3b3/src/libraries/System.Text.Json/Common/JsonConstants.cs#L9
> - https://github.com/dotnet/runtime/blob/3c94a470bc7fbdabafdf5d224b48dc5500e1b3b3/src/libraries/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.WriteValues.Double.cs#L101

PR #530, opened 2025-07-24
chore: Add README and Deprecation notice to DependencyInjection library

### LINKED ISSUE #472: OpenFeature.Hosting doesn't target netstandard or framework
The `OpenFeature.Hosting` project doesn't target the same frameworks as `OpenFeature.DependencyInjection`, which means users in .NET Framework cannot easily get started with the dependency injection initialization approach.

https://github.com/open-feature/dotnet-sdk/blob/48f57cd40c2754bbbb08202f338e2452522fe691/src/OpenFeature.Hosting/OpenFeature.Hosting.csproj#L3-L8

`OpenFeature.Hosting` only depends on [Microsoft.Extensions.Hosting.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions/9.0.5) which does have support for .NET Framework 4.6.2 and .NET Standard 2.0.

kylejuliandev:
## This PR

This adds a README to the Dependency Injection NuGet package with a deprecation notice or warning.

### Related Issues

Fixes #472

### DIFF
--- a/src/OpenFeature.DependencyInjection/README.md
+++ b/src/OpenFeature.DependencyInjection/README.md
@@ -1,0 +1,48 @@
+# OpenFeature.DependencyInjection
+
+> **⚠️ DEPRECATED**: This library is now deprecated. The OpenTelemetry Dependency Injection library has been moved to the OpenFeature Hosting integration in version 2.8.0.
+
+OpenFeature is an open standard for feature flag management, created to support a robust feature flag ecosystem using cloud native technologies. OpenFeature will provide a unified API and SDK, and a developer-first, cloud-native implementation, with extensibility for open source and commercial offerings.
+
+## Migration Guide
+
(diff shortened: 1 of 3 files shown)


src/OpenFeature.DependencyInjection/README.md
```diff
@@ -0,0 +1,48 @@
+# OpenFeature.DependencyInjection
+
+> **⚠️ DEPRECATED**: This library is now deprecated. The OpenTelemetry Dependency Injection library has been moved to the OpenFeature Hosting integration in version 2.8.0.
```
kylejuliandev: I think we need to release one more version of the Dependency Injection library, I assume in 2.8.0?

askpt: @kylejuliandev 2.8.0 will be the last version that will publish the DependencyInjection package. It should have the [Obsolete] all over the place and in 2.9.0 we remove the package when we deploy it again.

benjiro (approved):
LGTM 🏆

askpt (approved):

PR #531, opened 2025-07-24
feat: Deprecate AddHostedFeatureLifecycle method

### LINKED ISSUE #472: OpenFeature.Hosting doesn't target netstandard or framework

kylejuliandev:
## This PR

- Copy all Dependency Injection files over to the Hosting package
- Add [Obsolete()] attribute to AddHostedFeatureLifecycle()
- Ensure that the Hosted Service is injected when you make a call to `AddOpenFeature`
- Add new test project with the same unit tests from the Dependency Injection tests

This is technically a breaking change because anyone using both packages in their project, which everyone will be, they will run into issues around resolving the correct extension methods.

### Related Issues

Fixes #472


askpt:
> @kylejuliandev I suggest moving all files currently in the Hosting package (under the `Hosting` namespace) into the DependencyInjection package (also under `Hosting` namespace).

I am more inclined to move all `DI` into the `Hosting` namespace. We have `Hosting` specific stuff and `DI` in the same package. We rarely reference the DI package, and it usually comes included in the `Hosting` one.

The dependency tree generally looks like:
```
YourApp
 └─ OpenFeature.Hosting
      └─ OpenFeature.DependencyInjection
```
      
We should keep only the DI unless we provide only DI extensions. Since we also offer application configuration, I vote to keep the `Hosting` and merge the `DI` into that one.

We might need to make changes to the configuration file since I would like to keep the same namespaces we are using now.

---

> Also, in the **Hosting** package, we currently have `FeatureLifecycleStateOptions`, which were configured using `AddHostedFeatureLifecycle`. Since we're deprecating that, we should remember to update the docs to show that the hosted service can now be configured using the standard Microsoft configuration pattern.
> 
> For example:
> 
> ```cs
> services.ConfigureOptions<FeatureLifecycleStateOptions>(options =>
> {
>     // Optional: configure StartState, StopState, etc.
> });
> ```
> 
> All of these options are optional, but this makes it easier and more consistent with typical .NET configuration.

Agreed!

askpt:
@kylejuliandev the code coverage went quite down. You did a copy-paste of the current code without changes?

askpt (approved):

src/OpenFeature.Hosting/OpenFeatureBuilderExtensions.cs
```diff
@@ -9,6 +13,370 @@ namespace OpenFeature;
...
+        {
+            builder.Services.Configure(domain, configureOptions);
+        }
+
+        builder.Services.TryAddKeyedTransient(domain, (provider, key) =>
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
```
WeihanLi, line 131: Should we prefer the `Guard.ThrowIfNull(key)` method here?

askpt:
> As a follow up, we'll need to update the readme with the updated config. I believe @askpt updated the OTel demo to use DI, so we should help with with the migration once this has been released.
> 
> I'd like like to wait for @arttonoyan approval since he was the original author of the DI feature. Thanks!

Thanks for pointing out the main `README.md` update @beeme1mr! We should make sure to update to the latest version that doesn't include the `AddHosted` method.

When the next version is released, I will update the Otel-demo repository 👍

src/OpenFeature.Hosting/OpenFeatureBuilderExtensions.cs
kylejuliandev: Possibly a further refactoring we can take on. For this PR I've tried to minimise the changes to moving files (from DependencyInjection to Hosting) and updating the namespaces

PR #534, opened 2025-07-29
fix: update DI lifecycle to use container instead of static instance

### LINKED ISSUE #533: Support Non-Static-Singleton in OpenFeature.Hosting
### Requirements

My org is exploring adopting OpenFeature as our feature flagging solution in one of our greenfield projects. I was able to plug it into one of our apps relatively easily, but ran into a bit of a roadblock to adopting it. We use `WebApplicationFactory<T>` for our REST/gRPC API integration tests as outlined in Microsoft's docs [here](https://learn.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-9.0&pivots=xunit). The integration tests invoke the app's `Program.cs`, which in turn has the DI registration calls:

```csharp
builder.Services.AddOpenFeature(options =>
{
  // Provider registration
});
```

A few of our projects parallelize the integration tests to speed things up and each integration test class gets its own unique DI container so tests don't trample on each other. As I was plugging in OpenFeature I noticed that the test fixtures would pass when run independently but fail if I run the entire suite so I started to do some digging and this line of code is the culprit (Note: This is decompiled code):

```csharp
public static IServiceCollection AddOpenFeature(
    this IServiceCollection services,
    Action<OpenFeatureBuilder> configure)
  {
    Guard.ThrowIfNull((object) services, nameof (services));
    Guard.ThrowIfNull((object) configure, nameof (configure));
    services.TryAddSingleton<Api>(Api.Instance); // <-- This line
    services.TryAddSingleton<IFeatureLifecycleManager, FeatureLifecycleManager>();
    OpenFeatureBuilder builder = new OpenFeatureBuilder(services);
    configure(builder);
    if (builder.HasDefaultProvider && builder.DomainBoundProviderRegistrationCount == 0)
      return services;
    builder.Validate();
    if (!builder.IsPolicyConfigured)
      builder.AddPolicyName((Action<PolicyNameOptions>) (options => options.DefaultNameSelector = (Func<IServiceProvider, string>) (provider => provider.GetRequiredService<IOptions<OpenFeatureOptions>>().Value.ProviderNames.First<string>())));
    builder.AddPolicyBasedClient();
    return services;
  }
```

The issue is that while each DI container is independent, they are relying on this static singleton instance of `Api` which will be the same instance for all tests and when one test tears down its DI container, then the other tests will begin failing due to errors like:

```
ChannelClosedException : The channel has been closed.
```

Are there any plans to make the `OpenFeature.Hosting` library use singletons instantiated through the DI framework rather than static in-memory singletons? Any work arounds or other suggestions? Any help would be much appreciated!

askpt:
## This PR

This pull request refactors the `Api` class and its integration with `FeatureClient` to improve testability and dependency management by replacing the singleton access pattern with dependency injection. It also includes minor project configuration updates. Below are the key changes grouped by theme:

### Refactoring `Api` and Dependency Injection:

* Changed the `Api` constructor from `private` to `internal`, allowing it to be instantiated directly for dependency injection purposes (`src/OpenFeature/Api.cs`, [src/OpenFeature/Api.csL35-R35](diffhunk://#diff-dc150fbd3b7be797470374ee7e38c7ab8a31eb9b6b7a52345e05114c2d32a15eL35-R35)).
* Updated `AddOpenFeature` in `OpenFeatureServiceCollectionExtensions` to register a new instance of `Api` instead of using the singleton `Api.Instance` (`src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs`, [src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.csL27-R27](diffhunk://#diff-cda6bc81ad25c28afd42716109c7c5c2fd644c0b792295d35a83bfd4d8dca6d8L27-R27)).
* Modified `FeatureClient` to accept an `Api` instance as a constructor parameter, defaulting to `Api.Instance` if not provided. Updated all internal references to use the injected `Api` instance (`src/OpenFeature/OpenFeatureClient.cs`, [[1]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907R21) [[2]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907R76-R86) [[3]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907L102-R111) [[4]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907L216-R225) [[5]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907L313-R316).

### Updates to `FeatureClient`:

* Added a private `_api` field to store the injected `Api` instance and replaced all direct calls to `Api.Instance` with `_api` to improve testability (`src/OpenFeature/OpenFeatureClient.cs`, [[1]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907R21) [[2]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907L43-R44) [[3]](diffhunk://#diff-c23c8a3ea4538fbdcf6b1cf93ea3de456906e4d267fc4b2ba3f8b1cb186a7907L102-R111).

### Project Configuration:

* Added `InternalsVisibleTo` for `OpenFeature.DependencyInjection` to allow access to internal members of the `OpenFeature` assembly (`src/OpenFeature/OpenFeature.csproj`, [src/OpenFeature/OpenFeature.csprojR22](diffhunk://#diff-711ea17cbdebe419375c7684c8c39a1423d2bebcf8976ddd7bdd78deaab65b21R22)).

### Related Issues

Fixes #533 

### Notes

As pointed out, I made a couple of changes in the `FeatureClient` class to use the same reference all the time for the `Api`. I checked the rest of the codebase and the `Api.Instance` is only used in the testing projects.
This change doesn't affect any external methods.

### DIFF
--- a/src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
@@ -24,7 +24,7 @@ public static partial class OpenFeatureServiceCollectionExtensions
         Guard.ThrowIfNull(configure);
 
         // Register core OpenFeature services as singletons.
-        services.TryAddSingleton(Api.Instance);
+        services.TryAddSingleton(new Api());
         services.TryAddSingleton<IFeatureLifecycleManager, FeatureLifecycleManager>();
 
         var builder = new OpenFeatureBuilder(services);
--- a/src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
+++ b/src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
@@ -24,7 +24,9 @@ public static partial class OpenFeatureServiceCollectionExtensions
         Guard.ThrowIfNull(configure);
 
         // Register core OpenFeature services as singletons.
-        services.TryAddSingleton(new Api());
+        var api = new Api();
+        Api.SetInstance(api);
+        services.TryAddSingleton(api);
         services.TryAddSingleton<IFeatureLifecycleManager, FeatureLifecycleManager>();
 
         var builder = new OpenFeatureBuilder(services);
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -121,7 +121,7 @@ public sealed class Api : IEventBus
     /// <returns><see cref="FeatureClient"/></returns>
     public FeatureClient GetClient(string? name = null, string? version = null, ILogger? logger = null,
         EvaluationContext? context = null) =>
-        new FeatureClient(() => this._repository.GetProvider(name), name, version, this, logger, context);
+        new FeatureClient(this, () => this._repository.GetProvider(name), name, version, logger, context);
 
     /// <summary>
     /// Appends list of hooks to global hooks list
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -121,7 +121,7 @@ public sealed class Api : IEventBus
     /// <returns><see cref="FeatureClient"/></returns>
     public FeatureClient GetClient(string? name = null, string? version = null, ILogger? logger = null,
         EvaluationContext? context = null) =>
-        new FeatureClient(() => this._repository.GetProvider(name), name, version, logger, context);
+        new FeatureClient(() => this._repository.GetProvider(name), name, version, this, logger, context);
 
     /// <summary>
     /// Appends list of hooks to global hooks list
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -360,4 +360,12 @@ public sealed class Api : IEventBus
     {
         Instance = new Api();
     }
+
+    /// <summary>
+    /// This method should only be used in the Dependency Injection setup. It will set the singleton instance of the API using the provided instance.
+    /// </summary>
+    internal static void SetInstance(Api api)
+    {
+        Instance = api;
+    }
 }
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -32,7 +32,7 @@ public sealed class Api : IEventBus
     // not to mark type as beforeFieldInit
     // IE Lazy way of ensuring this is thread safe without using locks
     static Api() { }
-    private Api() { }
+    internal Api() { }
 
     /// <summary>
     /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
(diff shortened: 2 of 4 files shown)


src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
```diff
@@ -24,7 +24,7 @@ public static IServiceCollection AddOpenFeature(this IServiceCollection services
         Guard.ThrowIfNull(configure);
 
         // Register core OpenFeature services as singletons.
-        services.TryAddSingleton(Api.Instance);
+        services.TryAddSingleton(new Api());
```
toddbaert: I think managing the singleton via the DI framework makes sense, but doing this still results in 2 (we still have another call to `new Api()` Api.cs).

I guess there's no easy way around this? Is there anything we can do to avoid both instances?

askpt: Are you talking about the `static` constructor?
That one is just a technique to lazy load the singleton. My concern is more if people start to mismatch calls using the`IFeatureClient` via DI and `API.Instance`, which I don't know how we could prevent it.

toddbaert: I'm saying that we now have 2 instances of the singleton (unless I'm really misunderstanding something) one managed by the SDK, instantiated [here on line 35](https://github.com/open-feature/dotnet-sdk/pull/534/files#diff-dc150fbd3b7be797470374ee7e38c7ab8a31eb9b6b7a52345e05114c2d32a15eR29), and another in the DI framework.

askpt: I got your point, so for the DI cases to keep always the correct instance, I created a new internal method that will set the same instance for the DI and the static Instance method. Check: https://github.com/open-feature/dotnet-sdk/pull/534/commits/2241a77c67e524dcc88188c5fa884fbfab0c91ce

kylejuliandev (approved):

src/OpenFeature.DependencyInjection/OpenFeatureServiceCollectionExtensions.cs
toddbaert: I think that's at least "good enough" if we discourage mixing .NET DI and "plain" API usage.

toddbaert (approved):

Issue #440, opened 2025-04-16
[FEATURE] Investigate NativeAOT


askpt:
### Description

The blog post ["Creating AOT-compatible libraries"](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/) highlights strategies and best practices for making .NET libraries compatible with NativeAOT. Investigating NativeAOT compatibility for the dotnet-sdk could enhance its performance and usability in AOT scenarios.

### Tasks

- [ ] Analyze the current implementation of dotnet-sdk for potential AOT compatibility issues.
- [ ] Identify dependencies or components that may require changes to support NativeAOT.
- [ ] Documentation of any limitations or challenges.

PR #537, opened 2025-07-31
feat: expose ValueJsonConverter and add JsonSourceGenerator test cases

### LINKED ISSUE #536: [BUG] JSON source generators broken by 2.8.0
https://github.com/open-feature/dotnet-sdk/pull/529 added an internal `JsonConverter<Value>` that breaks source JSON source generators, as the `JsonTypeInfo<Value>` is not generated because it needs to create the converter in the generated code.

The possible fixes are to either make [ValueJsonConverter](https://github.com/open-feature/dotnet-sdk/blob/c0eb12aaa6b2dbdc474d88e242d0e7659a3cc122/src/OpenFeature/Model/ValueJsonConverter.cs#L7) public, so that people can add `Value` to their own source generators, or provide a public source generator for people to use e.g.

```csharp
[JsonSerializable(typeof(Value))]
public partial class OpenFeatureJsonSerializerContext : JsonSerializerContext
{
}
```

WeihanLi:
## This PR

- Expose ValueJsonConverter and add JsonSourceGenerator serialize test cases

### Related Issues

Fixes #536


.github/workflows/ci.yml
```diff
@@ -53,6 +53,10 @@ jobs:
       - name: Test
         run: dotnet test -c Release --no-build --logger GitHubActions
 
+      - name: aot-publish test
+        run: |
+          dotnet publish ./samples/AspNetCore/Samples.AspNetCore.csproj
```
WeihanLi, line 58: Add aot publish step to verify the aot publish would not fail

askpt:
> Looks like this does solve #440 as well, at least partially? Should we mention that in the commit or break it out to another?

For this I am happy to keep it in this PR and investigate it better in a separate effort. This PR adds the specific properties to announce this library is NativeAOT, which it seems it is by the test scenario in the pipeline. We can keep it in that PR and then make an official announcement or even add more testing around it.

> Looks good to me, let's just be careful to make sure to acknowledge that when we make this public we are adding more surface area to the SDK's API that we must not break; if there's any possible breaking changes we want to make to the serializer, lets make them now.

I agree. I think @kinyoklion and @WeihanLi might have a better answer to it. I cannot see any possible breaking changes to this serialiser in the future (unless we add a new `ValueType` in the spec).

WeihanLi:
@askpt @toddbaert thanks very much for the review, it does contribute to the #440 and agree with `For this I am happy to keep it in this PR and investigate it better in a separate effort.`, this introduces initial AOT support for the JSON source generator verification, think we could investigate better separately.

For the json converter, think we may need to confirm on the convert type name and namespace since it's going to be public, and maybe also the document comment, to avoid future breaking changes.

Also cc @kylejuliandev for a review and input.

PR #568, opened 2025-08-26
feat: Add events to the multi provider

### LINKED ISSUE #552: [FEATURE] Implement events in multi-provider
### Requirements

Implement event support in the multi-provider in alignment with the OpenFeature events concept (see https://openfeature.dev/docs/reference/concepts/events).

**Requirements:**
- Integrate event handling capabilities so that the multi-provider can emit and handle relevant events as defined by the OpenFeature specification.
- Ensure that events such as provider state changes, flag evaluations, and errors are surfaced according to the conceptual documentation linked above.
- Add appropriate hooks for consumers to subscribe to and react to these events.
- Include documentation and usage examples for the new event features.

Reference: https://openfeature.dev/docs/reference/concepts/events

askpt:
## This PR

This pull request introduces event handling and status management improvements to the `MultiProvider` class in the `OpenFeature.Providers.MultiProvider` package. The main changes include adding infrastructure to listen to events from underlying providers, emitting events based on provider and aggregate status changes, and ensuring proper shutdown of event processing. These enhancements improve observability and reliability of the multi-provider setup.

### Event Handling Infrastructure

* Added event listening infrastructure to `MultiProvider`, including a dictionary to track event listening tasks and a cancellation token source for clean shutdown. Event listeners are started for all registered providers during initialization. [[1]](diffhunk://#diff-c8bc3d83c706a7332716c6d2773556d86f89ccbdf1d081f54aa9982f183d874aL31-R35) [[2]](diffhunk://#diff-c8bc3d83c706a7332716c6d2773556d86f89ccbdf1d081f54aa9982f183d874aR60-R62) [[3]](diffhunk://#diff-c8bc3d83c706a7332716c6d2773556d86f89ccbdf1d081f54aa9982f183d874aR282-R437)

### Event Emission and Status Management

* Implemented logic to emit error and ready events during initialization and evaluation, improving visibility into provider failures and readiness. [[1]](diffhunk://#diff-c8bc3d83c706a7332716c6d2773556d86f89ccbdf1d081f54aa9982f183d874aR127-R151) [[2]](diffhunk://#diff-c8bc3d83c706a7332716c6d2773556d86f89ccbdf1d081f54aa9982f183d874aR181-R226)
* Added logic to handle provider events, update individual provider statuses, and emit aggregate status change events for `MultiProvider`. Includes special handling for configuration change events.

### Shutdown and Disposal

* Ensured proper shutdown of event processing by cancelling event listeners and waiting for event tasks to complete during disposal.

### Project Configuration

* Updated `OpenFeature.csproj` to add `InternalsVisibleTo` for `OpenFeature.Providers.MultiProvider`, allowing internal access for testing and integration.

### Related Issues

Fixes #552 

### Notes

- I cannot get the coverage up to the requirements since there are some very edge cases that I cannot easily unit test. The proposed coverage is already good which covers most of the current implementation for events.


src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
```diff
@@ -28,7 +28,11 @@ public sealed class MultiProvider : FeatureProvider, IAsyncDisposable
     private ProviderStatus _providerStatus = ProviderStatus.NotReady;
     // 0 = Not disposed, 1 = Disposed
     // This is to handle the dispose pattern correctly with the async initialization and shutdown methods
-    private volatile int _disposed = 0;
+    private volatile int _disposed;
+
+    // Event handling infrastructure
+    private readonly Dictionary<FeatureProvider, Task> _eventListeningTasks = [];
```
kinyoklion: I am a bit concerned about the lack of synchronization for `_eventListeningTasks` and the `_providerStatus`

I don't believe there is anything causes init and shutdown themselves to either be on the same thread or to be synchronized. (The repository starts/shutsdown outside of its provider lock.)

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
```diff
@@ -28,7 +28,11 @@ public sealed class MultiProvider : FeatureProvider, IAsyncDisposable
...
+    private readonly CancellationTokenSource _eventProcessingCancellation = new();
```
kinyoklion, line 39: This disposable and I think we need to dispose it.

A second concern is that I don't think this source could be re-used if you wanted this provider to support being initialized again after being stopped. I think you would need to dispose this one and create a new one.

askpt: You have a point in your second concern. If we shut down the provider, we cannot initialise again (the `_disposed` is not allowing us), which goes against the https://openfeature.dev/specification/sections/flag-evaluation/#17-provider-lifecycle-management

We need to fix it in a separate PR from this one.

askpt: Fixed the dispose of the CTS with https://github.com/open-feature/dotnet-sdk/pull/568/commits/445fdaef97495dbf0c964b2ac2aff5795a94a8b2

askpt: @kinyoklion created https://github.com/open-feature/dotnet-sdk/issues/569 to tackle the lifecycle issues

kylejuliandev (approved):

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
askpt: This should now be fixed with https://github.com/open-feature/dotnet-sdk/pull/568/commits/518cd5286e8bc511939ab3ab3b7dfae978742cfc

test/OpenFeature.Providers.MultiProvider.Tests/Utils/TestProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    private readonly string _name;
+    private readonly Exception? _initException;
+    private readonly Exception? _shutdownException;
+
+    public TestProvider(string name, Exception? initException = null, Exception? shutdownException = null)
+    {
+        this._name = name;
+        this._initException = initException;
+        this._shutdownException = shutdownException;
```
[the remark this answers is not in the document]
askpt: This is an .editorconfig rule. Keeping it.

test/OpenFeature.Providers.MultiProvider.Tests/MultiProviderEventTests.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        // Verify that subsequent operations on disposed provider throw
+        await Assert.ThrowsAsync<ObjectDisposedException>(() =>
+            multiProvider.ResolveBooleanValueAsync(TestFlagKey, false));
+    }
+
+    // Helper methods
+    private MultiProvider CreateMultiProvider(params FeatureProvider[] providers)
+    {
+        var entries = providers.Select((p, i) => new ProviderEntry(p, $"provider{i + 1}")).ToList();
```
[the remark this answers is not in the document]
askpt: Not a problem here.

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
```diff
@@ -220,6 +300,183 @@ private async Task<List<ProviderResolutionResult<T>>> ParallelEvaluationAsync<T>
         return resolutions;
     }
 
+    private void StartListeningToProviderEvents()
+    {
+        foreach (var registeredProvider in this._registeredProviders)
+        {
+            this._eventListeningTasks[registeredProvider.Provider] = this.ProcessProviderEventsAsync(registeredProvider);
```
chrfwow: should we use something like a `putIfAbsent` here?

askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/568/commits/98c2d313b231bf39b0bc59e22b7bcd7269174084

chrfwow: To be really thread safe, [this method](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd?view=net-9.0) should be used

askpt: You are correct. Did a minor change to also log if there are issues adding the provider: https://github.com/open-feature/dotnet-sdk/pull/568/commits/0ca0a86bb5896a386aea06ed0db3e30781b09baa

chrfwow: I am not too familiar with the task concept in c#, but do we have to do anything with the task if we do not store it in the dictionary?

askpt: Great finding. I was doing a couple of checks, and it seems that with the new changes, the `Task` will execute even if it already exists in the dictionary. This way, it will only run if the key is not in the dictionary. See: https://github.com/open-feature/dotnet-sdk/pull/568/commits/4cfdb9bd2616028b41638a8248920fa0a653d7c6

chrfwow: If the task will be run whenever the `ProcessProviderEventsAsync` method is called, we can still end up in a situation where the task will be executed more than once with your current double-check setup. More than one thread might pass the `if (!this._eventListeningTasks.ContainsKey(key))` barrier before inserting the element into the dictionary. If we want to be completely safe, we have to find a way to only start the task after we have inserted it into the dictionary and discard it if another thread won.

askpt: We had a quick chat on slack. This method is only called in the constructor. The constructor runs on a single thread before the object is accessible to other threads, so there are no potential threading issues here.

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
```diff
@@ -220,6 +307,192 @@ private async Task<List<ProviderResolutionResult<T>>> ParallelEvaluationAsync<T>
...
+            var key = registeredProvider.Provider;
+            if (!this._eventListeningTasks.ContainsKey(key))
+            {
+                this._eventListeningTasks.TryAdd(key, value: this.ProcessProviderEventsAsync(registeredProvider));
+            }
+            else
+            {
+                // Log a warning if the provider is already being listened to
+                this.LogProviderAlreadyBeingListenedTo(registeredProvider.Name);
+            }
```
arttonoyan: **Avoid TOCTOU race when registering event listeners**
`StartListeningToProviderEvents` first calls `ContainsKey` and then `TryAdd`. Between those calls another thread can insert the same key, causing a benign race and a missed warning. Please collapse to a single TryAdd and log based on its return value to make the operation atomic.
**Why this is a problem:**
A check-then-act pattern on a concurrent dictionary isn’t atomic and can produce inconsistent behavior under contention.
**Suggested fix:**
```
private void StartListeningToProviderEvents()
{
    foreach (var rp in _registeredProviders)
    {
        if (!_eventListeningTasks.TryAdd(rp.Provider, ProcessProviderEventsAsync(rp)))
        {
            LogProviderAlreadyBeingListenedTo(rp.Name);
        }
    }
}
```

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
```diff
@@ -281,16 +554,45 @@ public async ValueTask DisposeAsync()
...
+        {
+            try
+            {
+                await Task.WhenAll(this._eventListeningTasks.Values).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                // Expected when shutting down
+            }
+            this.Status = ProviderStatus.Fatal;
```
arttonoyan: `OperationCanceledException` is handled like a generic failure and child status is set to `Fatal`. Should we treat cancellation as cooperative shutdown instead - i.e., catch `OperationCanceledException` separately, set status to `NotReady`, and propagate the cancellation (or return a cancellation marker) - rather than marking the provider `Fatal`?

arttonoyan (changes requested):

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/568/commits/a4814ba7e7c5c8612b05479d32a78398e6b7164d

In this case, there is no problem of multi-threading because this only executes in the constructor. For clarity, I followed your advice. Thanks!

src/OpenFeature.Providers.MultiProvider/MultiProvider.cs
askpt: Great catch! Fixed with: https://github.com/open-feature/dotnet-sdk/pull/568/commits/edd102a8ca87f562408c9a8baf11544818e6ce29

The Fatal should only happen in a couple of scenarios only like `Dispose` or a bad `Shutdown`.

PR #596, opened 2025-10-04
feat: Extract minimal DI integration to OpenFeature.Providers.DependencyInjection

### LINKED ISSUE #587: [FEATURE] Split OpenFeature.DependencyInjection from OpenFeature.Hosting
> [!IMPORTANT] 
**Update. 2025-11-10**
Following the discussion in [https://github.com/open-feature/dotnet-sdk/pull/596#discussion_r2495975595](https://github.com/open-feature/dotnet-sdk/pull/596#discussion_r2495975595). the package name in this proposal is updated from `OpenFeature.DependencyInjection` to `OpenFeature.Providers.DependencyInjection`. Rationale. clearer scope for provider authors. thinner provider packages. reduced coupling with the SDK.

### Requirements

**Context**
`OpenFeature.DependencyInjection` was deprecated in favor of `OpenFeature.Hosting`, which now includes DI-related functionality. This simplifies app integration, but provider packages (e.g., LaunchDarkly) often need only a very small surface to integrate—without pulling in broader hosting concepts (hosted lifecycles, health checks, config binding, etc.).
**This idea came to me after integrating `launchdarkly/openfeature-dotnet-server` DI, specifically after this question ([https://github.com/launchdarkly/openfeature-dotnet-server/pull/49#issuecomment-3334928431](https://github.com/launchdarkly/openfeature-dotnet-server/pull/49#issuecomment-3334928431)). In that integration I used only an `OpenFeatureBuilder` and `AddProvider` methods with their overloads.**

**Problem**
Provider libraries that just need to plug a provider into OpenFeature end up depending on `OpenFeature.Hosting`, which can feel heavy and increase the transitive dependency surface for scenarios that don’t use hosting features.

**Proposal**
Introduce a *thin* DI-friendly package, **`OpenFeature.DependencyInjection.Abstractions`**, that exposes a small **builder** abstraction (e.g., `OpenFeatureBuilder`) for provider composition—methods like `AddProvider(...)`, `AddHook(...)`, and minimal configuration hooks.
Crucially, **all DI registrations remain in `OpenFeature.Hosting`**, which creates/owns the builder and applies configuration, lifecycle, health checks, and other hosting features. Providers would reference the abstractions package only, extending the builder without taking a hard dependency on hosting.

**Benefits**

* **Separation of concerns:** App authors get the full Hosting experience; provider authors get a tiny, stable surface to integrate.
* **Smaller dependency footprint:** Provider packages avoid bringing in hosting infrastructure they don’t use.
* **Broader compatibility:** Clean support for libraries, tests, console tools, and worker services where hosting may be unnecessary.
* **Future-proofing:** Hosting features (Aspire/telemetry/lifecycle) can evolve independently, while the builder surface remains small and stable.

**Trade-offs / Considerations**

* Another package to version and document.
* Clear guidance needed to avoid confusion about when to use Hosting vs. the abstractions.
* Keep the abstractions narrowly scoped to avoid duplicating Hosting responsibilities.

**Back-compat / Migration**

* Keep `OpenFeature.DependencyInjection` deprecated; optionally publish a final patch that forwards types where feasible and points to the new package.
* Provide a short migration guide:

  * **App developers:** use `OpenFeature.Hosting`.
  * **Provider authors:** target `OpenFeature.DependencyInjection.Abstractions` and extend the builder.
* Observability, lifecycle, and config binding remain in Hosting.

**Questions for the community**

1. Is there agreement that a DI-only **builder** abstraction (with Hosting retaining DI registrations) would help provider authors?
2. Any concerns about maintenance or user confusion that outweigh the benefits?
3. Naming preference: **`OpenFeature.DependencyInjection.Abstractions`** (my preferred option), or an alternative?
4. Which minimal builder methods are essential, and which should remain strictly in Hosting?
5. If there’s consensus, I can help draft an RFC and a small PoC to validate the split.

**Related context**
This proposal arose while updating a provider PR where the Hosting dependency felt heavier than necessary for provider-level integration. The builder-only approach allowed the provider to integrate cleanly without consuming Hosting features.

arttonoyan:
> [!IMPORTANT] 
**Update. 2025-11-10**
Following the discussion in [https://github.com/open-feature/dotnet-sdk/pull/596#discussion_r2495975595](https://github.com/open-feature/dotnet-sdk/pull/596#discussion_r2495975595). the package name in this proposal is updated from `OpenFeature.DependencyInjection` to `OpenFeature.Providers.DependencyInjection`. Rationale. clearer scope for provider authors. thinner provider packages. reduced coupling with the SDK.

## This PR
Extract the minimal, provider-agnostic DI surface into a new package: `OpenFeature.DependencyInjection.Abstractions`.
This isolates the contracts and lightweight wiring needed to integrate any OpenFeature provider without pulling in a concrete implementation.

### Related Issues

Fixes #587 

### Notes
* **Options split:** Extracted provider-agnostic configuration from `OpenFeatureOptions` into a new base options type: **`OpenFeatureProviderOptions`**.
* **Inheritance:** `OpenFeatureOptions` now **inherits** from `OpenFeatureProviderOptions`.
* **Internal wiring update:** Internal configuration that previously targeted `OpenFeatureOptions` now targets `OpenFeatureProviderOptions`.

### Before (internal)

```csharp
services.AddOptions<OpenFeatureOptions>()
    .Configure(options =>
    {
        options.AddProviderName(null);
    });
```

### After (internal)

```csharp
services.AddOptions<OpenFeatureProviderOptions>()
    .Configure(options =>
    {
        options.AddProviderName(null);
    });
```

> **Note:** This is an internal refactor. It **does not affect consumers** unless they were directly modifying `OpenFeatureOptions` via DI options configuration (which is not a supported/typical usage).

---

### Impact

* **No new functionality.**
* **Behavior unchanged** at runtime (same provider selection, same evaluation outcomes, same logs).
* **Binary/source compatibility:** Consumers using the documented extension methods and standard registration remain unaffected.
* **Potential touchpoints:** Only internal code paths (or unconventional consumer code directly configuring `OpenFeatureOptions`) require updates to target `OpenFeatureProviderOptions`.

---

### How to test

This is a **regression-only** refactor.

**Expectations**

* All existing unit/integration tests pass **unchanged**.
* No DI resolution or startup configuration differences.


OpenFeature.slnx
```diff
@@ -48,6 +48,7 @@
     <Project Path="samples/AspNetCore/Samples.AspNetCore.csproj" />
   </Folder>
   <Folder Name="/src/">
+    <Project Path="src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj" Id="607a23fa-2c2f-4d2e-9079-adae18326512" />
```
kylejuliandev: suggestion: I'm not sure the ID property is needed here

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderOptions.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.ObjectModel;
+
+namespace OpenFeature.DependencyInjection.Abstractions;
+
+/// <summary>
+/// Provider-focused options for configuring OpenFeature integrations.
+/// Contains only contracts and metadata that integrations may need.
+/// </summary>
+public class OpenFeatureProviderOptions
```
kylejuliandev, line 9: question: Whats the thinking behind renaming the `OpenFeatureOptions` to `OpenFeatureProviderOptions`?

src/OpenFeature.Providers.DependencyInjection/Diagnostics/FeatureCodes.cs
```diff
@@ -22,7 +22,7 @@ namespace OpenFeature.Hosting.Diagnostics;
 ///     - "001" - Unique identifier for a specific feature.
 /// </code>
 /// </example>
-internal static class FeatureCodes
+public static class FeatureCodes
```
kylejuliandev, line 25: question: Do we need these FeatureCodes anymore?

src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net8.0;net9.0;net462</TargetFrameworks>
+    <RootNamespace>OpenFeature.DependencyInjection.Abstractions</RootNamespace>
```
kylejuliandev: suggestion: Should we have these abstractions in their own namespace? Would it be more accessible to have them in say just the `OpenFeature` namespace?

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderBuilder.cs
```diff
@@ -0,0 +1,62 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace OpenFeature.DependencyInjection.Abstractions;
+
+/// <summary>
+/// Describes a <see cref="OpenFeatureProviderBuilder"/> backed by an <see cref="IServiceCollection"/>.
+/// </summary>
+public abstract class OpenFeatureProviderBuilder(IServiceCollection services)
```
kylejuliandev, line 8: question: What's the thinking with renaming this? From `OpenFeatureBuilder` to `OpenFeatureProviderBuilder`?

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
```diff
@@ -34,7 +43,12 @@ public async ValueTask EnsureInitializedAsync(CancellationToken cancellationToke
             var featureProvider = _serviceProvider.GetRequiredKeyedService<FeatureProvider>(name);
             await _featureApi.SetProviderAsync(name, featureProvider).ConfigureAwait(false);
         }
+    }
 
+    /// <inheritdoc />
```
kylejuliandev: suggestion: Not sure we need the `inheritdoc` section here

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderBuilder.cs
kylejuliandev: Oh wait I see, these are provider specific. Nvm this comment.

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderOptions.cs
kylejuliandev: Oh wait I see, these are provider specific. Nvm this comment.

test/OpenFeature.Hosting.Tests/Internal/FeatureLifecycleManagerTests.cs
```diff
@@ -15,7 +16,7 @@ public async Task EnsureInitializedAsync_SetsProvider()
         // Arrange
         var services = new ServiceCollection();
         var provider = new NoOpFeatureProvider();
-        services.AddOptions<OpenFeatureOptions>().Configure(options =>
+        services.AddOptions<OpenFeatureProviderOptions>().Configure(options =>
```
kylejuliandev, line 19: suggestion: it should be fine for this to remain `OpenFeatureOptions` right? `OpenFeatureOptions` extends `OpenFeatureProviderOptions`

src/OpenFeature.Hosting/OpenFeatureServiceCollectionExtensions.cs
```diff
@@ -49,7 +50,7 @@ public static IServiceCollection AddOpenFeature(this IServiceCollection services
             {
                 options.DefaultNameSelector = provider =>
                 {
-                    var options = provider.GetRequiredService<IOptions<OpenFeatureOptions>>().Value;
+                    var options = provider.GetRequiredService<IOptions<OpenFeatureProviderOptions>>().Value;
```
kylejuliandev, line 54: suggestion: it should be fine for this to remain `OpenFeatureOptions` right? `OpenFeatureOptions` extends `OpenFeatureProviderOptions`

src/OpenFeature.Hosting/OpenFeatureOptions.cs
```diff
@@ -1,52 +1,12 @@
...
-        else
-        {
-            lock (_providerNames)
-            {
-                _providerNames.Add(name!);
-            }
-        }
-    }
-
     private readonly HashSet<string> _hookNames = [];
```
kylejuliandev, line 8: thought: I wonder if we can simplify the Hook registration. If instead of relying on HookNames we just inject Hooks like `services.AddTransient<Hook>(myHook)`. And when we retrieve these hooks out to register them in the provider we just read it back out like `var hooks = _serviceProvider.GetServices<Hook>();`

Would eliminate the need to track hook names in this options file

OpenFeature.slnx
arttonoyan: Removed: https://github.com/open-feature/dotnet-sdk/pull/596/commits/3d680f3f054cba3b9547f276d09a8016bf26c66d

src/OpenFeature.Providers.DependencyInjection/Diagnostics/FeatureCodes.cs
arttonoyan: @kylejuliandev **Answer:** Yes - within the scope of this PR, `FeatureCodes` are still required; the current changes reference them.
**Follow-up proposal:** It’s a good time to graduate them from `Experimental`. Not in this PR - I suggest a separate task.
@askpt @beeme1mr what do you think about me opening a follow-up issue to:
* remove the `Experimental` attribute,
* update docs/usages,
* add a changelog entry and brief migration note?

src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj
arttonoyan: @kylejuliandev Great question - thanks for calling it out.
I’d keep these in their own namespace for now. `OpenFeature.DependencyInjection.Abstractions` targets provider authors (integrating new providers), not typical app consumers. Keeping it separate avoids cluttering the root `OpenFeature` namespace and lets us evolve DI-specific types independently.
Of course, I’m open to discussion about this.

src/OpenFeature.Providers.DependencyInjection/Diagnostics/FeatureCodes.cs
askpt: I think with @kylejuliandev efforts to deprecate the `OpenFeature.DependencyInjection` library and this would be great candidate for 2.10. Feel free to create a new issue and add it to 2.10 milestone 👍

src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj
askpt: I prefer the @arttonoyan. This is a very specific Abstraction layer. 

I am wondering if we should keep in a `OpenFeature.Abstractions` instead. This would be a "library" of just `OpenFeature` related abstractions like "hooks", "provider"... I am happy to keep `OpenFeature.DependencyInjection.Abstractions` if the rest of the team agrees.

src/OpenFeature.Providers.DependencyInjection/OpenFeature.Providers.DependencyInjection.csproj
askpt, line 1: This file misses the README file necessary for the package documentation.

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,145 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using OpenFeature.DependencyInjection.Abstractions;
+
+namespace OpenFeature;
```
askpt, line 5: Any particular reason why this namespace is `OpenFeature` and not `OpenFeature.DependencyInjection.Abstractions`?

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
```diff
@@ -22,7 +23,15 @@ public async ValueTask EnsureInitializedAsync(CancellationToken cancellationToke
     {
         this.LogStartingInitializationOfFeatureProvider();
 
-        var options = _serviceProvider.GetRequiredService<IOptions<OpenFeatureOptions>>().Value;
+        await InitializeProvidersAsync(cancellationToken).ConfigureAwait(false);
+        InitializeHooks();
+        InitializeHandlers();
+    }
+
+    /// <inheritdoc />
```
askpt: It is private method, so this is not needed.

```suggestion
```

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
askpt:
```suggestion
```

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
```diff
@@ -22,7 +23,15 @@ public async ValueTask EnsureInitializedAsync(CancellationToken cancellationToke
...
+    private async Task InitializeProvidersAsync(CancellationToken cancellationToken)
```
askpt, line 30: cancellationToken seems to not be used here.

src/OpenFeature.Hosting/Providers/Memory/FeatureBuilderExtensions.cs
```diff
@@ -1,27 +1,28 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
+using OpenFeature.DependencyInjection.Abstractions;
 using OpenFeature.Providers.Memory;
 
 namespace OpenFeature.Hosting.Providers.Memory;
```
askpt, line 6: I wonder if we should move to `OpenFeature.Providers.Memory`

src/OpenFeature.Hosting/OpenFeatureOptions.cs
askpt: This seems a great suggestion. My concern is if we register two different instance `Hooks` of the same type. What should happen? Should we allow this behaviour?

askpt (commented):
Great work! I only added a couple of minor comments. 

Please have a chat with @kylejuliandev to try understand the impact on his deprecations efforts. I remember he mentioned a couple of fixes he was working on for the DI and I don't want that effort to get lost during this PR.

src/OpenFeature.Hosting/OpenFeatureOptions.cs
kylejuliandev: I went ahead over the weekend and gave it a stab in https://github.com/open-feature/dotnet-sdk/pull/609. Seems simple enough. I think if we register two hooks of the same type, then we should still handle them both. Maybe let's take this conversation over there. Happy to close this here as it's not really relevant to the changes

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderBuilderExtensions.cs
kylejuliandev: These extensions are available to devs directly when they setup OpenFeature with the OpenFeature builder, I think leaving them in the `OpenFeature` namespace is less disruptive and makes it more easily accessible

src/OpenFeature.Hosting/Providers/Memory/FeatureBuilderExtensions.cs
arttonoyan: I agree. Moving to `OpenFeature.Providers.Memory` makes sense. I had the same thought, timing just got in the way. One tweak. I would name it `InMemory` rather than `Memory`.

src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj
arttonoyan: @askpt @kylejuliandev

**Proposal**
Rename the package to `OpenFeature.Providers.DependencyInjection`. This name better reflects scope. It is intuitive for provider authors who want to add DI integration.

**Follow-up idea**
In a separate discussion, consider extracting common provider contracts into `OpenFeature.Providers.Abstractions`. We would then have two clear packages. `OpenFeature.Providers.Abstractions`, and `OpenFeature.Providers.DependencyInjection`.

**Why this helps**

* Clear separation. abstractions live in one place, DI glue in another.
* Thinner provider packages. they depend on `Abstractions`, optionally on the DI helpers.
* Cleaner dependency graph. no DI references inside core contracts.
* Easier discovery on NuGet. consistent `OpenFeature.Providers.*` naming.

cc: @beeme1mr

beeme1mr: That seems reasonable to me. What would be the proposed release strategy and impact? I believe the DI package is still marked as experimental, so breaking changes are allowed, but we should try and make it as straightforward for users to migrate. Also, I'd hope that we could remove the experimental badge shortly after making this change.

arttonoyan: Right now our main pain point is coupling. The current abstractions package depends on the OpenFeature SDK because it includes the `OpenFeatureProvider` abstraction. This means any SDK release. even when it has no provider changes. forces us to bump the abstractions package. Then every provider sees a new version and feels compelled to upgrade. even though nothing changed for providers. Providers should only react to provider related changes.

Decoupling fixes this. If we move provider contracts into `OpenFeature.Providers.Abstractions`. we isolate them from SDK churn. The DI helpers stay in `OpenFeature.Providers.DependencyInjection`. Result. cleaner dependency graph. fewer unnecessary upgrades. thinner provider packages. and clearer ownership of changes.

I realize this is a significant change, and it needs attentive and careful attention. If the direction sounds reasonable, I can open a PR to rename the DI package, then draft a proposal for the abstractions split.

beeme1mr: Sounds reasonable to me. We've been looking into something similar (minus the DI part) in Java. FYI @chrfwow @toddbaert

src/OpenFeature.Hosting/OpenFeatureServiceCollectionExtensions.cs
arttonoyan: @kylejuliandev This is a good question. We cannot do that, because when registering a provider we configure `OpenFeatureProviderOptions`, not `OpenFeatureOptions`. If we resolve `OpenFeatureOptions`, the collection of registered providers will be empty.
This is on me. I made `OpenFeatureOptions` inherit from `OpenFeatureProviderOptions`, which is confusing, including for me. Your question highlights that design issue and it is totally valid.
I will remove the inheritance and keep these as two independent option types. There is no need for them to be related.

arttonoyan: @kylejuliandev Fixed: https://github.com/open-feature/dotnet-sdk/pull/596/commits/f52132178ebf244fd8de653cbe5897f42c731c33

src/OpenFeature.DependencyInjection.Abstractions/OpenFeature.DependencyInjection.Abstractions.csproj
arttonoyan: Great. I’ll proceed and rename the package to `OpenFeature.Providers.DependencyInjection` in this PR.

askpt: @arttonoyan @beeme1mr @kylejuliandev: I think this is a great discussion.

I would propose for 2.10 to create:
- `OpenFeature.Providers.DependencyInjection` (this PR)

For 2.11:
- `OpenFeature.Providers.Abstractions`
- `OpenFeature.Hooks.Abstractions`

The story that is happening in Providers, is also true for the Hooks.

src/OpenFeature.Providers.DependencyInjection/OpenFeature.Providers.DependencyInjection.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net8.0;net9.0;net462</TargetFrameworks>
```
askpt:
```suggestion
```

Please remove this. It is now managed centrally by https://github.com/open-feature/dotnet-sdk/pull/611

arttonoyan:
@askpt @kylejuliandev Could you please do a final review? I’d like to merge this PR.

test/OpenFeature.Hosting.Tests/Internal/FeatureLifecycleManagerTests.cs
arttonoyan: @kylejuliandev Fixed: https://github.com/open-feature/dotnet-sdk/commit/f52132178ebf244fd8de653cbe5897f42c731c33

src/OpenFeature.Providers.DependencyInjection/OpenFeature.Providers.DependencyInjection.csproj
arttonoyan: Will be done within the task https://github.com/open-feature/dotnet-sdk/issues/633

arttonoyan: Thanks for the heads-up. @kylejuliandev does this touch any of your DI deprecation items? If so, I’ll align with your plan and update this PR accordingly. Let me know what you’d like to see changed.

src/OpenFeature.Providers.DependencyInjection/Diagnostics/FeatureCodes.cs
arttonoyan: @kylejuliandev Yes, still needed while the DI integration is marked Experimental. Once we graduate it to standard support, we can remove them. For this PR, I recommend keeping them. If you’re OK with that, please resolve this thread.

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
kylejuliandev: I had always assumed SetProviderAsync would have a cancellationToken overload, but it doesn't appear like it does

https://github.com/open-feature/dotnet-sdk/blob/7805184b40d483d7a55a084b153de13593c1e76b/src/OpenFeature/Api.cs#L37-L66

src/OpenFeature.Providers.DependencyInjection/OpenFeatureProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// <typeparam name="TOptions">The type of options used to configure <see cref="OpenFeatureProviderOptions"/>.</typeparam>
+    /// <param name="builder">The <see cref="OpenFeatureProviderBuilder"/> instance.</param>
+    /// <param name="configureOptions">A delegate to configure <typeparamref name="TOptions"/>.</param>
+    /// <returns>The configured <see cref="OpenFeatureProviderBuilder"/> instance.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> or <paramref name="configureOptions"/> is null.</exception>
+    public static OpenFeatureProviderBuilder AddPolicyName<TOptions>(this OpenFeatureProviderBuilder builder, Action<TOptions> configureOptions)
+        where TOptions : PolicyNameOptions
+    {
+        if (builder == null) throw new ArgumentNullException(nameof(builder));
+        if (configureOptions == null) throw new ArgumentNullException(nameof(configureOptions));
```
kylejuliandev, line 129: suggestion(non-blocking): on a subsequent PR maybe we should include some sensible error messages when throwing these exceptions

kylejuliandev (approved):

src/OpenFeature.Hosting/Internal/FeatureLifecycleManager.cs
arttonoyan: @askpt Addressed in this comment: https://github.com/open-feature/dotnet-sdk/pull/596/commits/0843aac64eb3aee19a25cda03e09c4834c300110

askpt:
> @askpt @kylejuliandev I ran into a blocking issue. I tried to avoid any major changes, but it looks like I can’t solve it without doing at least some refactoring. If you can help with a decision, that would be great.
> 
> **The problem**
> 
> While working on the `OpenFeature.Providers.DependencyInjection` package, I hit an issue with this configuration:
> 
> ```cs
> builder.Services.AddOpenFeature(of => of
>     .AddProvider(_ => new TestProvider())
>     .AddHook(_ => new TestHook())
> );
> ```
> 
> This fails because:
> 
> * `AddProvider` is an extension for `OpenFeatureProviderBuilder`
> * `AddHook` is an extension for `OpenFeatureBuilder`
> * `AddProvider` returns `OpenFeatureProviderBuilder`, which does **not** have `AddHook`, so the second call is not available and we get a compile-time error.
> 
> If I reverse the order, it works:
> 
> ```cs
> builder.Services.AddOpenFeature(of => of
>     .AddHook(_ => new TestHook())
>     .AddProvider(_ => new TestProvider())
> );
> ```
> 
> Here `AddHook` returns `OpenFeatureBuilder`, which inherits from `OpenFeatureProviderBuilder`, so both methods are available and everything compiles.
> 
> For me this is a design problem, because the API becomes order-dependent and confusing. I tried to make `OpenFeatureProviderBuilder` generic, but that didn’t really solve it.
> 
> I see two options. I’m open to other ideas as well.
> 
> ### Option 1. Move `OpenFeatureBuilder` into `OpenFeature.Providers.DependencyInjection` and remove `OpenFeatureProviderBuilder`
> We would keep all extensions targeting `OpenFeatureBuilder`.
> 
> * **Pros**. Simple. fast to implement.
> * **Cons**. All extensions would depend on the `OpenFeature.Providers.DependencyInjection` package, even when they are not really about providers or DI (for example, hooks).
> 
> ### Option 2. Create `OpenFeature.DependencyInjection.Abstractions` package
> We create a new package. `OpenFeature.DependencyInjection.Abstractions`, move all general DI-related types there, and remove `OpenFeatureProviderBuilder`. All extensions would be built on top of `OpenFeatureBuilder`.
> 
> * **Pros**. Cleaner separation. All DI logic depends on a lightweight abstractions package. In the future, if we want to split hooks or other concerns into separate packages, it will be easier.
> * **Cons**. One more package to maintain. This new package will probably contain most of the logic from the old `OpenFeature.DependencyInjection` package.
> 
> I slightly prefer **Option 2**, even though it introduces an extra package, because it gives us a cleaner and more future-proof structure. cc: @beeme1mr @toddbaert @lukas-reining @thomaspoignant @kinyoklion

Need to have a thought. If this requires another package moving/creation/deprecation we might as well do a v3 like I mentioned in Slack.

I kinda prefer Option B also, but moving the code all the time doesn't bring the necessary stability, even if this is still in Experimental phase.

I would like to know what will be the consequences for the providers and hooks if we go ahead with option B. Which package should it consume? `OpenFeature.Providers.Abstractions` and `OpenFeature.DependencyInjection.Abstractions`?

Then the consumers would just install `OpenFeature`, `OpenFeature.Hosting` and `OpenFeature.Providers.ProviderA`?

To finish, I would believe if we go with option B, we would need to do a v3, make the DI not-experimental and this PR will be part of v3.

askpt:
@arttonoyan This PR has several compilation errors. I will have a look when they get fixed. Which option will you think you will implement for the extraction? If there are any necessary breaking changes please rename the title to `feat!: Extract...`

arttonoyan:
> @arttonoyan This PR has several compilation errors. I will have a look when they get fixed. Which option will you think you will implement for the extraction? If there are any necessary breaking changes please rename the title to `feat!: Extract...`

@askpt As we discussed earlier, there is a breaking change related to `OpenFeatureBuilder`, so I decided not to fix and merge this PR as is.

Instead, I’ve started working on the next step. extracting `OpenFeature.DependencyInjection.Abstractions`. I created a new branch from this one, and once it’s ready, I’ll open a new PR based on it.

This should properly address the breaking change and also make future work easier. It will let us extract things like handlers and hooks into separate projects without causing more breaking changes.

I’ll open the PR soon.

arttonoyan:
I have a follow-up PR https://github.com/arttonoyan/dotnet-sdk/pull/4 that builds on top of this one.

Because PR https://github.com/arttonoyan/dotnet-sdk/pull/4 is based on the branch in my fork (`arttonoyan:feature/di-abstractions-builder-api-587`), I can’t open it directly against `open-feature/dotnet-sdk` until this PR is merged (GitHub requires the PR base branch to exist in the target repo).

What it adds (high level):
- Introduces `OpenFeature.DependencyInjection.Abstractions` and moves the builder/component primitives there
- Updates Hosting + provider DI wiring to use the new split and build-time aggregation

At this stage I’m mainly looking for **architecture feedback** (layering, component approach, package boundaries), not a full nit-level review yet.  
If the community agrees the direction is good, I’ll merge PR https://github.com/arttonoyan/dotnet-sdk/pull/4 into the base branch in my fork, polish it (tests + README/docs), and then open the upstream PR once this PR lands so the code review can continue normally on the SDK side.
@beeme1mr @askpt @kylejuliandev

PR #604, opened 2025-10-19
test(e2e): Update e2e tests to support newer spec submodule

kylejuliandev:
## This PR

Add code to handle the new scenarios covered in the specification gherkin files. Some of the scenarios are skipped as the core SDK does not have support for it yet, for example like disabled InMemoryProvider flags.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -42,7 +42,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4
+      uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee # v4
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -56,7 +56,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4
+      uses: github/codeql-action/autobuild@0499de31b99561a6d14a36a5f662c2a54f91beee # v4
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
@@ -69,4 +69,4 @@ jobs:
     #   ./location_of_script_within_repo/buildscript.sh
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4
+      uses: github/codeql-action/analyze@0499de31b99561a6d14a36a5f662c2a54f91beee # v4
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -30,7 +30,7 @@
 
   <ItemGroup Label="test">
     <PackageVersion Include="AutoFixture" Version="5.0.0-preview0012" />
-    <PackageVersion Include="BenchmarkDotNet" Version="0.15.4" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.15.5" />
     <PackageVersion Include="coverlet.collector" Version="6.0.4" />
     <PackageVersion Include="coverlet.msbuild" Version="6.0.4" />
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1" />
@@ -40,7 +40,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="OpenTelemetry" Version="1.13.1" />
     <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.13.1" />
-    <PackageVersion Include="Reqnroll.xUnit" Version="2.4.1" />
+    <PackageVersion Include="Reqnroll.xUnit" Version="3.2.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
   </ItemGroup>
--- a/test/OpenFeature.E2ETests/OpenFeature.E2ETests.csproj
+++ b/test/OpenFeature.E2ETests/OpenFeature.E2ETests.csproj
@@ -24,6 +24,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <!-- Reqnroll 3.2.0 requires System.Threading.Channels >= 9.0.6 -->
+    <PackageVersion Update="System.Threading.Channels" Version="9.0.6" />
   </ItemGroup>
 
   <ItemGroup>


test/OpenFeature.E2ETests/Steps/Evaluationv2StepDefinitions.cs
```diff
@@ -0,0 +1,377 @@
...
+        foreach (var field in typeof(TEnum).GetFields(BindingFlags.Public | BindingFlags.Static))
+        {
+            var attr = field.GetCustomAttribute<DescriptionAttribute>();
+            if (attr != null && attr.Description == description)
+            {
+                return (TEnum)field.GetValue(null)!;
+            }
+        }
+        throw new ArgumentException($"No {typeof(TEnum).Name} with description '{description}' found.");
+    }
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [c0ee595](https://github.com/open-feature/dotnet-sdk/pull/604/commits/c0ee5957c69270825c97de0d470106e4f543f950)

test/OpenFeature.E2ETests/Steps/Evaluationv2StepDefinitions.cs
```diff
@@ -0,0 +1,377 @@
...
+                Assert.Equal(reason, booleanResult.Reason);
+                break;
+            case FlagType.Object:
+                Skip.If(true, "Object e2e test not supported");
+                break;
+            default:
+                Assert.Fail("FlagType not yet supported.");
+                break;
+        }
+    }
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [f5c2c76](https://github.com/open-feature/dotnet-sdk/pull/604/commits/f5c2c767045b80e401ea3b06875b894bec019a67)

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
```diff
@@ -17,43 +18,69 @@ public BaseStepDefinitions(State state)
...
+
+        var memProvider = new InMemoryProvider(flags);
         await Api.Instance.SetProviderAsync(memProvider).ConfigureAwait(false);
         this.State.Client = Api.Instance.GetClient("TestClient", "1.0.0");
     }
 
     [Given(@"a Boolean-flag with key ""(.*)"" and a default value ""(.*)""")]
+    [Given(@"a Boolean-flag with key ""(.*)"" and a fallback value ""(.*)""")]
     [Given(@"a boolean-flag with key ""(.*)"" and a default value ""(.*)""")]
+    [Given(@"a boolean-flag with key ""(.*)"" and a fallback value ""(.*)""")]
```
chrfwow: Why do we need this twice?

askpt: As far as I remember there were a couple of steps where the type was written as Boolean and boolean. So both scenarios, even they are the "same" code wise, we need to write twice. Correct me if I'm wrong @kylejuliandev

kylejuliandev: Yeah exactly that. Reqnroll is case sensitive when it comes to step definitions, so I needed to declare the permutations (upper and lower case, as the gherkin files have a mix)

aepfli: sidequestion, can we also handle this like we do it in java/python etc. and create a mapping on the first "type"-string? like

```csharp
[Given(@"a "(.*)"-flag with key ""(.*)"" and a fallback value ""(.*)""")]
public void GivenAFlagWithKeyAndADefaultValue(string type, string key, string defaultType)
{
    var flagType = ComputeFlagType(type);
    var flagState = new FlagState(key, defaultType, flagType);
    this.State.Flag = flagState;
}
```
and `ComputeFlagType` could take the string, lowercase it, and handle this like a switch case?

but anyways, we should normalize the gherkin files, and create a ticket for this, or pr, to make everything capitalized, i think it is just the "hooks.feature"

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
```diff
@@ -255,4 +255,18 @@ public void Giventhereasonshouldindicateanerrorandtheerrorcodeshouldindicateatyp
...
+        this.State.Flag = new FlagState(flagKey, defaultValue.ToString(), FlagType.String);
+        this.State.FlagEvaluationDetailsResult = await this.State.Client!.GetStringDetailsAsync(flagKey, defaultValue).ConfigureAwait(false);
+    }
+
+    [When(@"a string flag with key ""(.*)"" is evaluated as an integer, with details and a fallback value (.*)")]
+    public async Task WhenAStringFlagWithKeyIsEvaluatedAsAnIntegerWithDetailsAndAFallbackValue(string flagKey, int defaultValue)
+    {
+        this.State.Flag = new FlagState(flagKey, defaultValue.ToString(), FlagType.Integer);
+        this.State.FlagEvaluationDetailsResult = await this.State.Client!.GetIntegerDetailsAsync(flagKey, defaultValue).ConfigureAwait(false);
+    }
```
aepfli: those are part of the deprecated feature files, annotated with `@deprecated`-tag. As we are not planning to support them in the future, and they are covered with other newer gherkin tests, should we care about those, or should we simply ignore those steps for now?

test/OpenFeature.E2ETests/Steps/Evaluationv2StepDefinitions.cs
```diff
@@ -0,0 +1,314 @@
+using OpenFeature.Constant;
+using OpenFeature.E2ETests.Utils;
+using OpenFeature.Model;
+
+namespace OpenFeature.E2ETests.Steps;
+
+[Binding]
+[Scope(Feature = "Flag Evaluations - Complete OpenFeature Specification Coverage")]
+public class Evaluationv2StepDefinitions : BaseStepDefinitions
```
aepfli: In other languages, we started to separate steps based on their usage, rather their file. so we can reuse them for other files.

We try to generate reusable definitions for all of those steps, and reuse those steps for all gherkin files. This separation also makes it easier to maintain and debug, as it is not feature/gherkin file based, but implementation based. maybe it is worth also following this approach here.

so far we established this kind of separation:
- providerSteps - for handling provider related topic
- flagSteps - evaluation and flag related
- contextSteps - context related
- metadataSteps - metadata related
- hookSteps - hook related

the background is, in the long run, if we do this properly, these steps could theoretically also be used by provider implementations. the only thing they need to do is providing own providerSteps/Mappings, and they could test if they are openFeature compliant and everything works as intended. (really long run idea)

aepfli (approved):
I am happy to see, that we are moving further into the gherkin spec direction. i left some nits, i think this move to the json file, and the new format for test steps, helps us tremendiously to ensure compliance between sdks. thank you for this contribution

test/OpenFeature.E2ETests/Steps/BaseStepDefinitions.cs
kylejuliandev: Fixed in [56bf219](https://github.com/open-feature/dotnet-sdk/pull/604/commits/56bf21912f235a718a63fd78bac23cd7f7dd6c64)

I needed a seperate StepArgumentTransformation to handle the permutations of `<flagtype>-flag`, but this is definitley much more readable

test/OpenFeature.E2ETests/Steps/EvaluationStepDefinitions.cs
kylejuliandev: Removed these in [d0e3580](https://github.com/open-feature/dotnet-sdk/pull/604/commits/d0e3580bab6dd1ebb8071d55fef71b251d97c1b8). I think when I was trying to get the e2e tests initially updated I went through and added all the step definitions. I suspect I added these before I added the exclusion on `@deprecated`.

test/OpenFeature.E2ETests/Steps/Evaluationv2StepDefinitions.cs
kylejuliandev: I've attempted to align the step definition files to the types of separation. I've left the older `EvaluationStepDefinitions` file for now, this maps to one of the deprecated feature files. [f5470eb](https://github.com/open-feature/dotnet-sdk/pull/604/commits/f5470eb64d3aade957a3296551aa2110652ee93b) has the changes for this.

aepfli (approved):

PR #605, opened 2025-10-20
fix: Ensure EvaluationContext is reliably added to the injected FeatureClient

### LINKED ISSUE #542: [BUG] EvaluationContext is not added with OpenFeatureBuilder
### Observed behavior

Calling `OpenFeatureBuilder.AddContext()` after you have added any providers causes the `EvaluationContext `to not be injected into the Api instance.

### Expected Behavior

Calling `OpenFeatureBuilder.AddContext()` any place on the builder should add the `EvaluationContext` to the Api.

### Steps to reproduce

Example code:

```csharp
builder.Services.AddOpenFeature(featureBuilder =>
{
    featureBuilder
        .AddHostedFeatureLifecycle()
        .AddInMemoryProvider("InMemory", _ => new Dictionary<string, Flag>()
        {
            {
                "welcome-message", new Flag<bool>(
                    new Dictionary<string, bool> { { "show", true }, { "hide", false } }, "show")
            }
        })
        .AddContext(ctx => ctx.Set("region", "euw"));
});
```

Recording of the bug in action using a LoggingHook to read out the injected `EvaluationContext`.

https://github.com/user-attachments/assets/779c352d-5257-46be-a7bc-423ba9974d07

kylejuliandev:
## This PR

Fixes an issue with the `EvaluationContext` not being added to the API FeatureClient if a call to `AddContext()` was made after adding the Provider. For example, the following would fail to inject the `EvaluationContext`:

```csharp
builder.Services.AddOpenFeature(featureBuilder =>
{
    featureBuilder
        .AddInMemoryProvider(_ => new Dictionary<string, Flag>()
        {
            {
                "welcome-message", new Flag<bool>(
                    new Dictionary<string, bool> { { "show", true }, { "hide", false } }, "show")
            }
        })
        .AddContext(ctx => ctx.Set("region", "euw"));
});
```

The fix is not ported to the old Dependency Injection code as we are actively deprecating it.

### Related Issues

Fixes #542

#### Before
Screenshot 2025-10-20 201025

#### After
image

### DIFF
--- a/src/OpenFeature.Hosting/OpenFeatureBuilder.cs
+++ b/src/OpenFeature.Hosting/OpenFeatureBuilder.cs
@@ -12,6 +12,13 @@ public class OpenFeatureBuilder(IServiceCollection services)
     public IServiceCollection Services { get; } = services;
 
     /// <summary>
+    /// Indicates whether the evaluation context has been configured.
+    /// This property is used to determine if specific configurations or services
+    /// should be initialized based on the presence of an evaluation context.
+    /// </summary>
+    public bool IsContextConfigured { get; internal set; }
+
+    /// <summary>
     /// Indicates whether the policy has been configured.
     /// </summary>
     public bool IsPolicyConfigured { get; internal set; }
--- a/src/OpenFeature.Hosting/OpenFeatureBuilderExtensions.cs
+++ b/src/OpenFeature.Hosting/OpenFeatureBuilderExtensions.cs
@@ -40,6 +40,7 @@ public static partial class OpenFeatureBuilderExtensions
         Guard.ThrowIfNull(builder);
         Guard.ThrowIfNull(configure);
 
+        builder.IsContextConfigured = true;
         builder.Services.TryAddTransient(provider =>
         {
             var contextBuilder = EvaluationContext.Builder();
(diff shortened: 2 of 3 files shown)


src/OpenFeature.Hosting/OpenFeatureBuilder.cs
```diff
@@ -11,13 +11,6 @@ public class OpenFeatureBuilder(IServiceCollection services)
     /// <summary> The services being configured. </summary>
     public IServiceCollection Services { get; } = services;
 
-    /// <summary>
-    /// Indicates whether the evaluation context has been configured.
-    /// This property is used to determine if specific configurations or services
-    /// should be initialized based on the presence of an evaluation context.
-    /// </summary>
-    public bool IsContextConfigured { get; internal set; }
-
```
askpt, line 20: Is this a potential breaking change? It seems to alter the public API that we expose.

kylejuliandev: Yep, whoops! I got carried away and removed too much! I've restored this in [cef609a](https://github.com/open-feature/dotnet-sdk/pull/605/commits/cef609a7de56dd22cb3c4bb17abdf44d0971b0d5)

askpt (approved):

PR #740, opened 2026-04-12
feat: Add isolated API instance functionality

### LINKED ISSUE #734: Support isolated API instances
The spec now supports creating independent, non-singleton instances of the OpenFeature API. Each instance has its own providers, context, hooks, and events; totally separate from the global singleton and from each other. This is useful for things like DI frameworks and testing.

### What needs to happen

- Add a factory function (or equivalent) that creates new, independent API instances
- Each instance should conform to the same API contract as the existing singleton
- Ideally the factory lives in its own module/package for discoverability (recommended, not required)
- A single provider instance shouldn't be bound to more than one API instance at a time

### Links

- Parent issue: https://github.com/open-feature/spec/issues/374
- Spec section: https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances

askpt:
Signed-off-by: André Silva <2493377+askpt@users.noreply.github.com>

## This PR

This pull request introduces support for isolated API instances in the OpenFeature .NET SDK, enabling the creation of independent API objects that do not share state with the global singleton. This is particularly useful for testing, multi-tenant applications, and dependency injection scenarios. The changes also enforce provider ownership rules to prevent a provider from being registered with more than one API instance at a time, and update documentation and tests to demonstrate and verify this new functionality.

**Isolated API Instance Support:**

* Added a new static method `Api.CreateIsolated()` that creates a fully independent `Api` instance, with its own providers, context, hooks, and event handlers. These instances do not share state with the global singleton or with each other.
* Enforced provider ownership by adding a `_boundApiInstance` field to `FeatureProvider` and validating that a provider is not registered with more than one API instance at a time. Attempts to do so will throw an `InvalidOperationException`. Ownership is cleared when a provider is shut down. [[1]](diffhunk://#diff-96ebc8fc507d0a19d55b9a5cb57b72a0e8058e09f31ee7d0b39e99b00c5029d8R101-R107) [[2]](diffhunk://#diff-dc150fbd3b7be797470374ee7e38c7ab8a31eb9b6b7a52345e05114c2d32a15eR77) [[3]](diffhunk://#diff-dc150fbd3b7be797470374ee7e38c7ab8a31eb9b6b7a52345e05114c2d32a15eR113) [[4]](diffhunk://#diff-dc150fbd3b7be797470374ee7e38c7ab8a31eb9b6b7a52345e05114c2d32a15eR424-R440) [[5]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bR196-R202) [[6]](diffhunk://#diff-0c3f76ece1b10ab5874b65fadf5eabfe902e2fdac76db734ae03a17a4d9b958bR280-R285)

**Documentation and Sample Updates:**

* Updated `samples/Console/README.md` and `samples/Console/app.cs` to demonstrate the creation and use of isolated API instances, including example flag definitions and evaluation. [[1]](diffhunk://#diff-adee2b65df9cf9200d0ca9dedea6a718db206693fc5ef43af8cb008e8b6bcb37R46-R56) [[2]](diffhunk://#diff-e56b0bb1c7c90622236610cfe9c9c531b8f64da4b4f844c07152e539cf3f8a2eR47-R75)

**Testing Enhancements:**

* Added a new test `TestIsolatedApiAsync()` in `test/OpenFeature.AotCompatibility/Program.cs` to verify the creation, independence, and shutdown of isolated API instances. [[1]](diffhunk://#diff-2385141758cef849759e29b09a7c8507a8f0c46015f7487aa41a9789aef8bb02R30-R32) [[2]](diffhunk://#diff-2385141758cef849759e29b09a7c8507a8f0c46015f7487aa41a9789aef8bb02R55-R89)

### Related Issues

Fixes #734

### DIFF
--- a/samples/Console/app.cs
+++ b/samples/Console/app.cs
@@ -44,3 +44,32 @@ Console.WriteLine("The `float-flag` flag returned {0}", floatResult);
 // Evaluate the `object-flag` flag and print the result to the console
 var objectResult = await client.GetObjectValueAsync("object-flag", new Value("Ben"));
 Console.WriteLine("The `object-flag` flag returned {0}", objectResult.AsString);
+
+// --- Isolated API Instance ---
+// Create an independent API instance with its own provider and state.
+// This is useful for testing, multi-tenant apps, and DI scenarios.
+
+var isolatedFlags = new Dictionary<string, Flag>
+{
+    { "bool-flag", new Flag<bool>(new Dictionary<string, bool> { { "on", true }, { "off", false } }, defaultVariant: "off") },
+    { "string-flag", new Flag<string>(new Dictionary<string, string> { { "greeting", "Howdy, Isolated World!" }, { "farewell", "See ya!" } }, defaultVariant: "greeting") },
+};
+
+var isolated = Api.CreateIsolated();
+await isolated.SetProviderAsync(new InMemoryProvider(isolatedFlags));
+
+var isolatedClient = isolated.GetClient();
+
+Console.WriteLine();
+Console.WriteLine("--- Isolated API Instance ---");
+
+// The isolated instance has its own flag values
+var isolatedBool = await isolatedClient.GetBooleanValueAsync("bool-flag", true);
+Console.WriteLine("Isolated `bool-flag`: {0} (singleton was: {1})", isolatedBool, helloWorldResult);
+
+var isolatedString = await isolatedClient.GetStringValueAsync("string-flag", "default");
+Console.WriteLine("Isolated `string-flag`: {0} (singleton was: {1})", isolatedString, stringResult);
+
+// Shut down the isolated instance — does not affect the global singleton
+await isolated.ShutdownAsync();
+Console.WriteLine("Isolated instance shut down. Singleton is unaffected.");
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -35,24 +35,6 @@ public sealed class Api : IEventBus
     internal Api() { }
 
     /// <summary>
-    /// Creates a new, independent instance of the OpenFeature API with fully isolated state.
-    /// <para>
-    /// Each isolated instance maintains its own providers, evaluation context, hooks, event handlers,
-    /// and transaction context propagators. It does not share state with the global <see cref="Instance"/>
-    /// singleton or with any other isolated instance.
-    /// </para>
-    /// </summary>
-    /// <remarks>
-    /// <para>
-    /// A single provider instance should not be bound to more than one API instance at a time.
-    /// Attempting to do so will result in an <see cref="InvalidOperationException"/>.
-    /// </para>
-    /// </remarks>
-    /// <returns>A new, independent <see cref="Api"/> instance.</returns>
-    /// <seealso href="https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances">Specification 1.8 - Isolated API Instances</seealso>
-    public static Api CreateIsolated() => new Api();
-
-    /// <summary>
     /// Sets the default feature provider. In order to wait for the provider to be set, and initialization to complete,
     /// await the returned task.
     /// </summary>
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -424,14 +424,12 @@ public sealed class Api : IEventBus
 
     /// <summary>
     /// Validates that the given provider is not already bound to a different API instance.
-    /// Uses <see cref="Interlocked.CompareExchange{T}"/> for thread-safe check-and-set.
     /// </summary>
     /// <param name="featureProvider">The provider to validate ownership for.</param>
     /// <exception cref="InvalidOperationException">Thrown if the provider is already bound to a different API instance.</exception>
     private void ValidateProviderOwnership(FeatureProvider featureProvider)
     {
-        var previous = Interlocked.CompareExchange(ref featureProvider._boundApiInstance, this, null);
-        if (previous is not null && !ReferenceEquals(previous, this))
+        if (!featureProvider.TryBindApiInstance(this))
         {
             throw new InvalidOperationException(
                 "This provider instance is already bound to a different API instance. " +
--- a/src/OpenFeature/Api.cs
+++ b/src/OpenFeature/Api.cs
@@ -35,6 +35,24 @@ public sealed class Api : IEventBus
     internal Api() { }
 
     /// <summary>
+    /// Creates a new, independent instance of the OpenFeature API with fully isolated state.
+    /// <para>
+    /// Each isolated instance maintains its own providers, evaluation context, hooks, event handlers,
+    /// and transaction context propagators. It does not share state with the global <see cref="Instance"/>
+    /// singleton or with any other isolated instance.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// A single provider instance should not be bound to more than one API instance at a time.
+    /// Attempting to do so will result in an <see cref="InvalidOperationException"/>.
+    /// </para>
+    /// </remarks>
+    /// <returns>A new, independent <see cref="Api"/> instance.</returns>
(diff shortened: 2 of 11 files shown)


test/OpenFeature.Tests/IsolatedApiTests.cs
```diff
@@ -0,0 +1,368 @@
...
+        const int instanceCount = 10;
+        var provider = new TestProvider("contested-provider");
+        var instances = Enumerable.Range(0, instanceCount).Select(_ => Api.CreateIsolated()).ToList();
+        try
+        {
+            var tasks = instances.Select(api => api.SetProviderAsync(provider)).ToList();
+            var results = await Task.WhenAll(tasks.Select(async t =>
+            {
+                try { await t; return (Exception?)null; }
+                catch (Exception ex) { return ex; }
```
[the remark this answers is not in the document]
askpt: This is just a sample test case.

src/OpenFeature/Api.cs
```diff
@@ -401,4 +421,21 @@ internal static void SetInstance(Api api)
...
+
+    /// <summary>
+    /// Validates that the given provider is not already bound to a different API instance.
+    /// Uses <see cref="Interlocked.CompareExchange{T}"/> for thread-safe check-and-set.
+    /// </summary>
+    /// <param name="featureProvider">The provider to validate ownership for.</param>
+    /// <exception cref="InvalidOperationException">Thrown if the provider is already bound to a different API instance.</exception>
+    private void ValidateProviderOwnership(FeatureProvider featureProvider)
+    {
+        var previous = Interlocked.CompareExchange(ref featureProvider._boundApiInstance, this, null);
```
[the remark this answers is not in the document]
askpt: This is wrong. The library is compliant with null checks by default. As pointed out in the remarks, the provider cannot be set to null.

src/OpenFeature/Api.cs
```diff
@@ -34,6 +34,24 @@ public sealed class Api : IEventBus
...
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// A single provider instance should not be bound to more than one API instance at a time.
+    /// Attempting to do so will result in an <see cref="InvalidOperationException"/>.
+    /// </para>
+    /// </remarks>
+    /// <returns>A new, independent <see cref="Api"/> instance.</returns>
+    /// <seealso href="https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances">Specification 1.8 - Isolated API Instances</seealso>
+    public static Api CreateIsolated() => new Api();
```
kylejuliandev: question: What's the advantage of having a static method to initialize a new Api instance over just removing the private constructor?

test/OpenFeature.Tests/Isolated/IsolatedApiTests.cs
```diff
@@ -0,0 +1,368 @@
...
+    public void Isolated_Hooks_Should_Not_Affect_Other_Instances()
+    {
+        var isolated1 = Api.CreateIsolated();
+        var isolated2 = Api.CreateIsolated();
+
+        var hook1 = Substitute.For<Hook>();
+        var hook2 = Substitute.For<Hook>();
+
+        isolated1.AddHooks(hook1);
+        isolated2.AddHooks(hook2);
```
kylejuliandev, line 135: question: Can you add the same hook to multiple Providers (and multiple Api instances)?

test/OpenFeature.Tests/IsolatedApiTests.cs
```diff
@@ -0,0 +1,368 @@
...
+    public async Task Same_Provider_Should_Throw_When_Bound_To_Multiple_Api_Instances_Domain()
+    {
+        var isolated1 = Api.CreateIsolated();
+        var isolated2 = Api.CreateIsolated();
+        try
+        {
+            var provider = new TestProvider("shared-provider");
+            await isolated1.SetProviderAsync("domain-1", provider);
+
+            await Assert.ThrowsAsync<InvalidOperationException>(
```
kylejuliandev: suggestion: Is it worth while adding an assertion to check the Exception message?

kylejuliandev (approved):

test/OpenFeature.Tests/Isolated/IsolatedApiTests.cs
askpt: I don't remember any restrictions about it in the spec.

src/OpenFeature/Api.cs
askpt: The static method is there since it matches the pattern used by the spec and the name makes the intent explicit.

test/OpenFeature.Tests/IsolatedApiTests.cs
askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/740/commits/cdb1050024f0283206a5b4a777c77d8d2f378d58

src/OpenFeature/ProviderRepository.cs
```diff
@@ -193,6 +193,13 @@ private async Task ShutdownIfUnusedAsync(
             return;
         }
 
+        // Clear ownership while still under the write lock — the provider is confirmed unused.
```
kinyoklion, line 196: It feels like the setting and clearing could be internal methods on the FeatureProvider instead of just exposing the reference. For example `internal bool FeatureProvider.TrySetApiInstance(Api api)` and `internal void FeatureProvider.ClearApiInstance()`. Then it could privately use either a mutex or interlocked methods instead of the ProviderRepository and Api sharing this logic.

askpt: Great suggestion! Fixed with: https://github.com/open-feature/dotnet-sdk/pull/740/commits/3da0b3f92a3a943122390eea2458ce54a2553acd

toddbaert:
Nice work on this! One suggestion: the spec recommends (1.8.3, SHOULD) that the factory function live in a distinct module for discoverability. Right now `CreateIsolated()` is a static method on `Api` itself.

The SDK already uses sub-namespaces (`OpenFeature.Hooks`, `OpenFeature.Error`, etc.), so something like this would be consistent:

```csharp
namespace OpenFeature.Isolated;

public static class OpenFeatureFactory
{
    public static Api CreateApi() => new Api();
}
```

This way users opt in with `using OpenFeature.Isolated;` and the factory is easy to find. WDYT?

cc @kylejuliandev @kinyoklion

src/OpenFeature/Api.cs
toddbaert: Also, 1.8 is experimental in the spec, so this should probably get an `[Experimental]` attribute. If you go with the separate namespace approach, you could slap it on the class; otherwise just on `CreateIsolated()` itself.

askpt: Fixed with: https://github.com/open-feature/dotnet-sdk/pull/740/commits/be1b1ba3c23dc3dc6c7bb00688c448589d91d080

toddbaert (approved):
LGTM (except the .NET format CI) ❤️ @askpt .

samples/Console/app.cs
```diff
@@ -44,3 +44,32 @@
...
+// Create an independent API instance with its own provider and state.
+// This is useful for testing, multi-tenant apps, and DI scenarios.
+
+var isolatedFlags = new Dictionary<string, Flag>
+{
+    { "bool-flag", new Flag<bool>(new Dictionary<string, bool> { { "on", true }, { "off", false } }, defaultVariant: "off") },
+    { "string-flag", new Flag<string>(new Dictionary<string, string> { { "greeting", "Howdy, Isolated World!" }, { "farewell", "See ya!" } }, defaultVariant: "greeting") },
+};
+
+var isolated = Api.CreateIsolated();
```
toddbaert, line 58: Thanks for addressing the earlier comments; the namespace move and `[Experimental]` attribute look great! If I'm not mistaken though, this spot seems to have been missed; `Api.CreateIsolated()` no longer exists, so I think this should be `OpenFeatureFactory.CreateIsolated()` (with a `using OpenFeature.Isolated;` added at the top).

askpt: ...

Good catch...

askpt: https://github.com/open-feature/dotnet-sdk/pull/756

PR #743, opened 2026-04-16
fix: fix 'occured' -> 'occurred' in XML doc comment

SAY-5:
XML doc comment in `src/OpenFeature/Constant/ErrorType.cs` line 12 reads `no error occured`. Fixed to `occurred`. Comment-only change.


askpt:
@SAY-5 Thank you for your contribution. We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) in the titles of our PRs that will generate a line in the changelog. I will be editing it now so it will pass the action check.

askpt (approved):

kylejuliandev (approved):

PR #755, opened 2026-05-07
feat: Add trace when evaluating feature flags

### LINKED ISSUE #541: Update the trace support
Currently in the `TraceEnricherHook`, we're adding tags/events to the current activity,

https://github.com/open-feature/dotnet-sdk/blob/417f3fefbafc656f64069b284f741081b1d77113/src/OpenFeature/Hooks/TraceEnricherHook.cs#L39-L48

In this way, it's hard to listen to the event specifically, since the event is not determined, and sometimes there's no activity.

From the practice on docs, we should create an individual `ActivitySource` like the `Meter`, so that we could listen to the `ActivitySource` when we want to focus on the feature flag activity

https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs#best-practices

Another thought on the activity, we should try to avoid unnecessary operations to improve performance, for example, we could skip creating and adding tags/events when there's no activity or the `Activity.IsAllDataRequested` property is `false`

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.isalldatarequested

Finally, since we're creating a separate `ActivitySource`, think it could be built-in the main logic other than a hook, the users would need to listen to the `ActivitySource` explicitly to get related activities, so that no need to register additional trace hook.

### LINKED ISSUE #595: [FEATURE] Add support for spans instead of span events
### Requirements

Currently, the `TraceEnricherHook` hook creates a span event on the active span. This is in line with the OpenTelemetry semantic convention. However, in some cases, it's preferred to have a span instead of a span event. The hook should take an optional configuration that allows users to opt into a span instead of a span event. The default behavior should remain a span event but this would provide additional flexibility without minimal maintenance effort.

### Proposal

- Add an option to the `TraceEnricherHook` hook that allows users to choose between span events or spans. It must default to span events if unspecified.
- If spans are selected:
  - a span is started in the before hook and passed via hook data.
  - the span is fetched from hook data, attributes are added, and the span is closed
- If span events are selected:
  - a span is NOT created in the before hook
  - a span event is added to the current span

kylejuliandev:
## This PR

Add trace when evaluating feature flags. This is different from the `TraceEnricherHook` which enriches the current span or trace with an event, instead this more closely follows the semantic conventions: https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-events/. Also add OpenTelemetry logging to the sample application and ensure traces are always sampled.

> [!WARNING]
> This release adds native OpenTelemetry spans emitted on every flag evaluation — no hooks required. Users already using `TraceEnricherHook` should expect **new spans** to appear in their observability platform after upgrading. Use `TraceEnricherHook` when you want spans/activities *outside* of OpenFeature to be enriched with flag evaluation data.

#### Success trace

image

#### Error trace

image

### Related Issues

Fixes #595
Fixes #541

## Summary by CodeRabbit

* **New Features**
  * Flag evaluations now emit native OpenTelemetry spans with detailed outcome tags (reason/value/variant) on success and error tags (type + message) on failures.
  * The ASP.NET Core sample now includes OpenTelemetry logging setup and registers the OpenFeature trace source.
* **Documentation**
  * README updated for native span emission behavior (v2.14), including when `TraceEnricherHook` is needed and how to enable native tracing via `AddSource("OpenFeature")`.
* **Tests**
  * Added coverage for span creation, tag enrichment, and error tagging.


src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -211,6 +211,15 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
         var resolveValueDelegate = providerInfo.Item1;
         var provider = providerInfo.Item2;
 
+        var activity = OpenFeatureActivitySource.StartActivity("feature_flag.evaluation");
+        activity?.SetTag("feature_flag.key", flagKey);
+
+        var providerMetadata = provider.GetMetadata();
+        if (providerMetadata != null)
+        {
+            activity?.SetTag("feature_flag.provider.name", providerMetadata.Name);
```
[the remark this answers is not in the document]
kylejuliandev: Is it `feature_flag.provider.name` in the semantic conventions

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -261,8 +270,13 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
                     .ConfigureAwait(false))
                 .ToFlagEvaluationDetails();
 
+            activity?.SetTag("feature_flag.result.reason", OpenFeatureActivitySource.GetFlagEvaluationReasonDescription(evaluation.Reason));
```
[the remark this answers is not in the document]
kylejuliandev: It is `feature_flag.result.reason` in the conventions

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -261,8 +270,13 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
                     .ConfigureAwait(false))
                 .ToFlagEvaluationDetails();
 
+            activity?.SetTag("feature_flag.result.reason", OpenFeatureActivitySource.GetFlagEvaluationReasonDescription(evaluation.Reason));
+
             if (evaluation.ErrorType == ErrorType.None)
             {
+                activity?.SetTag("feature_flag.result.value", evaluation.Value);
+                activity?.SetTag("feature_flag.result.variant", evaluation.Variant);
```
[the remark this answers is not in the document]
kylejuliandev: These are correct already

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -290,6 +311,10 @@ await hookRunner.TriggerErrorHooksAsync(ex, options?.HookHints, cancellationToke
             var errorCode = ex is InvalidCastException ? ErrorType.TypeMismatch : ErrorType.General;
             evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, errorCode, Reason.Error, string.Empty,
                 ex.Message);
+
+            activity?.AddTag("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
+            activity?.SetTag("feature_flag.error.message", evaluation.ErrorMessage);
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [3dce87e](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3dce87ecf1cb8fccff28ef7824d2c0afefe7cd1b)

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -282,6 +299,10 @@ await hookRunner.TriggerErrorHooksAsync(exception, options?.HookHints, cancellat
             this.FlagEvaluationErrorWithDescription(flagKey, ex.ErrorType.GetDescription(), ex);
             evaluation = new FlagEvaluationDetails<T>(flagKey, defaultValue, ex.ErrorType, Reason.Error,
                 string.Empty, ex.Message);
+
+            activity?.AddTag("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
+            activity?.SetTag("feature_flag.error.message", evaluation.ErrorMessage);
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [3dce87e](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3dce87ecf1cb8fccff28ef7824d2c0afefe7cd1b)

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -271,6 +285,9 @@ await hookRunner.TriggerAfterHooksAsync(
             }
             else
             {
+                activity?.AddTag("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
+                activity?.SetTag("feature_flag.error.message", evaluation.ErrorMessage);
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [3dce87e](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3dce87ecf1cb8fccff28ef7824d2c0afefe7cd1b)

src/OpenFeature/OpenFeatureActivitySource.cs
```diff
@@ -0,0 +1,48 @@
...
+            Reason.Cached => "cached",
+            Reason.Unknown => "unknown",
+            Reason.Error => "error",
+            _ => reason
+        };
+
+    static string GetLibraryVersion()
+        => typeof(OpenFeatureActivitySource).Assembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?
+            .InformationalVersion ?? "UNKNOWN";
```
askpt: Could you please remove the `+<sha>` in the version. You can see it in the screenshots you posted.

Also I wonder if we should "cache" the version string somewhere to prevent doing `GetCustomAttribute` all the time.

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -211,6 +212,15 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
         var resolveValueDelegate = providerInfo.Item1;
         var provider = providerInfo.Item2;
 
+        var activity = OpenFeatureActivitySource.StartActivity("feature_flag.evaluation");
```
askpt: Wonder if we should set it as `Client`. See: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitykind?view=net-10.0

askpt (commented):
LGTM! I expect this change to be mentioned in a README since the consumer might need to add extra configuration to remove or add these traces.

cc: @beeme1mr, @open-feature/technical-steering-committee, @WeihanLi, @arttonoyan

src/OpenFeature/OpenFeatureClient.cs
WeihanLi: > Wonder if we should set it as Client

personally think it's fine to keep it using default `Internal`

WeihanLi:
```suggestion
        using var activity = OpenFeatureActivitySource.StartActivity("feature_flag.evaluation");
```

maybe we could add `using` for this

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -211,6 +212,15 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
         var resolveValueDelegate = providerInfo.Item1;
         var provider = providerInfo.Item2;
 
+        var activity = OpenFeatureActivitySource.StartActivity("feature_flag.evaluation");
+        activity?.SetTag("feature_flag.key", flagKey);
```
WeihanLi: for the `SetTag`/`SetStatus`, think we could add tags in one place and add tags only when `IsAllDataRequested`

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.isalldatarequested?view=net-10.0

WeihanLi:
maybe we want to deprecate the `TraceEnricherHook` related code also?

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -211,6 +212,15 @@ private async Task<FlagEvaluationDetails<T>> EvaluateFlagAsync<T>(
         var resolveValueDelegate = providerInfo.Item1;
         var provider = providerInfo.Item2;
 
+        var activity = OpenFeatureActivitySource.StartActivity("feature_flag.evaluation");
+        activity?.SetTag("feature_flag.key", flagKey);
+
+        var providerMetadata = provider.GetMetadata();
+        if (providerMetadata != null)
+        {
+            activity?.SetTag("feature_flag.provider.name", providerMetadata.Name);
```
WeihanLi: Would we like to keep these tagName/activityName to be constants in a separated place?

samples/AspNetCore/Program.cs
```diff
@@ -25,10 +26,23 @@
...
+
+    options.IncludeScopes = true;
+    options.IncludeFormattedMessage = true;
+});
+
 builder.Services.AddOpenTelemetry()
     .ConfigureResource(resource => resource.AddService("openfeature-aspnetcore-sample"))
     .WithTracing(tracing => tracing
         .AddAspNetCoreInstrumentation()
+        .SetSampler(new AlwaysOnSampler())
```
WeihanLi: Maybe we should add `AddSource("OpenFeature")` to collect OpenFeature activity?

src/OpenFeature/OpenFeatureActivitySource.cs
kylejuliandev: The `ActivitySource` is static, so it's initialized once during the first flag evaluation and then subsequent evaluations don't fetch a new version

samples/AspNetCore/Program.cs
kylejuliandev: Added trace source in [3b8eac5](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3b8eac5033a072c5888a54f9a7aa096f4c841908)

src/OpenFeature/OpenFeatureClient.cs
kylejuliandev: Fixed in [6e943fa](https://github.com/open-feature/dotnet-sdk/pull/755/commits/6e943fad17a23a4ad23fdccd4df63cc6e4c9a7aa) - moved the tag names (apart from error.type) to the OpenFeatureActivitySource class

src/OpenFeature/OpenFeatureClient.cs
kylejuliandev: Fixed in [6e943fa](https://github.com/open-feature/dotnet-sdk/pull/755/commits/6e943fad17a23a4ad23fdccd4df63cc6e4c9a7aa) - this was a great spot! It helps avoid allocations in high performance scenarios

I wonder if we should add `feature_flag.provider.name` in all instances or only when `IsAllDataRequested` is true

src/OpenFeature/OpenFeatureClient.cs
kylejuliandev: Great spot, this was probably leaving activity sources undisposed of (potentially leaving resources hanging). Fixed in [3b8eac5](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3b8eac5033a072c5888a54f9a7aa096f4c841908)

src/OpenFeature/OpenFeatureActivitySource.cs
kylejuliandev: I've opted to use AssemblyVersion instead of the InformationalVersion in [3b8eac5](https://github.com/open-feature/dotnet-sdk/pull/755/commits/3b8eac5033a072c5888a54f9a7aa096f4c841908)

In `Common.prod.props` we set the Assembly version

https://github.com/open-feature/dotnet-sdk/blob/b3fa7f41c926a93a69ac2d599089ce4045d62770/build/Common.prod.props#L22-L24

I do a `ToString` on the Version, so we get `major.minor.patch` in the trace:

Screenshot 2026-05-14 200334

src/OpenFeature/OpenFeatureClient.cs
kylejuliandev: I think if the Provider (InMemory, Flagd, etc..) traces, they may opt to use `Client` in this example. I figured here at this point we don't necessarily know if the flag evaluation is remote. But we are confident we are delegating the actual evaluation to another component that is running in the same process

src/OpenFeature/OpenFeatureActivitySource.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            Reason.TargetingMatch => "targeting_match",
+            Reason.Split => "split",
+            Reason.Disabled => "disabled",
+            Reason.Default => "default",
+            Reason.Static => "static",
+            Reason.Cached => "cached",
+            Reason.Unknown => "unknown",
+            Reason.Error => "error",
+            _ => reason
```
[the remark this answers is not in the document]
kylejuliandev: Leaving the casing as this may be specified by the provider (if using non standard reason codes).

src/OpenFeature/OpenFeatureActivitySource.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    internal static Activity? StartActivity(string name)
+        => Source.StartActivity(name, ActivityKind.Internal);
+
+    internal const string EvaluationActivityName = "feature_flag.evaluation";
+    internal const string FeatureFlagKeyName = "feature_flag.key";
+    internal const string FeatureFlagProviderName = "feature_flag.provider.name";
+    internal const string FeatureFlagReasonName = "feature_flag.result.reason";
+    internal const string FeatureFlagValueName = "feature_flag.result.value";
+    internal const string FeatureFlagVariantName = "feature_flag.result.variant";
```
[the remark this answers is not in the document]
kylejuliandev: This is already correct and aligned.

src/OpenFeature/OpenFeatureActivitySource.cs
```diff
@@ -0,0 +1,60 @@
...
+static class OpenFeatureActivitySource
+{
+    static readonly ActivitySource Source = new("OpenFeature", GetLibraryVersion());
+
+    internal static Activity? StartActivity(string name)
+        => Source.StartActivity(name, ActivityKind.Internal);
+
+    internal const string EvaluationActivityName = "feature_flag.evaluation";
+    internal const string FeatureFlagKeyName = "feature_flag.key";
+    internal const string FeatureFlagProviderName = "feature_flag.provider.name";
```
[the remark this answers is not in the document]
kylejuliandev: This is already correct and aligned.

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -302,6 +353,8 @@ await hookRunner.TriggerFinallyHooksAsync(evaluation, options?.HookHints, cancel
                 .ConfigureAwait(false);
         }
 
+        activity?.Dispose();
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [6d64f4f](https://github.com/open-feature/dotnet-sdk/pull/755/commits/6d64f4f0e2ad91db862eb98abe5a024e10781504)

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -271,6 +296,14 @@ await hookRunner.TriggerAfterHooksAsync(
             }
             else
             {
+                activity?.SetStatus(ActivityStatusCode.Error);
+
+                if (activity?.IsAllDataRequested == true)
+                {
+                    activity.AddTag("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
+                    activity.SetTag(OpenFeatureActivitySource.FeatureFlagErrorMessageName, evaluation.ErrorMessage);
+                }
```
WeihanLi: seems there's some duplication, maybe we could add one method to avoid duplicates

```cs
private static void AttachActivityTags(Activity? activity, FlagEvaluationDetails<T>? evaluation, Exception? exception = null)
{
    if (activity is null) return;
    if (elaluation?.ErrorType != ErrorType.None)
        activity.SetStatus(ActivityStatusCode.Error);
    
    if (activity.IsAllDataRequested != true) return;

    if (elaluation is null)
    {
        // ...
        return;
    }

    activity.SetTag(OpenFeatureActivitySource.FeatureFlagReasonName, OpenFeatureActivitySource.GetFlagEvaluationReasonDescription(evaluation.Reason));
    if (elaluation?.ErrorType == ErrorType.None)
    {
        activity.SetTag(OpenFeatureActivitySource.FeatureFlagValueName, evaluation.Value);
        activity.SetTag(OpenFeatureActivitySource.FeatureFlagVariantName, evaluation.Variant);
    }
    else
    {
        activity.AddTag("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
        activity.SetTag(OpenFeatureActivitySource.FeatureFlagErrorMessageName, evaluation.ErrorMessage);
    }
}

```

src/OpenFeature/OpenFeatureClient.cs
askpt: Fair. Thanks for the discussion. I am happy to keep using `Internal`. Like @kylejuliandev mentioned, the providers might prefer to use `Client` making it a bit awkward for the visualisation.

askpt:
> maybe we want to deprecate the `TraceEnricherHook` related code also?

Not at this point. Until OTEL formally deprecates `events` in traces, I would keep this `Hook` enabled since they have different behaviour.

src/OpenFeature/OpenFeatureClient.cs
askpt: Actually with all `IsAllDataRequested` I wonder if makes sense to have an internal extension method to have the `AddTag` and `SetTag`. Like `AddTagIfRequested` or `SetTagIfRequested`. It would make the code cleaner IMO.

What do you think @WeihanLi and @kylejuliandev?

kylejuliandev: I think I prefer the extension method approach. It keeps activity enrichment close to the code but simplifies the repeated if IsAllDataRequested checks

src/OpenFeature/OpenFeatureActivitySource.cs
```diff
@@ -0,0 +1,71 @@
...
+            .Version;
+
+        // "3" = major.minor.patch only
+        return version?.ToString(3) ?? "UNKNOWN";
+    }
+
+    internal static void AddTagIfRequested(this Activity activity, string tagName, object? value)
+    {
+        if (activity == null)
+            return;
```
askpt: Why do you need to check if `activity` is null? I believe we have nullables enabled in this project 🤔

src/OpenFeature/OpenFeatureClient.cs
```diff
@@ -271,6 +287,10 @@ await hookRunner.TriggerAfterHooksAsync(
             }
             else
             {
+                activity?.SetStatus(ActivityStatusCode.Error);
+                activity?.AddTagIfRequested("error.type", OpenFeatureActivitySource.GetFlagEvaluationErrorDescription(evaluation.ErrorType));
```
askpt: Could you add `error.type` as a constant since it happens twice in this class?

askpt (commented):
Added two minor comments.

src/OpenFeature/OpenFeatureActivitySource.cs
kylejuliandev: I think I was being over cautious here, you're right with nullable's enabled we should be confident this isn't null. Fixed in [06fc54b](https://github.com/open-feature/dotnet-sdk/pull/755/commits/06fc54bce33b23dd290a559e3411d9a91bdb4e9d)

src/OpenFeature/OpenFeatureClient.cs
kylejuliandev: Moved the `error.type` string to a constant in [06fc54b](https://github.com/open-feature/dotnet-sdk/pull/755/commits/06fc54bce33b23dd290a559e3411d9a91bdb4e9d)

askpt (approved):

benjiro (approved):

src/OpenFeature/OpenFeatureClient.cs
WeihanLi: think it's more cleaner to introduce extension method for `AddTagIfRequested`/`SetTagIfRequested`, but it may cause some duplicated not null and `IsAllDataRequested` checks, maybe it should be avoid for a hotpath.
While maybe we could evaluate this further if someone complains about this.

test/OpenFeature.Tests/OpenFeatureClientTracingTests.cs
```diff
@@ -0,0 +1,225 @@
...
+        var trace = this._exportedActivities[0];
+        Assert.Equal(ActivityStatusCode.Error, trace.Status);
+
+        var tags = trace.TagObjects.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+        Assert.Contains("error.type", tags);
+        Assert.Equal("provider_fatal", tags["error.type"]);
+
+        Assert.Contains("feature_flag.error.message", tags);
+        Assert.Equal("Error!", tags["feature_flag.error.message"]);
+    }
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [1ba2ef9](https://github.com/open-feature/dotnet-sdk/pull/755/commits/1ba2ef9cb287fe8095179ee8b5bbe01e883acb06)

Not every path returns `feature_flag.result.reason`

test/OpenFeature.Tests/OpenFeatureActivitySourceTests.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        var activity = OpenFeatureActivitySource.StartActivity("test_activity");
+
+        Assert.NotNull(activity);
+        Assert.Equal("test_activity", activity.OperationName);
+        Assert.Equal("OpenFeature", activity.Source.Name);
+        Assert.NotNull(activity.Source.Version);
+        Assert.NotEmpty(activity.Source.Version);
+        Assert.Equal(ActivityKind.Internal, activity.Kind);
+    }
```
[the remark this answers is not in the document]
kylejuliandev: Fixed in [1ba2ef9](https://github.com/open-feature/dotnet-sdk/pull/755/commits/1ba2ef9cb287fe8095179ee8b5bbe01e883acb06)

PR #439, opened 2025-04-16
chore(workflows): Add permissions for contents and pull-requests

askpt:
## This PR

This pull request includes updates to several GitHub Actions workflows to add permissions for reading contents and writing pull requests. Additionally, there is a minor change to the `lint-pr.yml` workflow file to standardize the quotation marks used in the `name` field.

Workflow permissions updates:

* [`.github/workflows/ci.yml`](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR15-R17): Added permissions for reading contents and writing pull requests to the `build` job.
* [`.github/workflows/code-coverage.yml`](diffhunk://#diff-49708f979e226a1e7bd7a68d71b2e91aae8114dd3e9254d9830cd3b4d62d4303R15-R17): Added permissions for reading contents and writing pull requests to the `build-test-report` job.
* [`.github/workflows/dco-merge-group.yml`](diffhunk://#diff-cbf8f01aa06b4aa3d0729c5bce44e4f919c801b55f19a781b15f62aa10e68e90R10-R12): Added permissions for reading contents and writing pull requests to the `DCO` job.
* [`.github/workflows/dotnet-format.yml`](diffhunk://#diff-ca8c2611c79b991c0fbe04fec3c97c14dc83419f5efb1e8a7a96dd51e7df3e2aR12-R14): Added permissions for reading contents and writing pull requests to the `check-format` job.
* [`.github/workflows/e2e.yml`](diffhunk://#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR16-R18): Added permissions for reading contents and writing pull requests to the `e2e-tests` job.
* [`.github/workflows/lint-pr.yml`](diffhunk://#diff-70c3a017bfdb629fd50281fe5f7ad22e29c0ddac36e7065e9dc6d4f0924104f4R14-R16): Added permissions for reading contents and writing pull requests to the `main` job.

Standardization:

* [`.github/workflows/lint-pr.yml`](diffhunk://#diff-70c3a017bfdb629fd50281fe5f7ad22e29c0ddac36e7065e9dc6d4f0924104f4L1-R1): Changed single quotes to double quotes in the `name` field.

PR #791, opened 2026-07-29
ci: add continuous BenchmarkDotNet publishing workflow

berkslv:
## This PR

- Adds a new `.github/workflows/benchmark.yml` GitHub Actions workflow that runs the BenchmarkDotNet benchmarks in `test/OpenFeature.Benchmarks` on push to `main` (path-filtered), on a daily schedule, and via manual dispatch.
- Publishes benchmark results to the `gh-pages` branch using [`martincostello/benchmarkdotnet-results-publisher`](https://github.com/martincostello/benchmarkdotnet-results-publisher), enabling tracking of performance over time via a GitHub Pages dashboard.
- Adds `net8.0` to `OpenFeature.Benchmarks.csproj`'s `TargetFrameworks`, since `OpenFeatureClientBenchmarks` declares `[SimpleJob(RuntimeMoniker.Net80, baseline: true)]` but the project previously didn't build for `net8.0`, which caused the baseline job to silently fail in CI.

### Related Issues

Fixes [#625](https://github.com/open-feature/dotnet-sdk/issues/625)

### Notes

- Per discussion in [#625](https://github.com/open-feature/dotnet-sdk/issues/625), this starts with GitHub-hosted runners as a baseline. A `TODO` comment is left in the workflow noting that self-hosted runners can be considered later if results turn out to be too noisy.
- This PR only adds CI plumbing to run/publish existing benchmarks; it does not add or modify benchmark test cases.
- A live preview of the resulting GitHub Pages dashboard is available on my fork: **https://berkslv.github.io/dotnet-sdk/**

### Follow-up Tasks

- Consider migrating to self-hosted runners if GitHub-hosted runner noise makes results unreliable (see #625 discussion).
- Consider whether the `gh-pages` dashboard (static HTML/JS charts) should be added upstream, or left as a maintainer decision. A working preview is available on my fork: https://berkslv.github.io/dotnet-sdk/

### How to test

- I validated this end-to-end on my fork (`berkslv/dotnet-sdk`): pushed to `main`, confirmed the `Benchmark` workflow runs the benchmarks and successfully publishes `data.json` to `gh-pages`.
- Live dashboard preview (from my fork): https://berkslv.github.io/dotnet-sdk/
- To test manually: trigger the `Benchmark` workflow via `workflow_dispatch` from the Actions tab, or push a change under `src/**` or `test/OpenFeature.Benchmarks/**` to `main`.

PR #801, opened 2026-08-03
ci: gate all checks behind a single required job and fix merge queue

askpt:
## Why

Branch protection on `main` requires exactly two status checks: `DCO` and `e2e-tests`. Everything else that runs on a pull request (build, unit tests, code coverage, `dotnet format`, and the six-way AOT matrix) is advisory. A pull request with a red build merges today.

Separately, `ci.yml`, `code-coverage.yml` and `dotnet-format.yml` have no `merge_group` trigger, so the merge queue has never validated build, test or format against the merged result. Only E2E, AOT and the dummy DCO reporter run there.

This adds a single required check, `CI Gate`, that sits downstream of every other job. Once branch protection requires only `DCO` + `CI Gate`, all of those checks become effectively required, and jobs or matrix legs can be added and removed later without anyone having to remember to edit branch protection.

## Approach

`needs:` cannot reference a job in another workflow file, so every gated job has to be reachable from one place. `ci.yml` becomes a thin orchestrator and the work moves into `workflow_call` building blocks:

```
ci.yml
  build      -> reusable-build.yml
  coverage   -> reusable-coverage.yml
  format     -> reusable-format.yml
  e2e        -> reusable-e2e.yml
  aot        -> reusable-aot.yml
  packaging  -> reusable-packaging.yml
  ci-gate    -- the single required status check
```

`code-coverage.yml`, `dotnet-format.yml`, `aot-compatibility.yml` and `e2e.yml` are folded in and deleted. `lint-pr.yml` stays out of the gate on purpose: it uses `pull_request_target` and structurally cannot report on `merge_group`, so requiring it would deadlock the queue. `dco-merge-group.yml` and `release.yml` are untouched.

The blocks take no inputs. They have one caller, so parameters would only ever receive their own defaults. The Codecov token is passed explicitly because secrets are not visible to a called workflow otherwise.

A `setup-dotnet` composite action absorbs the repeated setup-dotnet plus NuGet cache blocks.

## Things fixed along the way

- **Path filtering vs required checks.** `ci.yml` and `code-coverage.yml` used workflow-level `paths-ignore: "**.md"`. A workflow skipped by path filtering reports no check runs at all, so a required check inside one would leave docs-only pull requests blocked forever. Removed.
- **The gate needs `if: always()`.** Without it the job inherits the implicit `success()` condition and is reported as *skipped* when a dependency fails, and branch protection treats a skipped check as satisfied. That would let through exactly the pull requests it is meant to stop. The predicate also fails on `cancelled` and on any skip that is not explicitly allow-listed.
- **The fork guard was always true.** `github.event.pull_request.head.repo.fork == false` evaluates to true on `push` and `merge_group`, because the missing value is coerced to 0 before the comparison. Combined with `startsWith(github.ref, 'refs/heads/')`, which also matches `refs/heads/gh-readonly-queue/*`, simply adding a `merge_group` trigger would have published a NuGet package for every queue entry. The packaging job now checks `github.event_name` explicitly, always packs as a validation step, and publishes only on `push` or a non-fork pull request.
- **Never cancel a queue run.** `cancel-in-progress` is now scoped to `pull_request`. A cancelled run reports a non-success conclusion and ejects the pull request from the queue.
- Dropped the unused `id-token: write` and `attestations: write` permissions from packaging, which has no attest step.

Coverage is skipped on `merge_group`: the `gh-readonly-queue/*` ref is thrown away, so uploading it to Codecov is meaningless, and `build` already runs the same tests there. It is the only entry in the gate's skip allow-list.

## Required before merging this

Check runs produced through a reusable workflow are named `<caller job> / <called job>`, so the `e2e-tests` context goes away with this change. A required context that stops reporting does not fail, it stays pending, which means this pull request cannot merge itself while `e2e-tests` is still required.

So, unusually, the branch protection edit has to happen **before** the merge, not after:

```bash
# 1. before merging: drop e2e-tests, leaving DCO
gh api -X PATCH repos/open-feature/dotnet-sdk/branches/main/protection/required_status_checks \
  -F strict=false -F 'checks[][context]=DCO'

# 2. after merging: add the gate
gh api -X PATCH repos/open-feature/dotnet-sdk/branches/main/protection/required_status_checks \
  -F strict=false -F 'checks[][context]=DCO' -F 'checks[][context]=CI Gate'
```

Two approvals, DCO, conversation resolution and the merge queue's ALLGREEN strategy all still apply in between, so the window is narrow.

Worth doing at the same time: lower the merge queue `checkResponseTimeout` from 3600s, which is a long time to wait for a check that is never coming.

`v1` should be left alone. Workflows for a pull request come from its base branch, so `v1` pull requests still use `v1`'s own 2024-era workflow files and requiring `CI Gate` there would deadlock.

## New check names

`Build / ubuntu-latest`, `Build / windows-latest`, `Coverage / ubuntu-latest`, `Coverage / windows-latest`, `Format / dotnet format`, `E2E / dotnet test`, `AOT / {linux,win,osx}-{x64,arm64}`, `Packaging / dotnet pack`, and `CI Gate`.

## Notes for review

- The reusable workflows have to sit directly in `.github/workflows/`. GitHub does not support subdirectories for them, and a subdirectory reference fails at run time rather than being caught by a linter, so the filename prefix is standing in for a folder.
- Job bodies are otherwise carried over verbatim, including coverage still running an implicit Debug build while `build` runs Release. `build` and `coverage` therefore run the same suite twice. Worth merging later, but it would move the Codecov numbers, so it is deliberately not in this change.
- Validated with `actionlint`, which also cross-validates the reusable workflow inputs and secrets, and with `zizmor`. The gate predicate was exercised locally against all-green, coverage-skipped, failure, cancelled and unexpected-skip inputs.
- `pull-requests: write` is dropped from build, coverage, format and e2e. It was applied uniformly in #439 when permissions blocks were first added, rather than per job. `--report-github` emits workflow commands and a step summary, and codecov-action uploads with its own token, so none of the four touch the pull request API. `lint-pr` and `release` keep theirs.
- All six checkouts set `persist-credentials: false`. Nothing runs git after checkout, so there is no reason to leave the job token in `.git/config`.
- `OpenFeature.slnx` had stale entries for the deleted workflows, plus a dead `codeql-analysis.yml` left behind by #700. Repointed at the `reusable-*` files and added the two composite actions.


.github/workflows/ci.yml
```diff
@@ -3,106 +3,130 @@ name: CI
...
+    secrets:
+      codecov-token: ${{ secrets.CODECOV_UPLOAD_TOKEN }}
 
-    runs-on: ${{ matrix.os }}
+  format:
+    name: Format
+    permissions:
+      contents: read
+      pull-requests: write
+    uses: ./.github/workflows/reusable-format.yml
```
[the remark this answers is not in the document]
askpt: Good catch, fixed in 95c5504.

Repointed the three stale entries at their `reusable-*` successors and added `reusable-build.yml` / `reusable-packaging.yml` so all six are listed.

Two extras while in that block:
- Removed `codeql-analysis.yml`, a dead entry left over from #700 where the workflow file was deleted but the solution item was not.
- Added `dco-merge-group.yml`, which was never listed.

Every `<File Path>` in the solution now resolves, and every file in `.github/workflows/` is listed. Worth noting the CLI tolerates dangling solution items (`dotnet format --verify-no-changes OpenFeature.slnx` has been passing on `main` with the broken CodeQL entry for a while), so this is an IDE/navigation fix rather than a build break.

.github/workflows/reusable-packaging.yml
```diff
@@ -0,0 +1,65 @@
...
+      # `github.event.pull_request` is absent outside `pull_request`, and GitHub coerces the
+      # missing value to 0 when comparing against a boolean, so `... .fork == false` is also
+      # true on `push` and `merge_group`. The event name must therefore be checked
+      # explicitly, otherwise every merge queue entry would publish a package for a
+      # `gh-readonly-queue/*` ref that is thrown away moments later.
+      - name: Publish NuGet packages (base)
+        if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false)
+        env:
+          NUGET_API_KEY: ${{ secrets.GITHUB_TOKEN }}
+        run: dotnet nuget push "src/**/*.nupkg" --api-key "$NUGET_API_KEY" --source https://nuget.pkg.github.com/open-feature/index.json
```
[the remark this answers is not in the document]
askpt: Not correct - the glob does resolve, and this line is unchanged by this PR.

`dotnet pack` writes to `bin/<Configuration>/`, which for these projects is `src/<Project>/bin/Release/`. That is under `src/`, so `src/**/*.nupkg` matches it. There is no `PackageOutputPath`, `UseArtifactsOutput` or `ArtifactsPath` anywhere in `build/` or `src/`, so the default layout applies.

Verified two ways.

Locally, running the exact quoted command against a local feed:

```
$ dotnet pack -c Release --version-suffix "test.20260803T154137"
  Successfully created package ".../src/OpenFeature.Hosting/bin/Release/OpenFeature.Hosting.2.14.0-test.20260803T154137.nupkg"
  ...

$ dotnet nuget push "src/**/*.nupkg" --source /tmp/localfeed
Pushing OpenFeature.Hosting.2.14.0-test.20260803T154137.nupkg to /tmp/localfeed...
Your package was pushed.
Pushing OpenFeature.2.14.0-test.20260803T154137.nupkg to /tmp/localfeed...
Your package was pushed.
Pushing OpenFeature.Providers.MultiProvider.2.14.0-test.20260803T154137.nupkg to /tmp/localfeed...
Your package was pushed.
```

And in production, from the most recent successful `ci.yml` run on `main` (run 30824223751, today):

```
packaging  Publish NuGet packages (base)  Pushing OpenFeature.Hosting.2.14.0-ci.20260803T145310.nupkg to https://nuget.pkg.github.com/open-feature/...
packaging  Publish NuGet packages (base)  Your package was pushed.
packaging  Publish NuGet packages (base)  Pushing OpenFeature.2.14.0-ci.20260803T145310.nupkg ...
packaging  Publish NuGet packages (base)  Your package was pushed.
packaging  Publish NuGet packages (base)  Pushing OpenFeature.Providers.MultiProvider.2.14.0-ci.20260803T145310.nupkg ...
packaging  Publish NuGet packages (base)  Your package was pushed.
```

The quoting is what makes it work: the pattern is passed through to NuGet unexpanded so NuGet does the globbing itself, which is the documented workaround for the shell-expansion problem. The analysis script found zero matches because a fresh checkout has no `bin/` directories - nothing had been packed yet.

The related `if-no-files-found` point was valid and is fixed in 7225364.

.github/workflows/ci.yml
```diff
@@ -3,106 +3,126 @@ name: CI
...
+# filtered out. A workflow skipped by `paths`/`paths-ignore` reports no check runs at all,
+# and a required check that never reports blocks the pull request forever and eventually
+# times the entry out of the merge queue.
+#
+# `needs:` cannot reference a job in a different workflow file, which is why every gated
+# job is called from here. The work itself lives in the `reusable-*.yml` building blocks.
+# Those have to sit directly in `.github/workflows/` - GitHub does not support reusable
+# workflows in subdirectories, and a `./.github/workflows/<dir>/<file>.yml` reference fails
+# at run time rather than being caught by a linter. Hence the filename prefix instead of a
+# folder.
```
kylejuliandev: thought(minor): is this comment needed? it feels like context you'd find in the commit or PR description

.github/workflows/ci.yml
```diff
@@ -3,106 +3,126 @@ name: CI
...
+concurrency:
+  # Supersede in-flight runs for the same pull request, but never cancel a merge queue run:
+  # a cancelled run reports a non-success conclusion and ejects the pull request from the
+  # queue. Deliberately not set in the called workflows - sharing a group between caller and
+  # callee makes a run cancel itself.
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
```
kylejuliandev, line 20: thought: is this needed if every job has the permissions section defined? or vice versa, does every job need a separate permissions if we declare on at the ci level?

.github/workflows/ci.yml
```diff
@@ -3,106 +3,126 @@ name: CI
...
-      attestations: write
+    uses: ./.github/workflows/reusable-packaging.yml
+
+  # The single required status check. Branch protection lists `CI Gate` and nothing else
+  # from this workflow, so jobs and matrix legs can be added, removed or renamed without
+  # silently detaching branch protection.
+  #
+  # `if: always()` is mandatory. Without it this job inherits the implicit `success()`
+  # condition, so a failing dependency makes it *skipped* - and branch protection treats a
+  # skipped check as satisfied, which would let exactly the broken pull requests through.
```
kylejuliandev: thought: is this comment needed?

.github/workflows/ci.yml
askpt: Yes. If we don't specify permissions, the CodeQL can trigger a warning here.

.github/workflows/ci.yml
askpt: Fair. Removed with https://github.com/open-feature/dotnet-sdk/pull/801/commits/078e72510c01f1d109175e5017faca3303bc7864

.github/workflows/ci.yml
askpt: Removed with https://github.com/open-feature/dotnet-sdk/pull/801#discussion_r3706390756

.github/workflows/ci.yml
```diff
@@ -3,106 +3,107 @@ name: CI
...
       contents: read
-      pull-requests: write
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest]
+    uses: ./.github/workflows/reusable-build.yml
+
+  coverage:
+    name: Coverage
+    # Coverage is a reporting concern, not a merge gate. The `gh-readonly-queue/*` ref a
```
kinyoklion: I am not completely sure this comment is correct. While this step could be skipped, I don't think anything actively skips it for a PR? I am not exactly sure what happens for a fork where the token theoretically won't be accessible.

But when it does run `fail_ci_if_error` is set to true.

kylejuliandev (approved):

.github/workflows/ci.yml
askpt: Coverage is intentionally skipped for merge_group runs. `ci-gate` still depends on this job and is allowed to treat only this job's `skipped` result as non-failing. If you see the `ci-gate` needs, `coverage` is allowed to end up with result `skipped`. Makes sense?

I can reword the comment to clarify.

toddbaert (approved):
This is awesome - I'm aware of the problem you're describing but this is a good solution. I support this and if it works here, I'd like to reproduce it in other repos.

askpt:
> This is awesome - I'm aware of the problem you're describing but this is a good solution. I support this and if it works here, I'd like to reproduce it in other repos.

Yeah. This is because I literally broke `contrib-dotnet` this weekend and `go` was lacking some checks also.

.github/workflows/ci.yml
kinyoklion: I would keep it simple. "Coverage isn't required for the merge queue."

But I am still a little confused how it works correctly for PRs from forks.

askpt: For forks, it is the same way as previously. See the file diff here: https://github.com/open-feature/dotnet-sdk/pull/801/changes#diff-d65398dd7d238281d77ad96107b34cc2a5c1a2737c683e34b3fcb8cffb39fb89

Also: #791 

Fixed comment: https://github.com/open-feature/dotnet-sdk/pull/801/changes/2f4501212b8d5be4bdeda384472efe7e6878ac81

.github/workflows/lint-pr.yml
```diff
@@ -1,6 +1,8 @@
 name: "Lint PR"
 
 on:
+  # Required for the merge queue mandatory check
+  merge_group:
   pull_request_target:
```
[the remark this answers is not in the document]
askpt: Fixed with https://github.com/open-feature/dotnet-sdk/pull/801/commits/4e6d373cef69254bb157544cde0fc5b143f05439

